PR #3, opened 2018-01-11
sync with METADATA

KristofferC:


KristofferC:
Forgot to update my local metadata. Will make new commit

PR #5, opened 2018-02-05
add archives

KristofferC:


KristofferC:
GitHub releases are not intended to be stable and should therefore not be hashed like this.

PR #10, opened 2018-02-21
the great bikeshed rename + updates

StefanKarpinski:
Implement https://github.com/JuliaLang/Pkg3.jl/issues/142, goes with https://github.com/JuliaLang/julia/pull/26152.


StefanKarpinski:
I need to merge this first so that https://github.com/JuliaLang/julia/pull/26152 can pass tests. This will break the heck out of Pkg3 until that's merged so please bear with me.

PR #13, opened 2018-07-29
Correct Registry.toml

fredrikekre:
I use the name entry to decide the name of the registry (duhh) but this was not updated, so when adding this registry it still ends up in `registries/Uncurated`.


KristofferC:
I think the updater script will clobber this straight away. I think I need to fix it on my end.

fredrikekre:
Cheers.

PR #28, opened 2019-02-09
add missing versions for DiffEqBase, DiffEqNoiseProcess, and StochasticDiffEq

ChrisRackauckas:
These 3 packages did not add their recent updates to the Versions file. This should at least temporarily fix the package manager so that way it's not down until this is fixed.


ChrisRackauckas:
This won't work. I need to change all of the deps.toml and compat.toml files.

StefanKarpinski:
Manual changes here will just get overwritten in a few minutes by the automatic sync script.

PR #31, opened 2019-02-28
Fix a dependency

EricForgy:
Hi,

I hope this can get merge, but understand if it is better to wait for the new registration process.

Neither my REQUIRE file nor my Project.toml indicate a dependency on Test (v0.12.0), yet, for some reason, it is showing up as a dependency here on General.

This is not the end of the world, but slightly annoying because I have 7 other packages (and growing) that depend on this package and all of the manifests are being polluted by this extraneous dependency on Test. I understand they are stdlib dependencies, but still, it would be nice to clean it up if possible.

In any case, I truly appreciate all the hard work and brainpower going into this and I look forward to the new registration process. Thank you 🙏

### DIFF
--- a/C/Currencies/Deps.toml
+++ b/C/Currencies/Deps.toml
@@ -11,8 +11,6 @@ DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 CurrenciesBase = "a33ca353-0707-5c2b-b398-646075a850cd"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
-
-["0.10-0.12"]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 ["0.11-0.12"]


StefanKarpinski:
This registry is autogenerated so this will get overwritten if it gets merged. Stdlib dependencies are determined by regex search of code in the package. Is the extra dependency causing any particular problems?

EricForgy:
Ah. Ok. No worries. 

> Is the extra dependency causing any particular problems?

No problems other than the minor annoyance of having all my Manifests polluted, but I can live with that for now. There must be something wrong with the regex because this package doesn't depend on Test, but I assume that will be fixed. Thanks 🙏

StefanKarpinski:
https://github.com/JuliaFinance/Currencies.jl/blob/61d3a636a02417e301a65607123b0b584dbf5f3e/test/runtests.jl#L4

EricForgy:
In my other packages, I'm pretty sure `[extras]`  are not dependencies, but no biggie:

https://github.com/JuliaFinance/Currencies.jl/blob/master/Project.toml#L9-L10

StefanKarpinski:
The registry generation doesn't look at the project file at all.

PR #74, opened 2019-04-16
fix compatf or DrWatson

fredrikekre:
Would be good to keep an eye out for these before merging, and fix the registrator issua asap.


StefanKarpinski:
Yes, I'm reviewing PRs and fixing them before merging until we can fixed version of Registrator deployed which won't happen until I fix the bug and @nkottary can deploy the fix, so not until late tonight, I'd guess.

fredrikekre:
I will push CI testing to General soonish too

PR #89, opened 2019-04-17
Fix: Adding TranscodingStreams 0.9.4

vdayanand:
Looks like `Compat.toml`, `Deps.toml` were not updated


fredrikekre:
I think it is already correct. TranscodingStreams/Project.toml does not specify those deps or compatibilities you add in this PR.

tanmaykm:
Should a warning be thrown if no compat entries are found in Project.toml?

KristofferC:
We discussed enforcing compat entries for all dependencies yesterday.

PR #139, opened 2019-04-19
Clean the registry.

fredrikekre:
https://github.com/JuliaComputing/Registrator.jl/pull/95 seems to be deployed. This PR cleans out unecessary information (like user emails etc) from the registry.


chakravala:
why are the `author` names being removed? i'm just curious about it

fredrikekre:
Because the registry is not a package metadata database.

PR #146, opened 2019-04-19
StatsModels 0.5.0 supports StatsBase 0.30

nalimilan:
Next release will be breaking, so better adjust bounds here.

Cc: @kleinschmidt


kleinschmidt:
LGTM

PR #236, opened 2019-04-23
Fix incorrect Julia upper bound in Nabla 0.11.1

ararslan:
Julia's upper bound was incorrectly identified as 1.1 for the most recent tag of Nabla, making it uninstallable on Julia 1.2 and nightly.

### DIFF
--- a/N/Nabla/Compat.toml
+++ b/N/Nabla/Compat.toml
@@ -26,7 +26,7 @@ DiffRules = "0.0"
 DualNumbers = "0.6"
 FDM = "0.1-0.4"
 SpecialFunctions = "0.5-0.7"
-julia = "1.0-1.1"
+julia = "1.0-1.3"
 
 ["0.2-0.10"]
 DualNumbers = "0.6"


ararslan:
I noticed that the Julia bound for Nabla 0.11.0 is `1.0.0-1`, but all other versions have Julia specified as `x.y-z.w`, e.g. `1.0-1.3`. That inconsistency is a bit odd, but was the only reason Nabla was installable at all on nightly.

fredrikekre:
should this somehow be merged with the julia compat for other versions?

ararslan:
I'm not sure what you mean. Could you elaborate?

fredrikekre:
Not sure it matters, but this could be compressed according to my edit above, but not sure we care.

fredrikekre:
Would be nice with some functionality to modify compat for a version, it is not as easy as before to do that with the compressed structure.

ararslan:
Oh, I see. Up to you.

StefanKarpinski:
This is fine, we're going to normalize this stuff in a few days. I'm working on a discourse post explaining the planned process.

PR #252, opened 2019-04-24
normalize compat bounds with new version compression

StefanKarpinski:
Update all compat files to match the new behavior introduced in https://github.com/JuliaComputing/Registrator.jl/pull/113.


A/ACME/Compat.toml
```diff
@@ -1,61 +1,61 @@
-["0.1"]
-julia = "0.3-0.5"
+[0]
```
fredrikekre: Should this be a string like the other ones?

StefanKarpinski: This is just how TOML prints keys that don't have any characters besides alphanumerics, underscores and dashes. Keys are always strings so this doesn't make it an integer or anything like that. Might be good to modify TOML to print values that look like integers in quotes to make this less confusing but it doesn't cause any problems and changing it would require patching TOML.

PR #227, opened 2019-04-23
Increase upper bound of some StatsBase dependencies to 0.30

nalimilan:
These are dependencies for which tests passed with StatsBase 0.30.0 under Julia 1.1. Some false negatives might still need manual fixing.

PR #256, opened 2019-04-24
uncap StatsBase dependencies from 0.29 to 0.*

StefanKarpinski:
This changes the compat range for all packages that depend on StatsBase that were capped at 0.29, assuming that this cap was introduced by the switch from METADATA to General and that they should continue to be compatible with all 0.x versions of StatsBase for now.


StefanKarpinski:
cc @nalimilan, @KristofferC

nalimilan:
Sounds reasonable for now, though of course some packages may not fully work with StatsBase 0.30.0 (e.g. GLM fails its tests because of a small change, which doesn't affect 99% of uses though). There's no ideal solution except stabilizing the API and running automated tests for each breaking release.

StefanKarpinski:
So I should merge?

nalimilan:
I guess we should also do the same thing to all packages whose upper cap was increased from 0.29 to 0.30 by https://github.com/JuliaRegistries/General/pull/227?

StefanKarpinski:
Those were already "uncapped" by the new normalization scheme, which considers all `0.x` versions compatible and therefore renders a concrete upper bound of `0.30` in the absence of any `0.31` or higher versions of StatsBase as the equivalent of a `0.*` upper bound. On the flip side, if and when a `0.x` release is made that is actually incompatible and breaks dependencies, they will all have to change their caps, but that is the sour fruit of pretending that `0.x` releases are non-breaking.

nalimilan:
OK, cool. So what's the recommendation now for compat in Project.toml files? Use `>= X.Y.Z` for all dependencies that still haven't reached 1.0?

PR #265, opened 2019-04-24
uncap dependencies on TransformVariables

StefanKarpinski:
cc @tpapp, please review.


tpapp:
I am fine with this. @goedman, can you cake a look, as it involves your packages too? In principle it should be fine.

goedman:
Yes, this should be fine. Thanks for asking.

PR #389, opened 2019-04-28
ImageTransformations: Update ImageCore bounds to 0.1-0.8

IanButterworth:
@timholy does this look right?

Related to
https://github.com/JuliaImages/ImageTransformations.jl/pull/67
https://github.com/JuliaIO/VideoIO.jl/pull/156

### DIFF
--- a/I/ImageTransformations/Compat.toml
+++ b/I/ImageTransformations/Compat.toml
@@ -46,7 +46,7 @@ julia = "0.6-1"
 ColorTypes = "0"
 
 ["0.5-0"]
-ImageCore = "0.7"
+ImageCore = "0.1-0.8"
 julia = "0.7-1"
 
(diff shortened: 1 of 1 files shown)


timholy:
As long as it's actually true, it's right. (I assume the package passes tests with ImageCore 0.8, right?) In other words, the goal here is to encode reality; the question of whether it will be sufficient to un-stick the JuliaImages ecosystem is a little separate. ImageMorphology just cropped up as a new blocker, perhaps that will be fixed by #390.

ianshmean:
You're right. I tested and it passes locally.

Also I thought it might be worth rooting this issue out across all packages that have `ImageCore = 0.7...` bounds over in #392.

ianshmean:
Closed as covered by #392

PR #578, opened 2019-05-06
Gently loosen Compability to fix installation of OMETIFF

kellertuer:
Fixes an Unsatisfyability error, see https://github.com/JuliaImages/Images.jl/issues/804#issue-440566268.


StefanKarpinski:
Thanks!

PR #704, opened 2019-05-13
reintroduce upper bounds to debugger stack

KristofferC:
These either didn't exist in the first place or were lost in the "normalization process" that ended up removing many upper bounds.

### DIFF
--- a/A/Atom/Compat.toml
+++ b/A/Atom/Compat.toml
@@ -89,12 +89,12 @@ Traceur = "0"
 Juno = "0.6-0"
 
 ["0.8-0.8.3"]
-CodeTracking = "0.4-0"
-JuliaInterpreter = "0.3-0"
+CodeTracking = "0.4-0.5"
+JuliaInterpreter = "0.3-0.5"
 
 ["0.8.2-0"]
 Juno = "0.7-0"
 
 ["0.8.4-0"]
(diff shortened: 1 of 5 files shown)


KristofferC:
cc @pfitzseb, @timholy

A/Atom/Compat.toml
```diff
@@ -89,12 +89,12 @@ Traceur = "0"
...
+CodeTracking = "0.4-0.5"
+JuliaInterpreter = "0.3-0.5"
 
 ["0.8.2-0"]
 Juno = "0.7-0"
 
 ["0.8.4-0"]
-CodeTracking = "0.5-0"
-JuliaInterpreter = "0.4-0"
+CodeTracking = "0.5-0."
```
pfitzseb: Should be just `"0.5"`, no?

KristofferC: oops

timholy:
With big stacks we need better tools to make this less painful (https://github.com/JuliaRegistries/Registrator.jl/issues/122), but this stack is short enough that it's basically a non-issue, and far better to have the bounds.

fredrikekre:
Good to go?

KristofferC:
Yes, if you looked it over and think it seems reasonable.

PR #718, opened 2019-05-13
Update IteratorInterfaceExtensions and TableTraits bounds

davidanthoff:
I tagged non-breaking 1.0.0 release for IteratorInterfaceExtensions and TableTraits. This PR updates all the compat bounds of packages that depend on those two packages to declare that they work with the 1.0.0 versions.

I'm 90% I got this right, but just to confirm: something like `IteratorInterfaceExtensions = "0.1.1-1"` means compatible with `[0.1.1-2)`, correct? I.e. it means compatible with any 1.x.y version. If not, please don't merge.


StefanKarpinski:
Your understanding is correct.

davidanthoff:
Cool, thanks! I tried it locally and it all seems to work fine and as hoped, so I guess that means this can be merged?

PR #812, opened 2019-05-16
Update TensorValues.jl repo url

fverdugo:
The `TensorValues.jl` repo is now on another url (the github organization name has changed).

The new url is:
https://github.com/gridap/TensorValues.jl

Please accept the PR so that the registry info is updated accordingly.

I am the person who triggered JuliaRegistrator for registering this package (see PR #753 ).

Thanks!

### DIFF
--- a/T/TensorValues/Package.toml
+++ b/T/TensorValues/Package.toml
@@ -1,3 +1,3 @@
 name = "TensorValues"
 uuid = "31c64edf-cdeb-50e4-845d-ed2334360c20"
-repo = "https://github.com/lssc-team/TensorValues.jl.git"
+repo = "https://github.com/gridap/TensorValues.jl.git"


fredrikekre:
Its not the same repo though, e.g. the tags are different. Can you transfer the repo instead of duplicating? Can be done on github from the repo settings.

fverdugo:
@fredrikekre good point! 

As you suggested, I have deleted the duplicated repo and transferred the original one to the new organization

fredrikekre:
Great. GitHub redirects from the old URL when you do that, so it is the preferred way to do it.

fverdugo:
@fredrikekre thanks a lot! 

I'll do the same with another repo soon. I will follow the same approach.

Thanks again!

PR #1296, opened 2019-06-09
Update Unitful.jl repo location.

ajkeller34:
I developed Unitful.jl as a post-doc, but have since transitioned to a new employer. Correspondingly I moved Unitful.jl from my personal account to a GitHub org managed by my former research group. Unitful.jl currently lives at https://github.com/PainterQubits/Unitful.jl.git, however nobody in my former group will be maintaining it, and I no longer have admin access to the repo.

Going forward, I will be maintaining a fork of Unitful.jl at [rigetti/Unitful.jl](https://github.com/rigetti/Unitful.jl). As primary developer of this package, I think it makes sense for the general registry to point to this fork. I have discussed this with my former research group.

### DIFF
--- a/U/Unitful/Package.toml
+++ b/U/Unitful/Package.toml
@@ -1,3 +1,3 @@
 name = "Unitful"
 uuid = "1986cc42-f94f-5a68-af5c-568840ba703d"
-repo = "https://github.com/ajkeller34/Unitful.jl.git"
+repo = "https://github.com/PainterQubits/Unitful.jl.git"
--- a/U/UnitfulIntegration/Package.toml
+++ b/U/UnitfulIntegration/Package.toml
@@ -1,3 +1,3 @@
 name = "UnitfulIntegration"
 uuid = "6043864f-155a-5464-b5be-64e090a46784"
-repo = "https://github.com/ajkeller34/UnitfulIntegration.jl.git"
+repo = "https://github.com/PainterQubits/UnitfulIntegration.jl.git"
--- a/U/UnitfulUS/Package.toml
+++ b/U/UnitfulUS/Package.toml
@@ -1,3 +1,3 @@
 name = "UnitfulUS"
 uuid = "7dc9378f-8956-57ef-a780-aa31cc70ff3d"
-repo = "https://github.com/ajkeller34/UnitfulUS.jl.git"
+repo = "https://github.com/PainterQubits/UnitfulUS.jl.git"


fredrikekre:
Seem unfortunate. If they have no interest in maintaining, can't you just transfer the repo to your user again?

ajkeller34:
No, I cannot, as I don't hold the copyright. However, the repo is MIT-licensed, so forking it is allowed and moreover my former PI is okay with this plan. Please see https://github.com/PainterQubits/Unitful.jl/issues/228 for some context. Unfortunately, no code has been committed to Unitful since February, which I understand has been frustrating to some users, but over the past several weeks I've tried to come up with a solution that is acceptable to everyone.

c42f:
This is certainly unfortunate: what we would be loosing here are mainly the list of issues / PRs / other project history which was created in significant part by the user community. I wonder who even owns the copyright of those issues. It's not clear they're covered by the license.

If your group was agreeable, a better technical solution would be to move the repo and then reverse-fork to create an copy under PainterQubits as an archive of the version developed while at Caltech.

If that's not possible for you I guess we've hit a corner case; where despite the development being completely open the project can't be properly moved due to the limitations of the way it's hosted on github. In that case I guess it's just best merge this and get on with life, unfortunate as it is. In a year or so the confusion may be largely forgotten (apart from all those issue references in the git logs).

StefanKarpinski:
> If they have no interest in maintaining, can't you just transfer the repo to your user again?

> No, I cannot, as I don't hold the copyright.

These are not related. The only thing copyright law requires in this case is that the MIT license file indicating that the project's copyright "California Institute of Technology and other contributors" be preserved in all copies, whether they are forks or not. GitHub's concept of transfer versus fork makes no legal difference. There is no legal requirement that the location of repo remain where it was, nor does where it is hosted have any relationship with who owns the copyright.

The only impediment to transferring the original Unitful.jl repo elsewhere is if the PainterQuibits organization is unwilling to do so. Is the head of the Painter Lab unwilling to transfer Unitful.jl to JuliaPhysics? That seems like the right home for it. Doing so would have no effect on the copyright and the LICENSE file would stay the same. It would also preserve the continuity of the issues and so on.

ajkeller34:
As written on PainterQubits/Unitful.jl#228, after discussing further with all relevant parties, the new plan is to maintain Unitful in the PainterQubits organization as an outside collaborator admin, which addresses concerns including retention of issue and pull request history, and is least disruptive for all involved. I had misunderstood the copyright issue due to some conflicting information from multiple people. Thank you to Stefan for some helpful discussion.

 This PR now just updates the registry to reflect where the Unitful family of packages currently lives.

PR #1481, opened 2019-06-20
Update repo URL for RadiationSpectra and SolidStateDetectors

oschulz:
Both packages were moved from JuliaHEP to JuliaPhysics.


oschulz:
I hope this is the correct way of handling a package move?

oschulz:
Or would Registrator automatically update the URL if I make a new release of the packages under their new GitHub org?

christopher-dG:
This is the way to go for now, although we have some ideas on automating this (https://github.com/JuliaRegistries/Registrator.jl/issues/181).

PR #1566, opened 2019-06-25
extended julia versions in Compat.toml

apaoluzzi:


fredrikekre:
You will have to make a new release, this one does not work on Julia 1.0.

PR #1585, opened 2019-06-26
Set Upper Bound for StatsModels

matthieugomez:


fredrikekre:
You will have to do this for all versions, otherwise Pkg will just use an old version of the package.

matthieugomez:
Thanks. Is that better?

fredrikekre:
For versions pre 0.5.2? Maybe its not applicable there?

matthieugomez:
I put StatsModels in the REQUIRE file starting from version 0.5.1 (before, it was part of DataFrames).
That's why the current Compat is bounding it starting from 0.5.1, so maybe that is enough?
Pinging @kleinschmidt

kleinschmidt:
I'm preparing a big PR that will do this for all statsmodels dependents.

matthieugomez:
Thanks @kleinschmidt

PR #1736, opened 2019-07-02
Fix repo URL due to username change

JLTastet:
I unexpectedly broke registration by changing my GitHub user name :flushed:. This PR fixes the repo URL.

Additionally, there is a UUID mismatch between the Project.toml and Package.toml. Not sure how that happened and what is the correct way to fix it. I would happily welcome your suggestions!


fredrikekre:
You should use the one existing in the registry.

PR #1798, opened 2019-07-05
Move MUMPS to MUMPSjInv

abelsiqueira:
Following the discussion [here](https://github.com/JuliaInv/MUMPSjInv.jl/issues/21), MUMPS.jl changed url and name. A different MUMPS.jl packages will be registered afterwards (https://github.com/JuliaSmoothOptimizers/MUMPS.jl/commit/35ac4c77322b819ea360311137a6afe9abf2dd1c#commitcomment-34191000).


fredrikekre:
Not sure how to best handle this, @StefanKarpinski ?

StefanKarpinski:
This is good. Note that any new versions of this package that are released after the move will be strictly incompatible so the next release should be a major one and you need to make sure you cap all packages that depend on this at the old version when that's done.

StefanKarpinski (approved):

abelsiqueira:
Thanks

PR #1817, opened 2019-07-05
Switch Brane Project repositories' url to GitLab

roshii:
I recently discovered https://pkg.julialang.org/registrator but ran into the following error when trying to update a package.

```
ERROR: Registration failed: Changing package repo URL not allowed, please submit a pull request with the URL change to the target registry and retry.
```

If I'm not mistaking I have to update my repo urls with this pull request.

### DIFF
--- a/B/BitConverter/Package.toml
+++ b/B/BitConverter/Package.toml
@@ -1,3 +1,3 @@
 name = "BitConverter"
 uuid = "3a3ce9e8-98e7-11e9-0fa0-055639f146d3"
-repo = "https://gitlab.com/braneproject/bitconverter.jl.git"
+repo = "https://gitlab.com/braneproject/BitConverter.jl.git"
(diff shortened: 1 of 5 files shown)


B/BitConverter/Package.toml
```diff
@@ -1,3 +1,3 @@
 name = "BitConverter"
 uuid = "3a3ce9e8-98e7-11e9-0fa0-055639f146d3"
-repo = "https://github.com/roshii/BitConverter.jl.git"
+repo = "https://gitlab.com/braneproject/bitconverter.jl.git"
```
StefanKarpinski: Shouldn't the package name still be `BitConverter`?

StefanKarpinski:
Does GitLab not support capitals in repo names or something?

roshii:
> Does GitLab not support capitals in repo names or something?

GitLab defaults to lower case letters regardless on capitals in repo name.  
Repository url at GitLab looks like not being case sensitive, i.e. `https://gitlab.com/braneproject/BiTcOiN.jL` redirects to `https://gitlab.com/braneproject/Bitcoin.jl`

URL can and has been changed to include capitals though.

B/BitConverter/Package.toml
roshii: https://github.com/JuliaRegistries/General/pull/1817/commits/6b37209c941569278e19e2f689e8f0178ca3ab27

PR #2051, opened 2019-07-16
Add ApproxFunBase upper bounds

dlfivefifty:
This was due to an internal change in ApproxFunBase.


dlfivefifty:
Realised I was testing on the wrong branch.

PR #2298, opened 2019-07-26
Fix ApproxFunBase upper bounds

dlfivefifty:


A/ApproxFunBase/Compat.toml
```diff
@@ -1,54 +1,51 @@
 ["0-0.0.1"]
-RecipesBase = "0.5-0"
```
fredrikekre: Why do you need to change e.g. this? There is not RecipesBase 1? Similar on the other changes

dlfivefifty: Ah good point. To be honest I thought `-*` and `-0` meant the same thing...

fredrikekre: `-0` means "up to 1"

dlfivefifty: Ok I’ll fix it

PR #2314, opened 2019-07-27
Purge packages which had the source deleted

Nosferican:
These repositories have been deleted or made private for some time now.

Reference: https://github.com/JuliaLang/METADATA.jl/issues/15100


DilumAluthge:
Should you also remove those packages from `Registry.toml`?

Nosferican:
I think the releases that depend on those packages should be yanked...

Nosferican:
Were those merged?

StefanKarpinski:
No, this PR was out of date and I think this needs to be done a bit more carefully.

Nosferican:
I can probably draft something along those lines. Should those yank releases by compat or remove entirely? I am not sure how Pkg.jl behaves in those instances.

PR #2381, opened 2019-07-30
Fix FilePaths compat

davidanthoff:
FilePathsBase v0.6.0 is a breaking release, so we need to add these bounds here.

@rofinn or @oxinabox could you briefly confirm that this is right?

Speed with this one here would be great, I'm supposed to teach a workshop tomorrow, and this is breaking our whole stack right now :)

### DIFF
--- a/F/FilePaths/Compat.toml
+++ b/F/FilePaths/Compat.toml
@@ -19,7 +19,7 @@ Compat = "0.17-2"
 julia = "0.6-0"
 
 ["0.6"]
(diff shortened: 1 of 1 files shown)


rofinn:
This looks correct to me.

davidanthoff:
Ok, in terms of process: lets merge https://github.com/JuliaRegistries/General/pull/2382 first, then I'll update this PR and sort out any merge conflicts, then we can merge this one here.

PR #2438, opened 2019-08-01
rename FDM => FiniteDifferenceMethods

StefanKarpinski:


oxinabox:
@StefanKarpinski  ok yes.

KristofferC:
> ERROR: failed to clone from https://github.com/invenia/FiniteDifferences.jl.git, error: GitError(Code:ERROR, Class:Net, unexpected HTTP status code: 404)

with the current URL.

Please try this out locally with the dependencies etc. I am pretty sure this will break all dependencies using this for all versions. (They are using a package called FDM but after this PR it will be installed as FiniteDifferenceMethods).

F/FiniteDifferences/Package.toml
```diff
@@ -0,0 +1,3 @@
+name = "FiniteDifferences"
+uuid = "e25cca7e-83ef-51fa-be6c-dfe2a3123128"
+repo = "https://github.com/invenia/FiniteDifferences.jl.git"
```
oxinabox:
```suggestion
repo = "https://github.com/JuliaDiff/FiniteDifferences.jl.git"
```

PR #2684, opened 2019-08-14
Fix JuMP compatibility

mlubin:
Pkg is allowing MathOptInterface 0.9 to be installed with JuMP 0.19.0. They're not compatible. This attempts to fix the issue, although I don't understand the format of the version specifiers used here (they're not documented at https://julialang.github.io/Pkg.jl/dev/compatibility/#Compatibility-1).

### DIFF
--- a/J/JuMP/Compat.toml
+++ b/J/JuMP/Compat.toml
@@ -89,7 +89,7 @@ julia = "1"
 Calculus = "0"
 DataStructures = "0"
 ForwardDiff = "0.5-0"
-MathOptInterface = "0.8.1-0"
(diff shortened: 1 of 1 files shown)


rschwarz:
I think the problem is that the `Project.toml` was only added to `JuMP.jl` after version `0.19` was tagged. Back then, it was still based on `REQUIRE`.

The [current syntax](https://github.com/JuliaOpt/JuMP.jl/blob/v0.19.2/Project.toml#L19) should be fine and allow MOI in the range `[0.8.1, 0.8.2)`

PR #2869, opened 2019-08-22
Remove Julia v1.2 compatibility from MOI v0.9.0

blegat:
As discussed in https://github.com/JuliaOpt/MathOptInterface.jl/pull/834, MOI v0.9.0 does not work on Julia v1.2.
MOI v0.9.1 is coming with a fix and this PR should prevents Julia v1.2 users to use v0.9.0.
Let me know if the syntax `1-1.1` is correct.


fredrikekre:
#2889

blegat:
@fredrikekre Even if MOI v0.9.1 fixes the issue with Julia v1.2, it still does not prevent Julia v1.2 users to use MOI v0.9.0 so this PR is still worth merging.

fredrikekre:
IMO it just complicates things, there might be users that use 0.9.0 just fine and this would break things for them.

blegat:
I am not aware of how Pkg will handle this case so I believe you if you say so. JuMP v0.20 will require MOI v0.9.1 in order to force users to upgrade and limit the number of MOI v0.9.0 users on Julia v1.2.

PR #2883, opened 2019-08-23
Upper bound dependencies of OrdinaryDiffEq

devmotion:
We plan to add some breaking changes to DiffEqBase, and hence the compatibility of old versions of OrdinaryDiffEq with DiffEqBase should be corrected (see https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/891). In addition this PR upper bounds all other currently unrestricted dependencies as well.

I'm editing `Compat.toml` for the first time, and I'm not completely sure about the syntax. Is there a difference between specifiers for versions 0.x and 1.x, similar to https://julialang.github.io/Pkg.jl/v1/compatibility/#Caret-specifiers-1, i.e., does
```julia
DiffEqDiffTools = "0.13.0-0"
```
mean the version is compatible with DiffEqDiffTools 0.x.y for all x >= 13 and y >= 0 or only with DiffEqDiffTools 0.13.y for all y >= 0?

### DIFF
--- a/O/OrdinaryDiffEq/Compat.toml
+++ b/O/OrdinaryDiffEq/Compat.toml
@@ -257,15 +257,15 @@ DiffEqBase = "5"
 DiffEqBase = "5.1-5"
 
 ["5.11"]
-DiffEqDiffTools = "0.13.0-*"
+DiffEqDiffTools = "0.13.0-0"
 
 ["5.12-5"]
-DiffEqDiffTools = "0.14.0-*"
-SparseDiffTools = "0.3.0-*"
+DiffEqDiffTools = "0.14.0-1"
+SparseDiffTools = "0.3.0-0.9"
 
 ["5.14-5"]
 ArrayInterface = "1.1.0-1"
-DiffEqBase = "5.20.0-*"
+DiffEqBase = "5.20.0-5"
 
 ["5.2-5.3"]
 DiffEqBase = "5.3.2-5"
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Any numbers that are left off are considered to be wild cards. So `0.13.0-0` means `0.13.0-0.*.*`, i.e. `[0.13.0, 1.0.0)` or in other words anything after 0.13.0 up to but not including 1.0.0 versions.

devmotion:
OK, great, then the changes in this PR seem to work as intended.

StefanKarpinski:
You may have done this already, but if you make the changes to your checked out registry, then you can test whether it works as expected by trying to install packages.

devmotion:
I followed your advice and checked that all affected OrdinaryDiffEq versions (5.4.0, 5.5.0, 5.6.0, 5.6.1, 5.7.0, 5.7.1, 5.8.0, 5.8.1, 5.9, 5.10, 5.11.0, 5.11.1, 5.12.0, 5.13.0, and 5.14.0) can be installed. However, things got a bit more complicated when I started to check which of these releases can actually be compiled (in the initial commit I had set the upper bounds of most dependencies to the latest release). I figured out that apparently (unsurprisingly) most of the older versions are not compatible with the latest versions of DiffEqBase and  other dependencies. Hence I tightened some of the bounds even more.

For all affected OrdinaryDiffEq versions now `using OrdinaryDiffEq` works as well. Unfortunately, for most versions `Pkg.test("OrdinaryDiffEq")` fails since https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/pull/50 actually requires DiffEqOperators >= 4.0.0 but unfortunately DiffEqProblemLibrary's Project.toml was not updated accordingly.

I guess there might be other compatibility issues with previous versions of OrdinaryDiffEq but I'm fine with these changes for now. At least the PR reduces the compatibility issues with DiffEqBase and allows us to introduce breaking changes in DiffEqBase without breaking the whole DiffEq ecosystem, and it seems that old versions of OrdinaryDiffEq can still be installed and compiled.

Any comments @ChrisRackauckas ?

devmotion:
That's already done: https://github.com/JuliaRegistries/General/pull/2921

PR #3091, opened 2019-08-30
Register JuliaBinaryWrappers/p7zip_jll.jl v16.2.0+1

staticfloat:
Autogenerated registration for JuliaBinaryWrappers/p7zip_jll.jl v16.2.0+1


staticfloat:
Previously, this used the `x86` binaries for windows on both `x86` and `x64` windows.  This actually differentiates the two.

PR #3135, opened 2019-09-01
Change Compat bounds on Atom-related packages

timholy:
An attempt to fix https://discourse.julialang.org/t/juno-support-for-julia-1-2-0-not-ready-yet/27804. CC @pfitzseb. Whoever reviews this, please check these bounds carefully, I am not sure this is right (https://github.com/JuliaLang/Pkg.jl/issues/1190).

### DIFF
--- a/A/ASTInterpreter2/Compat.toml
+++ b/A/ASTInterpreter2/Compat.toml
@@ -1,3 +1,3 @@
 [0]
 DebuggerFramework = "0.1.1-0"
-julia = "0.6-0"
+julia = "0.6"
--- a/A/Atom/Compat.toml
+++ b/A/Atom/Compat.toml
@@ -1,5 +1,5 @@
 ["0-0.4"]
-julia = "0.4-0"
+julia = "0.4-0.7"
 
 ["0-0.5.2"]
 Lazy = "0"
@@ -24,7 +24,7 @@ Requires = "0"
 MacroTools = "0"
 
 ["0.5"]
-julia = "0.5-0"
+julia = "0.5-0.7"
 
 ["0.5-0.6.0"]
 ASTInterpreter = "0"
@@ -40,7 +40,7 @@ Lazy = "0.11.3-0"
 Juno = "0.2.6-0"
 
 ["0.6-0.6.15"]
-julia = "0.6-0"
+julia = "0.6"
 
 ["0.6.14-0.7.1"]
 Juno = "0.4.1-0"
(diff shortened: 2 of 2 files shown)


A/ASTInterpreter2/Compat.toml
```diff
@@ -1,3 +1,3 @@
 [0]
 DebuggerFramework = "0.1.1-0"
-julia = "0.6-0"
```
timholy: Note that ASTInterpreter2 got close (on the `master` branch) to working on 0.7/1.0, but there was never a release.

fredrikekre:
I can't tell from the discourse post what the problem was, can it be triggered by some combination of `Pkg.add`s?

timholy:
I think you should be able to trigger it if you `add JuliaInterpreter@0.7.0` followed by `add Atom`.

fredrikekre:
Alright, so the problem is that Atom does not have a version compatible with JuliaInterpreter 0.7

fredrikekre:
Seems like this is enought then
```diff
diff --git a/A/Atom/Compat.toml b/A/Atom/Compat.toml
index cbd3839..d2e5ffe 100644
--- a/A/Atom/Compat.toml
+++ b/A/Atom/Compat.toml
@@ -104,7 +104,7 @@ CodeTools = "0.6.4-0.6"
 CodeTracking = "0.5.7-0.5"
 HTTP = "0.8"
 Hiccup = "0.2.2-0.2"
-JuliaInterpreter = "0.6"
+JuliaInterpreter = "0.6-0.7"
 Juno = "0.7"
 Lazy = "0.13"
 MacroTools = "0.5"
```

fredrikekre:
I'm guessing that JuliaInterpreter 0.7 was not released when Atom 0.9 was released, and hence Atom specified (correctly) that it only supported JuliaInterpreter 0.6. So I think the proper workflow here would be that Atom verifies that it works with the new breaking Juliainterpreter version 0.7, and then make a new release, right? Should be roughly equivalent with my suggested patch though.

timholy:
I was also attempting to head off the error message involving ASTInterpreter2 by imposing julia bounds, but that didn't seem to work. I guess the bounds on julia don't take precedence over other things, and the resolver still tries to install an old version of Atom.

I do think @pfitzseb is going to be making a release shortly, see https://github.com/JunoLab/Atom.jl/pull/162.

A/Atom/Compat.toml
```diff
@@ -99,7 +99,7 @@ Juno = "0.7-0"
 CodeTracking = "0.5.0-0.5"
 JuliaInterpreter = "0.4-0.5"
 
-["0.8.6-0"]
+["0.8.6-0.8"]
```
fredrikekre: I believe this is incorrect, this removes all the compat bounds for Atom 0.9.

timholy:
Thanks @fredrikekre! Really appreciate your reliable help on these matters.

PR #3326, opened 2019-09-06
Place an upper bound on Gtk's Cairo version

timholy:
In reference to https://github.com/JuliaGraphics/Cairo.jl/pull/292. Until we get a shiny new-technology release of Gtk, we have to bound the Cairo version.

CC @lobingera, @tknopp, @giordano, @staticfloat.


davidanthoff:
We might also need an upper bound for Rsvg.jl? Not sure.

tknopp:
yes

timholy:
Nope. All released versions of Rsvg depend on Gtk. This PR bounds Cairo for all Gtk releases, so until Gtk releases a Cairo-0.7 compatible version it's not necessary.

I went through and grepped all the dependencies of Cairo that didn't include Gtk, then wrote a script that tested them against the `master` branch of Cairo in a "clean" environment (Cairo was the only explicitly-installed package):
```sh
Agents: unrelated failures
AsyPlots: uninstallable
Brim: uninstallable
Compose: Fontconfig warnings but the tests pass nevertheless
CairoMakie: unrelated failure (PR)
Drawing: uninstallable
DiffEqTutorials: unrelated failure (permission error)
EcologicalNetwork: uninstallable
Gtk: segfault, fixed by this PR
Harbuzz_jll: uninstallable
Jags: uninstallable
KernelDensityEstimatePlotting: passed (fontconfig warnings)
KernelDensityEstimate: passed
Luxor: passed
Mamba: uninstallable
Mads: uninstallable
ProteinEnsembles: uninstallable
SimpleDirectMediaLayer: segfault (but it's already upper-bounded at Cairo 0.5)
Stan: passed
SimilarityNetworkFusion: uninstallable
Tk: uninstallable
VegaLite: passed
VennEuler: uninstallable
```
Those marked "uninstallable" are *successes*, the ones I've checked are already upper-bounded (in most cases at Cairo 0.5). I submitted one PR fixing issues (https://github.com/JuliaPlots/CairoMakie.jl/pull/20).

Bottom line: this is all we need. Please merge and then we can tag Cairo 0.7.

davidanthoff:
> Nope. All released versions of Rsvg depend on Gtk.

Wait, I don't think that is correct. I think starting with Rsvg.jl v0.2.1 the dependency on Gtk.jl was removed and there were a number of versions released since then that don't depend on Gtk.jl.

I'm also not entirely sure that the logic of your tests works (but I might be wrong about that): say you install a package A that depends on Cairo.jl. Say that package A pulls some of its own binaries via e.g. WinRPM, and those binaries depend on cairo binaries. My understanding is that now your package A will actually trigger WinRPM to install the binaries for package A _and_ the cairo binaries. If you now run the tests for package A, it will probably load the cairo binaries that were deployed via WinRPM. Now they are loaded, and if you _then_ try to run the Cairo.jl tests, things might break, because now you are trying to load two different versions of the same binary. I should say I haven't tested this, but @lobingera pointed that scenario out to me and it seems likely that this is what would happen.

timholy:
>  I think starting with Rsvg.jl v0.2.1 the dependency on Gtk.jl

You're right. But all Rsvg releases are also upper-bounded at Cairo 0.6. Still no problem.

> Say that package A pulls some of its own binaries via e.g. WinRPM, and those binaries depend on cairo binaries.

I only tested packages that have an explicit dependence on the Cairo package. Presumably that means they all load Cairo.jl itself (though I didn't check that). For packages that don't depend on Cairo.jl but do load `libcairo` by other means, how do you propose detecting them?

davidanthoff:
Thinking a bit more about this, maybe we should actually just not worry too much about it, and just make sure the next Cairo.jl uses a version number that is considered breaking (so I guess v0.7.0 or v1.0.0). If any of the depending packages doesn't properly use semantic versioning and then ends up with a version of Cairo.jl that doesn't work for them, it really is on them to fix things, on some level... So I'm in favor of merging this at this point.

timholy:
I should have mentioned at the outset (sorry) that it will be 0.7.0: https://github.com/JuliaGraphics/Cairo.jl/issues/294. This is why I limited Gtk to 0.6.x.

I agree, time to merge this.

PR #3622, opened 2019-09-18
Register JuliaBinaryWrappers/gdk-pixbuf_jll.jl v2.38.2+0

staticfloat:
Autogenerated registration for JuliaBinaryWrappers/gdk-pixbuf_jll.jl v2.38.2+0


KristofferC:
This is an invalid package name (cannit be `import`ed).

staticfloat:
We should have CI catch that. :(

PR #3638, opened 2019-09-18
Remove package Dynare

svillemot:
The package is currently unmaintained and does not work with Julia 1.x.

We (the Dynare Team) are currently working on a complete rewrite of the
package.

But in the meantime, we prefer to keep the package out of the registry, to
avoid misleading potential users.


fredrikekre:
The package can not even be installed on Julia 1.X, so no reason to remove it. See also https://discourse.julialang.org/t/ann-plans-for-removing-packages-that-do-not-yet-support-1-0-from-the-general-registry/28003/14

svillemot:
Still, as the original author, I prefer to have it removed.

StefanKarpinski:
Once you've published something, unless there is a legal or security reason that it must be taken down, it remains published. Going forward we will be automatically archiving all published versions of packages forever rather than just relying on the maintainers git repos to preserve history. In this case, this package will be removed from General soon simply because it does not work on Julia 1.x, so you can just wait for that removal.

PR #3640, opened 2019-09-18
replace old NOMAD package from jbrea with new one

ppascal97:
As mentioned in https://discourse.julialang.org/t/replace-registered-package/28830/2 , I'd like to replace the old NOMAD package registered with mine. I'm in touch with jbrea who is the owner of the former package and he agrees to replace his with mine. I tried to modify all the .toml from N/NOMAD correctly, and hope everything is ok. 

Thanks for your help and hard work !

### DIFF
--- a/N/NOMAD/Compat.toml
+++ b/N/NOMAD/Compat.toml
@@ -1,12 +1,5 @@
-[0]
-Cxx = "0"
-MathProgBase = "0"
+[1.0.0]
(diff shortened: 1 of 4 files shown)


fredrikekre:
Can I suggest the following instead: (i) update the URL to point to your repository, (ii) update the UUID of your package to match the UUID of the old one (iii) make a new release with Registrator.

PR #3660, opened 2019-09-18
Remove illegal names from registry

staticfloat:
This removes the illegal dashed names that Kristoffer pointed out to me; I have incoming PRs to register new names with underscores instead of dashes.


staticfloat:
These packages are all new enough that there should be no problem removing them, as far as I can tell.

PR #3919, opened 2019-09-29
Register JuliaBinaryWrappers/gdk_pixbuf_jll.jl v2.38.2+1

staticfloat:
Autogenerated registration for JuliaBinaryWrappers/gdk_pixbuf_jll.jl v2.38.2+1


giordano:
Why not merged? :confused:

staticfloat:
I think the first time I tried it it didn't work, then the second time I tried it I forgot to add `--register` maybe?

staticfloat:
Or there's a bug in my registration code when a PR already exists.  I'll try running again to see if that's the case.

PR #3994, opened 2019-10-01
Register JuliaBinaryWrappers/Pango_jll.jl v1.42.4+6

staticfloat:
Autogenerated registration for JuliaBinaryWrappers/Pango_jll.jl v1.42.4+6


staticfloat:
Does not include the libraries it should; I hit `register` too quickly.  :P

PR #4080, opened 2019-10-03
Register JuliaBinaryWrappers/GTK3_jll.jl v3.24.11+0

staticfloat:
Autogenerated registration for JuliaBinaryWrappers/GTK3_jll.jl v3.24.11+0


staticfloat:
Mac and Windows only....... _for now_

PR #4190, opened 2019-10-08
Register JuliaBinaryWrappers/Git_jll.jl v2.23.0+0

staticfloat:
Autogenerated registration for JuliaBinaryWrappers/Git_jll.jl v2.23.0+0


staticfloat:
Experimenting with a multi-stage build that needs to properly coordinate registration.

Issue #4121, opened 2019-10-04
Change url repo of package EnKF


mleprovost:
Hello, 

I would like to change the url of the package EnKF.jl (reference of the pull request https://github.com/JuliaRegistries/General/pull/2801) to:
https://github.com/mleprovost/EnKF.jl   but I can't do it with Registrator

Best,

fredrikekre:
You can create the pull request for that yourself without registrator.

mleprovost:
Sorry, this is the first time I register a package. 
I understand that I should make a pull request, but I don't really know how to do it. Should I download JuliaRegistries, modify the file in EnKF to update the url of the file and then make a git push origin master?

Nosferican:
The registry works like any other repository (no magic behind it). Fork the repository, make a new branch and commit the changes to it. After pushing it to the remote, GitHub will ask you whether you want to make a PR (select from yourfork branch to JuliaRegistries/General). The maintainers will review and can accept/commit it.

PR #4228, opened 2019-10-09
Modify url of the package EnKF.jl

mleprovost:
Hello,

I have modified the url of the package EnKF.jl


Nosferican:
Closes #4121.

PR #4442, opened 2019-10-16
Only run Travis builds for pull requests, API builds, and cron builds

DilumAluthge:


DilumAluthge:
@fredrikekre

DilumAluthge:
@fredrikekre After you merge this PR, you can disable the "Auto cancel" settings in Travis. That way, we can feel confident that a CRON job will never get cancelled under any circumstances (because we will disable cancelling).

PR #5000, opened 2019-11-03
Yank ProgressMeterLogging.jl

### LINKED ISSUE #4917: What to do after renaming a package?
I moved my package from https://github.com/tkf/ProgressMeterLogging.jl to https://github.com/tkf/ConsoleProgressMonitor.jl and re-registered it with a different UUID.  ProgressMeterLogging is still installable presumably because github redirects the old repository to the new one.  I renamed old tags so the old release of ProgressMeterLogging.jl is "rooted" (and it's accessible via `ConsoleProgressMonitor#master` anyway).

But are there any other actions I should take after re-registration?  In particular:

* Should I re-create the old repository (https://github.com/tkf/ProgressMeterLogging.jl) with all the releases?  (and maybe archive it afterward.)

* Is there any recommended way to discourage people from installing the old package?  Ideally, `Pkg.add` should fail and it should not be listed in pkg.julialang.org (although it looks like it hasn't been updated for a while ATM).  An approximation I can implement is to recommend installing the new package via `@error` in `__init__` of the old package.

tkf:
close #4917

As I mentioned in #4917, I renamed ProgressMeterLogging.jl to ConsoleProgressMonitor.jl so I would like to make ProgressMeterLogging.jl uninstallable.  From the discussion in #4917, it seems yanking is the best approach, IIUC.

### DIFF
--- a/P/ProgressMeterLogging/Versions.toml
+++ b/P/ProgressMeterLogging/Versions.toml
@@ -1,2 +1,3 @@
 ["0.1.0"]
 git-tree-sha1 = "0cdd35e361f0d00ea1d19e903150080cbb28c029"
+yanked = true


fredrikekre:
Have you tried out what happens? I have only tested yanking when there were other releases left.

tkf:
So it seems that this breaks `resolve` (which I expected to work):

```console
$ cd $(mktemp -d)
...

$ jlpkg --project=. add ProgressMeterLogging
...

$ git checkout tkf/patch-1
...

$ jlpkg --project=. resolve
 Resolving package versions...
ResolverError: Unsatisfiable requirements detected for package ProgressMeterLogging [b0e84e9e]:
 ProgressMeterLogging [b0e84e9e] log:
 ├─ProgressMeterLogging [b0e84e9e] has no known versions!
 └─restricted to versions 0.1.0 by an explicit requirement — no versions left
```

Instantiation works, though:

```console
$ mkdir -p depot/registries

$ ln -s ~/.julia/registries/General depot/registries

$ JULIA_DEPOT_PATH=$PWD/depot jlpkg --project=. instantiate
 Installed LoggingExtras ──────── v0.3.0
 Installed ProgressMeter ──────── v1.1.0
 Installed ProgressMeterLogging ─ v0.1.0

$ ls depot/packages
LoggingExtras  ProgressMeter  ProgressMeterLogging
```

fredrikekre:
Ok, I feel like that should maybe work. Can you open an issue on Pkg?

tkf:
I opened an issue in Pkg.jl.  I'll close this as it's not _super_ important compared to the possibility of breaking things in older julia/Pkg (even though I guess this won't happen in practice).

PR #5065, opened 2019-11-05
Zygote compat

MikeInnes:


fredrikekre:
Why are you also removing bounds?

MikeInnes:
I added what I thought were sensible bounds for the last release, but they turned out to make all packages that depend on Flux uninstallable.

andreasnoack:
I don't know if you have looked into the failures in more detail but I'd suspect that the main source of conflict is the bound on `SpecialFunctions` and that just allowing `0.7` as well would resolve most or all of the conflicts.

Z/Zygote/Compat.toml
```diff
@@ -25,15 +25,16 @@ NNlib = "0.6"
...
 ["0.3.1-0.3"]
 IRTools = "0.2"
+ZygoteRules = "0.1"
 
 ["0.4-0"]
 DiffRules = "0.0"
 FFTW = "1"
 FillArrays = "0"
-ForwardDiff = "0.10"
+ForwardDiff = "0"
```
fredrikekre:
```suggestion
ForwardDiff = "0.10"
```

Z/Zygote/Compat.toml
```diff
@@ -25,15 +25,16 @@ NNlib = "0.6"
...
 IRTools = "0.3"
 MacroTools = "0.5"
-NaNMath = "0.3"
+NaNMath = "0"
```
fredrikekre:
```suggestion
NaNMath = "0.3"
```

Z/Zygote/Compat.toml
```diff
@@ -25,15 +25,16 @@ NNlib = "0.6"
...
 Requires = "0.5"
-SpecialFunctions = "0.8"
+SpecialFunctions = "0"
```
fredrikekre:
```suggestion
SpecialFunctions = "0.7-0.8"
```

MikeInnes:
The `0` bounds seem like a good compromise to me. I've only used this bound for packages that typically have very minimal to no API changes, and that we've never had compatibility issues with before (even when using over-loose "*" bounds). Clearly adding over-tight bounds risks fragility, so short of testing every version individually, this seems sensible.

fredrikekre:
But you are now claiming to be compatible with e.g. `ForwardDiff@0.11`.

MikeInnes:
Yeah, I'm fine with that. ForwardDiff tends to release new minor versions that don't break the parts of the API we care about. Claiming not to be compatible is more likely to be incorrect and introduce issues, in my reckoning.

fredrikekre:
But why can't you just increase the bound when you can actually verify that? E.g. when `ForwardDiff@0.11` is released?

MikeInnes:
I can, just as I can tighten the bound if it proves to cause breakage. Either option can cause issues so I'm making a judgement call about which one is least likely to in practice.

KristofferC:
> Yeah, I'm fine with that. ForwardDiff tends to release new minor versions that don't break the parts of the API we care about.

Is that really true? It was over a year ago ForwardDiff 0.10 was released.

andreasnoack:
FWIW I'd prefer the bounds as well. It's hard to know what the future brings. As a package user, I've experienced breakages lately and I think upper bounds would be an improvement. The only caveat is not to be too restrictive and accidentally exclude versions that actually work which was what happened with SpecialFunctions 0.7 in this case.

MikeInnes:
Yes, and it was released with two small, non-breaking commits compared to 0.9.0: https://github.com/JuliaDiff/ForwardDiff.jl/compare/v0.9.0...v0.10.0

KristofferC:
Which was not correct to do.

The change to make based on that is to more clear for authors registering breaking version that they are in fact making what is considered breaking release and perhaps show a list of packages that now upper bound the package. Not to just go "ForwardDiff did a bad release, this compat thing doesn't work".

MikeInnes:
My understanding is that new API features need to be minor releases; it just happens that minor `0.x` versions are allowed to break as well, but not that they must be breaking. If that's not correct then yes, that needs to be communicated more clearly somewhere.

KristofferC:
> My understanding is that new API features need to be minor releases

That is not correct. Before 1.0, patch releases both encompass bugfixes and features. 

It is described here but clearly that is not enough https://julialang.github.io/Pkg.jl/dev/compatibility/#Behavior-of-versions-with-leading-zeros-(0.0.x-and-0.x.y)-1

> In particular, a package may set version = "0.2.4" when it has feature additions compared to 0.2.3 as long as it remains backward compatible with 0.2.0.

MikeInnes:
Yes, it's not enough to have that note buried in docs that otherwise are mostly about what the resolver treats as compatible (which is not the same thing). I had to look through a couple of times to find that quote. Most references say that Pkg uses semver and link to that spec, so it's not unreasonable for package maintainers to assume that it adheres to it without exceptions like this. Are there other exceptions? Are they listed, or do I have to comb through the Pkg.jl docs to find them?

I have no doubt that any of these issues can be traced back to user error, but if system is fragile to user error over subtleties like this, then the issues are going to continue.

I'll update the bounds because at this point it's more important that the current version of Zygote is installable. But I suspect this is a bad tradeoff in terms of my future time spent dealing with these issues, and it's annoying that I'm not allowed to make that judgement myself.

KristofferC:
>  and it's annoying that I'm not allowed to make that judgement myself.

You are but you just might have to fight a bit for it :P

The problem is when people registered a bunch of versions with these loose bounds, then things suddenly break and they want to fix that. So they make a new release or a PR to the registry to tighten a bunch of bounds. Firstly, this is hard to get right, all previous versions have to be tightened so that no loose old versions are left. Secondly, someone using that set of versions locally is suddenly in an "impossible state" according to the resolver and will get bad error messages and complain and write blog posts about how the Julia package manager sucks. These complaints will not influence the person releasing the package but the persons working on the package manager and maintaining the registry.

MikeInnes:
Those seem like a good reasons *not* to add over-tight bounds to only the most recent version; it is highly likely to either break the resolver or accidentally force people onto older versions of the package.

I see that the ideal solution is to add precise bounds to this version and all previous versions. I'm happy to do that but don't, at this instant, have a deep understanding of the `Compat.toml` format or the issues involved. So I figure a good compromise is to add over-approximate bounds, which is strictly better than the old situation (no bounds) and the new one (over-tight bounds causing the above issues).

If that's not right, please tell me how I can make this mergeable so that my users can install these packages and get back to work.

KristofferC:
> Those seem like a good reasons not to add over-tight bounds to only the most recent version; it is highly likely to either break the resolver or accidentally force people onto older versions of the package.

The bounds are not to the most recent version. It is to the set of versions except that are documented to be backwards compatible. The versions that are not backwards compatibler are because the author of a package intentionally and explicitly decided that packages *should not* automatically use them.

MikeInnes:
I mean the most recent version of Zygote. IIUC, after Fredrik's suggestions only Zygote 0.4+ have tight bounds, which will encourage the resolver to downgrade Zygote to 0.3 if a conflict occurs, which is an issue.

fredrikekre:
You should add bounds to 0.3 then, not remove from 0.4.

MikeInnes:
> I see that the ideal solution is to add precise bounds to this version and all previous versions. I'm happy to do that but don't, at this instant, have a deep understanding of the Compat.toml format or the issues involved. So I figure a good compromise is to add over-approximate bounds, which is strictly better than the old situation (no bounds) and the new one (over-tight bounds causing the above issues).

MikeInnes:
I tagged a new Zygote version just to get this in; this is very much a "perfect is the enemy of the good" situation. I plan to follow up with the better solution, but I'm not sure why that needs to block this improvement. Closing this in favour of https://github.com/JuliaRegistries/General/pull/5131, which should be easier to merge.

PR #5098, opened 2019-11-05
Add bounds to BangBang/Compat.toml

tkf:
This patch adds (upper) bounds to BangBang/Compat.toml so that old versions will not be installed in the future when the latest versions do not support some set of upstream versions.  (As discussed in https://discourse.julialang.org/t/backpropagate-compat-bounds/30737/5)

I tried to mimic how `Compat.toml` is "compressing" the bounds but I'm not sure if I did it correctly.  Would `RegistryCI` find incorrect bound specification?


fredrikekre:
> Would RegistryCI find incorrect bound specification?

It checks that Pkg can actually read the bounds, but not if they "make sense".

tkf:
Thanks. I resolved the conflict.

> but not if they "make sense"

I think it would be nice if the CI can check that the compat bound is not tightened than the existing versions (i.e., it's "no-op").  Is it hard to verify this?  If I were to feature request it, should it go to RegistryCI.jl?  Or maybe to Pkg.jl to add some kind of `Pkg.Registry.*` API to create the correct `Compat.toml` in the first place?

fredrikekre:
We should probably have a package/repo for modification of registry files. Right now it is split up over Pkg/bin and Registrator, but neither place is a good fit IMO.

PR #5131, opened 2019-11-06
Zygote 0.3 compatibility

MikeInnes:
Versions prior to 0.3.3 did not depend on ZygoteRules, so this means that all versions have correct compat bounds for this specific dependency. I'll follow up with more improvements, but would like to get this in so that version 0.3 of Zygote is usable; currently it will just throw an error on load.

### DIFF
--- a/Z/Zygote/Compat.toml
+++ b/Z/Zygote/Compat.toml
@@ -25,6 +25,7 @@ NNlib = "0.6"
 
 ["0.3.1-0.3"]
 IRTools = "0.2"
+ZygoteRules = "0.1"
 
 ["0.4-0"]
 DiffRules = "0.0"


fredrikekre:
This was not correct, Zygote does not depend on ZygoteRules until 0.3.3

fredrikekre:
#5188

MikeInnes:
Are you fixing this for correctness in principle, or did you expect it to cause problems?

fredrikekre:
It caused people not being able to install Zygote (or any package depending on it):
```
pkg> add Zygote
 Resolving package versions...
ERROR: Unknown package ZygoteRules found in the compatibility requirements of Zygote [e88e6eb3]
Stacktrace:
 [1] error(::String) at ./error.jl:33
[...] 
```

MikeInnes:
I see, thanks. Yes, that's definitely worth knowing about.

PR #5135, opened 2019-11-06
Zygote bounds for older releases

MikeInnes:
This is an attempt to make Zygote's compatibility more correct; it's not critical but will hopefully make maintenance easier and avoid spurious downgrades etc. in future. Could use an audit from someone who knows what they're doing to make sure it's not going to go wrong in unexpected ways. If it looks good I'll try to follow up on other packages so that we can start to improve the compatibility situation a bit.

### DIFF
--- a/Z/Zygote/Compat.toml
+++ b/Z/Zygote/Compat.toml
@@ -27,7 +27,7 @@ NNlib = "0.6"
 IRTools = "0.2"
 
 ["0.4-0"]
-DiffRules = "0.0"
+DiffRules = "0.0, 0.1"
 FFTW = "1"
 FillArrays = "0"
 IRTools = "0.3"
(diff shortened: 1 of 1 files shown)


fredrikekre:
Ready?

MikeInnes:
I'm happy with it if you are.

MikeInnes:
The CI error is 

```julia
ArgumentError: invalid version range: "0.0, 0.1"
```

presumably caused by the line:

```toml
DiffRules = "0.0, 0.1"
```

I don't understand why this isn't working; I'm following the docs from Pkg.jl here. Any advice?

MikeInnes:
Ok, I think I have the right syntax, it's just different to what's documented there.

fredrikekre:
Whats documented in Pkg is just `Project.toml` syntax. The registry uses a slightly different format; see e.g. https://github.com/JuliaLang/Pkg.jl/issues/1190

torfjelde:
Bump :heart:

PR #5184, opened 2019-11-07
Adjust bounds for IterableTables in multiple packages

davidanthoff:


fredrikekre:
Have you tested the packages with this new version?

davidanthoff:
It is actually not a breaking release. But I wasn't able to register that version as a patch release because this is the first release that adds a bound on the Julia version, and the bot complained on the patch release that I'm not allowed to narrow the bounds on Julia in a patch release.

PR #5189, opened 2019-11-07
JuLIP.jl repository transfer

cortner:
moved `JuLIP.jl` from `https://github.com/cortner/JuLIP.jl.git` to `https://github.com/JuliaMolSim/JuLIP.jl`

### DIFF
--- a/J/JuLIP/Package.toml
+++ b/J/JuLIP/Package.toml
@@ -1,3 +1,3 @@
 name = "JuLIP"
 uuid = "945c410c-986d-556a-acb1-167a618e0462"
-repo = "https://github.com/libAtoms/JuLIP.jl.git"
+repo = "https://github.com/JuliaMolSim/JuLIP.jl.git"


fredrikekre:
Hmm. It doesn't redirect?

cortner:
Well, technically I'm not even updating this registry anymore but I've moved to a private registry. My preference would have been to just delete `JuLIP.jl` from `General`, but I was told this is not a good idea.

Regarding redirecting: Are you suggesting to just leave the old link?

fredrikekre:
GitHub has a feature were you can transfer repositories and then the old link just redirects to the new. If you can, it would be preferable to do that.

cortner:
ok, I'll reopen this if I run into any problems.

fredrikekre:
Wait, it was correct I think looking at the diff, its just your first comment that has the wrong URL in it.

cortner:
indeed, I used the wrong url there. So are you now suggesting to make this change?

fredrikekre:
Yes.

PR #5230, opened 2019-11-09
Add upper bounds to DSP.jl

### LINKED ISSUE #320: FFTW and DSP collision over Periodograms
Tried to build my package today only to find it was broken by some dependencies.

```Julia
[ Info: Precompiling ChemometricsTools [a9718f02-dbee-5ae5-ad0e-dfbd07fa387b]
WARNING: both FFTW and Util export "Frequencies"; uses of it in module Periodograms must be qualified
ERROR: LoadError: LoadError: UndefVarError: Frequencies not defined
Stacktrace:
 [1] top-level scope at none:0
 [2] include at ./boot.jl:317 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:1044
 [4] include at ./sysimg.jl:29 [inlined]
 [5] include(::String) at /home/caseykneale/.julia/packages/DSP/wwKNu/src/DSP.jl:1
 [6] top-level scope at none:0
 [7] top-level scope at none:2
in expression starting at /home/caseykneale/.julia/packages/DSP/wwKNu/src/periodograms.jl:186
in expression starting at /home/caseykneale/.julia/packages/DSP/wwKNu/src/DSP.jl:13
ERROR: LoadError: LoadError: Failed to precompile DSP [717857b8-e6f2-59f4-9121-6e50c889abd2] to /home/caseykneale/.julia/compiled/v1.0/DSP/OtML7.ji.
Stacktrace:
 [1] include at ./boot.jl:317 [inlined]
 [2] include_relative(::Module, ::String) at ./loading.jl:1044
 [3] include at ./sysimg.jl:29 [inlined]
 [4] include(::String) at /home/caseykneale/Desktop/ChemometricsTools/ChemometricsTools.jl/src/ChemometricsTools.jl:1
 [5] top-level scope at none:0
 [6] top-level scope at none:2
in expression starting at /home/caseykneale/Desktop/ChemometricsTools/ChemometricsTools.jl/src/Preprocess.jl:1
in expression starting at /home/caseykneale/Desktop/ChemometricsTools/ChemometricsTools.jl/src/ChemometricsTools.jl:56
```

dlfivefifty:
This should fix https://github.com/JuliaDSP/DSP.jl/issues/320 though I'm not sure I've done the upper bounds correctly.

I'm not the maintainer for the package so @ararslan might have better ideas.


ararslan:
Thanks for doing this, I think it's a good idea. I can't speak the correctness of the bounds though, so I'll request some reviews.

D/DSP/Compat.toml
```diff
@@ -1,19 +1,18 @@
 ["0-0.5"]
-AbstractFFTs = "0"
+AbstractFFTs = "0-0.4.1"
```
fredrikekre: Should be
```suggestion
AbstractFFTs = "0-0.4"
```
?

galenlynch:
This seems like a great idea. It seems like people are still experiencing issues caused by JuliaMath/AbstractFFTs.jl#26 (e.g. JuliaDSP/DSP.jl#323 and maybe JuliaDSP/DSP.jl#324) that these bounds should prevent. As such, it would be great if this PR could be merged sometime soon to fix those problems.

Is this waiting for @dlfivefifty to implement the changes suggested by @fredrikekre or for the additional reviews?

fredrikekre:
Reverted, causes
```
ERROR: Unknown package AbstractFFTs found in the compatibility requirements of DSP [717857b8]
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] Pkg.GraphType.Graph(::Dict{Base.UUID,Set{VersionNumber}}, ::Dict{Base.UUID,Dict{Pkg.Types.VersionRange,Dict{String,Base.UUID}}}, ::Dict{Base.UUID,Dict{Pkg.Types.VersionRange,Dict{String,Pkg.Types.VersionSpec}}}, ::Dict{Base.UUID,String}, ::Dict{Base.UUID,Pkg.Types.VersionSpec}, ::Dict{Base.UUID,Pkg.Types.Fixed}, ::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.3/Pkg/src/GraphType.jl:278
```

Please at least try to use the packages you modify before submitting PRs.

galenlynch:
Weird, I did get that error transiently on my own computer, but not reliably. AbstractFFTs is a registered package, and the UUID in DSP.jl's `Deps.toml` matches that of AbstractFFTs' `Package.toml`. @fredrikekre what could be causing that error?

dlfivefifty:
> Please at least try to use the packages you modify before submitting PRs.

I did. Someone else can make the PR then, I was just trying to help since its affecting a lot of users.

fredrikekre:
For a version X you are not allowed to put compat on Y unless version X depends on Y.

dlfivefifty:
So how do we fix the bug then? version X depends on Z which in turn depends on Y.

galenlynch:
Oh I think I understand the problem... the merge with `master` changed the headers in `Compat.toml` but not `Deps.toml`

fredrikekre:
You can try to use a tool such as https://github.com/bcbi/RetroCap.jl

dlfivefifty:
More precisely: DSP depends on FFTW which depends on AbstractFFTs. So I guess we need to put the upper bound on FFTW?

fredrikekre:
I don't know what the original problem was, but yea, sounds like it.

galenlynch:
I was mistaken. However, it seems like this fixes it...

in `Deps.toml` AbstractFFTs is currently under the "0-0.5" section, but it's also required for DSP v0.6. If you move the AbstractFFTs line from "0-0.5" to "0" then it works.

fredrikekre:
> but it's also required for DSP v0.6

No it isn't: https://github.com/JuliaDSP/DSP.jl/blob/1ca3c74582a4ea6082609993539b1b8ee2db5ea0/Project.toml#L5-L12

galenlynch:
Ah, you're right.

galenlynch:
So what's the path forward? DSP versions v0-0.6.0 are incompatible with AbstractFFTs v0.5, but because DSP v0.6 does not explicitly depend on AbstractFFTs, it can't specify which versions it's compatible with. But as @ararslan pointed out in JuliaMath/FFTW.jl#124, FFTW re-exports AbstractFFTs, and FFTW v0-v1.0 does not specify an upper bound on AbstractFFTs. Therefore, it seems like DSP cannot restrict which version of AbstractFFTs it is used with by setting an upper bound on FFTW. Is the only solution to make DSP depend on AbstractFFTs so it can specify its compatibility?

galenlynch:
Maybe another solution would be to add compatibility upper bounds to older versions of FFTW.

galenlynch:
It seems weird that you have no control over what FFTW exports if you use a version of it lower than v1.1, because it re-exports AbstractFFTs and and has no upper bound on the version of AbstractFFTs.

dlfivefifty:
Probably a simple solution is just restrict FFTW to < 1

galenlynch:
I don't think that will fix it, since FFTW < 1 doesn't place any bounds on which version of AbstractFFTs you get.

dlfivefifty:
Then we should add upper bounds to FFTW.jl

fredrikekre:
I guess this is an example why it is a bad idea to `@reexport` one of your dependencies -- packages depend on something they can't control.

PR #5277, opened 2019-11-11
Cap compat entries for all versions of a package EXCEPT the latest version

DilumAluthge:
## How to generate these changes

Step 1: Install the [RetroCap.jl](https://github.com/bcbi/RetroCap.jl) package:
```julia
] add https://github.com/bcbi/RetroCap.jl#master
```

Step 2: Clone the General registry and make a new branch:
```bash
git clone https://github.com/JuliaRegistries/General.git
cd General
git checkout -b yourinitials/retrocap-addcaps
```

Step 3: Run RetroCap with the `NoUpperBound` strategy:
```julia
julia> using RetroCap
julia> add_caps(NoUpperBound(), pwd())
```


KristofferC:
I feel this need to a pretty detailed description of why this change is made. Changing the registry should be as rarely as possble so what's up with this wholesale upper bounding?

DilumAluthge:
This is mainly a proof of concept implementation of @StefanKarpinski‘s suggestion on Slack from last week.

https://julialang.slack.com/archives/CBF3Z1D7V/p1573141684258100

> here’s a stray thought: don’t require upper bounds on new versions, but whenever someone registers a new version, automatically generate upper bounds for all older versions based on the current claimed compatibility and what versions actually exist
> so it’s fine to register `A@1.2` with `B = "≥ 2.3"`
> but then when you register `A@1.3` regardless of whether it has an upper bound or not, you automatically generate a cap for `A@1.2` based on what the current highest version of `B` is; for example, if `B` has a `3.4.1` version, then you would cap `A@1.2` with `B = "2.3 - 3"` since that’s the next breaking release
> the invariant that you want to maintain is that only the newest version of any package has no upper bounds
> since that’s the situation that makes the resolver do bad things
> that way people who are adamant about having upper bounds on their packages being literally the worst burden that anyone has ever been asked to bear in the history of the world, don’t have to

PR #5406, opened 2019-11-14
Upper bound TimerOutputs for release NNlib

DhairyaLGandhi:
Ref: https://github.com/FluxML/NNlib.jl/issues/141

Also, just to confirm, will this guarantee version `0.5.0-0.5.0`?

cc. @MikeInnes

### DIFF
--- a/N/NNlib/Compat.toml
+++ b/N/NNlib/Compat.toml
@@ -5,3 +5,4 @@ julia = "0.7-1"
 
 ["0.6-0"]
 julia = "1"
+TimerOutputs = "0.5.0"


MikeInnes:
This looks right to me, yeah. `0.5.1` should be considered incompatible here.

fredrikekre:
If this is an issue with TimerOutputs, it would be better to just release a new version of that.

MikeInnes:
We're opening an issue there too, but this fixes the immediate compatibility problem and makes things usable; we don't know how long it will take to debug TimerOutputs, discuss whether it can be fixed, and create and tag that fix, since we're not in control of that repo.

Presumably even with a new TimerOutputs tag, the bounds would have to be updated to something like `0.5.0, 0.5.2+`, since NNlib is still not in fact compatible with 0.5.1.

dhairyagandhi96:
Relevant issue on Timer outputs https://github.com/KristofferC/TimerOutputs.jl/issues/65

fredrikekre:
There is already a PR with a version bump (https://github.com/KristofferC/TimerOutputs.jl/pull/66), then we can yank TimerOutputs 0.5.1. It just becomes very complicated if all packages have to fix their compat bounds when a dependency screws up. cc @KristofferC

MikeInnes:
Ok, if we can get that tagged and yank 0.5.1 in a reasonable time frame, then that does sound like a reasonable way to go.

KristofferC:
https://github.com/JuliaRegistries/General/pull/5413

dhairyagandhi96:
With #5413 merged, could we yank the breaking release?
Thanks for the quick fix, @KristofferC!

KristofferC:
The breaking release will never be picked unless someone explicitly starts upper bounding just above it, or add that version specifically. Yanking has some problems now that I rather just leave the release in.

PR #5632, opened 2019-11-20
Adjust bounds for Nullables package

davidanthoff:
The package was stuck in a state with `REQUIRE`, so the update to 1.0.0 really just brings it forward to the current state of doing things. There are no breaking changes in this release, but the previous release was 0.0.8, so there was no way to tag a non-breaking release, so we tagged a 1.0.0 to get into a version range that is easier to handle.

CC @nalimilan


fredrikekre:
Needs a rebase.

davidanthoff:
Done!

PR #5713, opened 2019-11-21
upper bound all previous versions of packages in JuliaDiffEq

ChrisRackauckas:
Script:

```julia
# github
import GitHub

auth = GitHub.authenticate("ENV["GITHUB_AUTH"]") # you need to set this in your ENV

drepos = GitHub.repos("JuliaDiffEq", auth = auth)[1]

#run(`git clone https://github.com/JuliaRegistries/General.git`)
cd(raw"C:\Users\accou\.julia\external\General")
using RetroCap

pkgs = [RetroCap.Package(r.name[1:end-3]) for r in drepos]
pkg_to_path, pkg_to_num_versions, pkg_to_latest_version = RetroCap.parse_registry(pwd())

for pkg in pkgs
    try
        add_caps(NoUpperBound(),pwd(),
                 pkg_to_latest_version,
                 pkg,
                 pkg_to_path[pkg])
    catch
        println("Package $pkg not affected")
    end
end
```

PR #6046, opened 2019-11-30
Fix compat bounds in JuliaDiffEq packages

DilumAluthge:
This fixes compat bounds like `PkgA = "0"` and such.

This was first noticed by @MikeInnes and was reported by @ChrisRackauckas in https://github.com/JuliaRegistries/General/pull/5713#issuecomment-559818536 
> https://github.com/JuliaRegistries/General/blob/697ed8734543666895b70c02a2f65f79812a7669/D/DiffEqFlux/Compat.toml#L8


ChrisRackauckas (approved):
Looks like that did it.

DilumAluthge:
I'll soon be making a breaking release `0.2.0` of RetroCap that has this fix in it.

D/DASKR/Compat.toml
```diff
@@ -1,15 +1,16 @@
...
 ["2-2.3"]
 DiffEqBase = "3-5"
+Reexport = "0.0.0 - 0.2"
 
 ["2.3-2"]
 julia = "1"
 
 ["2.4-2"]
 DiffEqBase = "5"
+Reexport = "0"
```
fredrikekre: What about this?

DilumAluthge: Yeah. That’s because this PR was run with the “exclude the latest version of each package” option.

If we would want to also cap the latest version of each package, we would just need to run it with the “cap the latest version of each package“ option: https://github.com/bcbi/RetroCap.jl#run-on-all-repos-in-a-registry

fredrikekre: Why would you not want to run it for the latest version?

ChrisRackauckas: Yeah, that seems odd. This is the only library I forgot to put a tag in for, so therefore it's the only one in the future that won't have an upper bound? I don't quite get it. But it seems this is the only one- so it can be spot fixed.

DilumAluthge: So here is the backstory. The original motivation for RetroCap was an idea by Stefan in which we would allow package authors to not have upper bounds on the latest versions of their packages, and then every time they released a new version, we would go back and retroactively add caps to all earlier versions.

So that’s why RetroCap was originally set up to add caps for all versions of a package except the latest version.

Anyway, the new version of RetroCap lets you pick whether or not you want to add caps for the latest version: https://github.com/bcbi/RetroCap.jl/blob/master/README.md#run-on-all-repos-in-a-registry

PR #6047, opened 2019-11-30
Urgent: remove duplicated `["3.0"]` section

DilumAluthge:
@ChrisRackauckas This happened because we merged my "compat fix" PR and then shortly thereafter AutoMerge merged a new release of `DiffEqPhysics`.


DilumAluthge:
This PR needs to be merged urgently, because until it is merged, `DiffEqPhysics.jl` will not be installable.

DilumAluthge:
It's actually worse than I thought.

Until this PR is merged, `DiffEqPhysics.jl` will not be installable. But also, AutoMerge will always fail because it won't be able to parse all of the TOML files in the General registry.

staticfloat:
This is interesting; so `AutoMerge` merged something that was invalid TOML; do we not have a check for validity _post_ merge?

KristofferC:
Haven't looked at the details of what happened here but without a bors like queue, stuff like this will happen.

staticfloat:
I think what you're saying is that without invalidating the mergeability of other PR B when PR A is merged, things like this can happen.  That makes sense; I thought that was the difference between the "branch" builds and the "PR" builds that Travis does, but apparently I might be wrong.

KristofferC:
Yes, without running CI through a linear queue, a CI result can just be invalid by something being merged in between.

staticfloat:
Bots such as [autosquash](https://github.com/marketplace/actions/autosquash) and [autorebase](https://github.com/tibdex/autorebase) help with this by automatically rebasing all PRs on top of latest `master` (if it can be done so cleanly).  Perhaps we can write a GitHub action that does this for all open PRs whenever there is a push to `General`.

KristofferC:
Merging a PR gonna cause 40 rebase and CIs to run again on those? And then 39 when the next merge.

staticfloat:
.......possibly with some extra smarts to not rebase PRs with no shared file changes.

DilumAluthge:
> Haven't looked at the details of what happened here but without a bors like queue, stuff like this will happen.

That’s exactly what happened here. There was no syntactic conflict, but there was a semantic conflict.

KristofferC:
> .......possibly with some extra smarts to not rebase PRs with no shared file changes.

Maybe it can be extra smart and not rebase unless the same lines in a file have been changed... Oh wait.

PR #6235, opened 2019-12-04
Specify xvfb and xauth as addons

timholy:
These are in support of GUI packages like Gtk.jl that require a display. Coupled to https://github.com/JuliaRegistries/RegistryCI.jl/pull/121

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -5,6 +5,13 @@ on:
     - cron: '15,45 * * * *'
   pull_request:
 
+addons:
+  apt:
+    packages:
+      - xvfb
+      - xauth
+      # - libgtk-3-0   # libgtk-3-0 would be needed if this ever runs on Julia < 1.3
+
 jobs:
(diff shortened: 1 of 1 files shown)


timholy:
Hmm, while [this](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners) doesn't mention xvfb, [this](https://github.com/actions/virtual-environments/blob/7463ea032be0ad4b38dc7a88ebd5375880a53f7f/images/linux/scripts/installers/1804/basic.sh) seems to suggest it might already be installed.

Perhaps we should just close this and see if https://github.com/JuliaRegistries/RegistryCI.jl/pull/121 is enough on its own?

DilumAluthge (approved):

timholy:
I am guessing that https://github.com/JuliaRegistries/RegistryCI.jl/pull/121 is enough, and since GitHub actions doesn't support `addons` it would need rework anyway.

PR #6336, opened 2019-12-06
Several important RegistryCI updates

DilumAluthge:
This pull request makes the following changes:

Major changes:
1. Update the version of RegistryCI to `0.4.0` (latest version).
2. Update the `julia-actions/setup-julia` action to `1.0.1` (latest version).
3. Run the AutoMerge schedule/cron job five times per hour. Each cron job only takes 2-3 minutes to complete, so I think that it is very reasonable to run it five times per hour. Honestly, we could run it ten times per hour, but I picked the more conservative five times.
4. Add `jlbuild` as an authorized AutoMerge user. The list of authorized AutoMerge users is now: `authorized_authors=["JuliaRegistrator", "jlbuild"]`

Minor changes:
- Rename the AutoMerge job from the generic `build` to `AutoMerge`.
- Add a more descriptive name for the `Install dependencies` step.

---

Please note that the waiting periods have not been modified, i.e. `new packages` remains unchanged at `Day(3)` and `new versions` remains unchanged at `Minute(15)`. This means that a new package PR will never be merged if it is less than three days old, and a new version PR will never be merged if it is less than 15 minutes old.

### DIFF
--- a/.ci/Manifest.toml
+++ b/.ci/Manifest.toml
@@ -5,9 +5,9 @@ uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
 [[BinaryProvider]]
 deps = ["Libdl", "SHA"]
-git-tree-sha1 = "29995a7b317bbd06be147e1974a3541ce2502dca"
+git-tree-sha1 = "5b08ed6036d9d3f0ee6369410b830f8873d4024c"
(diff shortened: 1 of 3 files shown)


DilumAluthge:
The RegistryCI integration tests pass on 0.4.0, so in theory this PR should be safe.

PR #6834, opened 2019-12-17
Impose some upper bounds retrospectively

timholy:
Prompted by https://github.com/JuliaGraphics/ColorTypes.jl/pull/142#issuecomment-566505471. Hopefully I got this right.


timholy:
With the recent releases this is getting confusing.

PR #6856, opened 2019-12-18
Remove mentioning of users

fredrikekre:
Given that actions are enabled by default on repos I get pinged from every fork of the registry too. I suspect you also find this annoying @DilumAluthge ?


DilumAluthge (approved):
Yeah I have no idea why this workflow is running on forks

PR #6903, opened 2019-12-18
Teh/colors

timholy:
This may be the cause of https://github.com/JuliaGraphics/ColorTypes.jl/pull/143#issuecomment-567136645. If this looks OK, can this be merged quickly?


timholy:
Thanks!

PR #7030, opened 2019-12-21
Adding artifacts

jstrube:
Cxx and CxxWrap need a julia binary at build time. Following a discussion with @giordano and @Gnimuc on slack how to specify julia as a binary dependency in BinaryBuilder, this is my first attempt to help with that.
I don't really know what I'm doing so this may or may not work.


giordano:
I think it's better to put this into a JLL package.  Also, I don't know if a `Artifacts.toml` is expected to live in this registry.

jstrube:
OK, so do you suggest to create a new package under `JuliaBinaryWrappers/`?

giordano:
Probably, but I'd suggest to wait for @staticfloat.

PR #7094, opened 2019-12-23
Widen Einsum's bounds on Compat

mcabbott:
Einsum.jl depends on Compat.jl, but gives no version bounds. It's actually only needed for Julia 0.6.

This causes a CI error in my package which needs Compat 2.2, and has Einsum as a test dependency. On Travis it seems that Compat 3.0 gets added before trying to add Einsum. 

The next version of Einsum should drop this dependency, but https://github.com/ahwillia/Einsum.jl/pull/38 is stalled.


DilumAluthge (approved):
Seems fine to me

PR #7145, opened 2019-12-25
Fix WebIO's compat for JSExpr

twavv:
I'm not 100% on the version specification format here. My understanding is that `0.5` means `[0.5.0, 0.6)`.

Please merge ASAP if looks good. :^)

### DIFF
--- a/W/WebIO/Compat.toml
+++ b/W/WebIO/Compat.toml
@@ -25,9 +25,11 @@ Observables = "0.2.2-0"
 
 ["0.8-0.8.1"]
 Widgets = "0.5.1-0"
+JSExpr = "0.5"
 
 ["0.8.11-0"]
 julia = ["0.7", "1"]
+JSExpr = "0.5"
 
 ["0.8.13-0"]
 AssetRegistry = "0.1"
@@ -37,6 +39,8 @@ Observables = "0.2.3-0.2"
 Requires = ["0.4.4-0.5", "1"]
 WebSockets = "1.5.0-1"
 Widgets = "0.6.2-0.6"
+JSExpr = "0.5"
 
 ["0.8.3-0.8.9"]
 julia = "1"
+JSExpr = "0.5"


travigd:
> To register WebIO v0.8.13 (lucky number 13!), I added JSExpr to the compat because I was trying to fix an issue where JSExpr v1 has just been released and was incompatible.

> JSExpr v0.5 depends on WebIO, but in v1.0, we dropped the dependency (and were planning on reversing it for WebIO v1.0). I forgot that WebIO doesn't actually depend on JSExpr though when I added the compat entry.

Is there any way to resolve this issue? Does Pkg allow circular dependencies? :^) Another "solution" might be to un-release JSExpr v1.0 since it was only really released in preparation for WebIO v1.0 (these issues would still exist but they wouldn't be as bad because someone installing JSExpr and WebIO from scratch would get the latest versions that work with each other).

DilumAluthge:
> Does Pkg allow circular dependencies?

No.

DilumAluthge:
What are you trying to accomplish here?

travigd:
JSExpr depended upon WebIO but that was kind of bad™ so we removed the dependency in v1 and were going to have WebIO depend on JSExpr to implement some webio-specific functionality in WebIO v1 (but that's a WIP).

As it stands, because we reversed the direction of the dependency, we can't constraint what version of JSExpr gets installed.

DilumAluthge:
This can’t be merged in its current state because it is causing CI to fail.

I’m not sure this is the correct way to fix this situation.

Perhaps one of @fredrikekre @KristofferC @StefanKarpinski can help out with this after the holidays.

travigd:
> This can’t be merged in its current state because it is causing CI to fail.

Yeah, I made this PR thinking that WebIO depended on JSExpr and then figuring out that we're in the current situation. :^)

One solution would be to allow "peer dependency" specifications (à la "A doesn't depend on B, but if B is installed it must be version v1.2.3") but that's a structural change to the Pkg system and I have a feeling that it wouldn't be well received by most (for probably pretty valid reasons).

I suspect that the immediate fix would be to yank JSExpr v1 for now and un-yank them once we we have WebIO v1.0 ready to go. There would still be some of these issues but at least `]add WebIO JSExpr` would install compatible packages.

DilumAluthge:
If the issue here is that you made a bad release of `JSExpr`, then you should yank that bad release.

DilumAluthge:
Just be aware that you cannot “unyank” a release. So if you released 1.0 and now you want to yank it, you can yank 1.0, but you can’t later unyank it. You can however make a new release, e.g. 1.1 or 2.0.

DilumAluthge:
I think the correct solution here is for you to yank `JSExpr` 1.0. And then, when you are ready, you can release either `JSExpr` 1.1 or 2.0.

The solution definitely isn’t to add all of these dependencies retroactively, so I’m closing this PR.

DilumAluthge:
https://github.com/JuliaRegistries/General/pull/7165

PR #7220, opened 2019-12-28
Update RegistryCI to 0.7.0

DilumAluthge:


DilumAluthge:
Unfortunately, because RegistryCI now uses GitCommand (which in turn uses Git_jll), this means that RegistryCI now requires at least Julia 1.3. So we have to stop running CI on Julia before 1.3.

DilumAluthge:
It is an unfortunate but necessary step caused by the fact that Artifacts are not supported prior to Julia 1.3.

PR #7532, opened 2020-01-05
add upper version bound for julia

johnnychen94:
Instead, we follow the existing version compat:

```toml
["1.2"]
julia = "1"
```

cc: @timholy


DilumAluthge:
That's not really how the compressed registry files work.

The effect of this pull request is that ImageSegmentation 1.3 will now have no Julia compat entry.

This is a better approach: #7533

PR #7632, opened 2020-01-08
Transfer AbstractTrees to JuliaCollections

timholy:


timholy:
[noblock] xref https://github.com/JuliaCollections/Roadmap/issues/1#issuecomment-570904845

PR #7797, opened 2020-01-11
Create labeler.yml

mikolajhojda:


DilumAluthge:
What is the purpose of this?

logankilpatrick:
Hey @mikolajhojda  this repo uses a bot to label the PR's. This is a task for GCI @DilumAluthge where students add a PR label to a repo of their choice.

DilumAluthge:
Ah gotcha. I think we can consider this task complete then!

(I’d rather not add extra GitHub Actions to the General registry, since that would increase the chance that we hit the rate limits, which would be a problem for automerge.)

PR #7829, opened 2020-01-13
Fix JSExpr compat for Knockout.jl

twavv:
JSExpr v1 was yanked (because reasons) but yanking doesn't exist for Julia 1.0.

This should fix the compat so that only versions `[0.0, 1.0)` can be installed (please make sure that's what the bound is actually doing - I'm not 💯 percent certain).

### DIFF
--- a/K/Knockout/Compat.toml
+++ b/K/Knockout/Compat.toml
@@ -10,10 +10,12 @@ JSExpr = "0"
 JSON = "0"
 
 ["0.2.0-0.2.2"]
+JSExpr = "0"
 Observables = "0.2.2-0"
 WebIO = "0.3-0"
 julia = ["0.7", "1"]
 
 ["0.2.3"]
+JSExpr = "0"
 WebIO = "0.8.0-*"
 julia = ["0.7", "1"]


DilumAluthge:
Editing the compressed registry files like this can often break things.

Why not just register a new release of Knockout.jl with the appropriate company entry?

travigd:
My understanding is that this is essentially what RetroCap would have done, just limited to Knockout in particular.

I can definitely just release a patch of Knockout.jl with the appropriate compat, but I'm not certain that would truly solve all of the issues. I'm not sure how the compat solver resolves between candidates where upgrading one package will downgrade another and vice versa; e.g., if Knockout v0.2.0 depends on JSExpr unbounded and Knockout v0.2.1 depends on JSExpr v0.5, what will Pkg do (especially in Julia 1.0 since that's where the issue lies)?

As always, advice is appreciated. :^) I wish we weren't in this situation but a lot of this code was written before the Julia community had really started to take compatibility seriously (and also before I knew what kinds of hell can emerge).

DilumAluthge:
Alright let’s give it a try. If it breaks things, we can revert.

travigd:
Installing Knockout on Julia 1.0 installs the correct version of JSExpr, so that's a good sign! Thank you so much for being so patient throughout all of this. If you ever end up in Ann Arbor, I owe you a coffee/tea (or six).

PR #7856, opened 2020-01-13
Loosen Flux CUDA bounds

DhairyaLGandhi:
cc @MikeInnes 

Hopefully these bounds look about right.

### DIFF
--- a/F/Flux/Compat.toml
+++ b/F/Flux/Compat.toml
@@ -38,7 +38,7 @@ ZipFile = "0.7-0.8"
 
 ["0.10.1-0"]
 AbstractTrees = "0.2-0.3"
-CuArrays = "1.4.3-1."
+CuArrays = "1.4.3-1"
 ZipFile = "0.7-0.9"
 
 ["0.5-0.7"]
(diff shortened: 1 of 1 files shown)


F/Flux/Compat.toml
```diff
@@ -38,7 +38,7 @@ ZipFile = "0.7-0.8"
 
 ["0.10.1-0"]
 AbstractTrees = "0.2-0.3"
-CuArrays = "1.6.0-1"
+CuArrays = "1.4.3-1."
```
MikeInnes: What does the `1.` mean here, as opposed to `1`?

dhairyagandhi96: It was a typo; should be fixed now

Thanks!

MikeInnes: Ah, great.

DilumAluthge:
@dhairyagandhi96 @MikeInnes Is this ready to merge?

dhairyagandhi96:
As long as the format looks reasonable, I think I am good with that

MikeInnes:
Yeah it could do with a sanity check from the experts but LGTM

DilumAluthge (approved):

PR #8257, opened 2020-01-21
Restrict HalfIntegers 0.1 to Julia ≤ 1.3

sostock:
Due to https://github.com/JuliaLang/julia/issues/34134, HalfIntegers 0.1 is incompatible with Julia 1.4.


DilumAluthge:
@fredrikekre What are your thoughts?

fredrikekre:
Either this is a bug in Julia which should be fixed, or this is a bug in the package which should be fixed by releasing a new patch release. This PR will mess with people that uses this package in Julia 1.4.

DilumAluthge:
> Either this is a bug in Julia which should be fixed, or this is a bug in the package which should be fixed by releasing a new patch release. This PR will mess with people that uses this package in Julia 1.4.

Keno’s comment in https://github.com/JuliaLang/julia/issues/34134 implies that this won’t be fixed in Julia, so we are left with the latter option.

DilumAluthge:
@sostock Can you take the bug fix from HalfIntegers 1.0.0, and backport it and make a patch release HalfIntegers 0.1.4? That seems like the best approach.

sostock:
The bugfix from HalfIntegers 1.0.0 works by implementing `div(x::T, y::T, r::RoundingMode) where T<:HalfInteger`, which did not exist before. Therefore, I thought it would be inappropriate for a patch release (since it introduces new functionality), so I called the new release 1.0.0 instead of 0.1.4.

I will try to figure out a way to make `fld`/`cld` in HalfIntegers compatible with Julia 1.4 *without* defining `div(x,y,r)`, then I can release that as 0.1.4.

fredrikekre:
If it is too much of a hassle you can just tell people to use the latest version (1.0.0) instead.

DilumAluthge:
> If it is too much of a hassle you can just tell people to use the latest version (1.0.0) instead.

This seems like the easiest approach. Especially because semver describes pre-1.0 versions as unstable. I think it is perfectly reasonable to just tell users of your package that they need to upgrade to 1.0.0.

sostock:
> I think it is perfectly reasonable to just tell users of your package that they need to upgrade to 1.0.0.

Ok, then I will do that.

PR #8297, opened 2020-01-22
Widen HTTPClient's bounds for Compat.jl

mcabbott:
In trying to update `add Compat@3.1`, I learned that

* MKL https://github.com/JuliaComputing/MKL.jl/blob/master/Project.toml needs PackageCompiler = "0.6.5"
* https://github.com/JuliaLang/PackageCompiler.jl/blob/master/Project.toml needs WinRPM = "0.4.3"
* https://github.com/JuliaPackaging/WinRPM.jl/blob/master/Project.toml needs HTTPClient = "0.1, 0.2"
* and https://github.com/JuliaWeb/HTTPClient.jl is an archived repository, and only has a REQUIRE file, but is listed here as needing Compat <= 2. 

So is this the right way to fix it? 

Had a similar problem in https://github.com/JuliaRegistries/General/pull/7094. Should there perhaps be some global change to all the registry's upper bounds on Compat.jl? These must all have been generated somehow.

### DIFF
--- a/H/HTTPClient/Compat.toml
+++ b/H/HTTPClient/Compat.toml
@@ -1,4 +1,4 @@
 ["0.2"]
-Compat = "0.8-2"
+Compat = "0.8-3"
 LibCURL = "0"
 julia = "0.4-1"


DilumAluthge:
HTTPClient.jl is deprecated. The correct thing to do is for someone to make a pull request to WinRPM.jl that removes the dependency on HTTPClient.jl and uses HTTP.jl instead.

StefanKarpinski:
Does the resulting configuration actually work? If so then this could be ok, but if it doesn't then the registry certainly shouldn't claim that it works. @DilumAluthge's suggested upgrade path does seem like "the right way" to fix it though, although of course it's more work.

mcabbott:
As far as I know, Compat 3 dropped support for Julia <1, but didn't break anything else. Maybe that ought not to have been a major version?

When I try to make a local version of HTTPClient.jl with a looser explicit bound, I run into another similar chain: 

* https://github.com/JuliaPackaging/WinRPM.jl/blob/master/Project.toml also needs Libz = "1"
* https://github.com/BioJulia/Libz.jl/blob/master/Project.toml needs BufferedStreams = "1"
* https://github.com/BioJulia/BufferedStreams.jl also only has a REQUIRE file, but the registry likewise bounds Compat https://github.com/JuliaRegistries/General/blob/master/B/BufferedStreams/Compat.toml

So I make a local version of BufferedStreams.jl... and then I can install!  

Testing HTTPClient.jl  gives me 
```
ERROR: LoadError: LoadError: syntax: extra token "RequestOptions" after end of expression
Stacktrace:
 [1] top-level scope at /Users/me/.julia/dev/HTTPClient/src/HTTPC.jl:18
in expression starting at /Users/me/.julia/dev/HTTPClient/src/HTTPC.jl:18
in expression starting at /Users/me/.julia/dev/HTTPClient/src/HTTPClient.jl:3
ERROR: LoadError: Failed to precompile HTTPClient [0862f596-cf2d-50af-8ef4-f2be67dfa83f] to /Users/me/.julia/compiled/v1.3/HTTPClient/MYzen_fPMgF.ji.
```
and testing BufferedStreams.jl give me
```
ERROR: LoadError: ArgumentError: Package Test not found in current path:
- Run `import Pkg; Pkg.add("Test")` to install the Test package.
```
so I have no idea if these packages work, nor what they do.

DilumAluthge:
> @DilumAluthge's suggested upgrade path does seem like "the right way" to fix it though, although of course it's more work.

https://github.com/JuliaPackaging/WinRPM.jl/pull/177

DilumAluthge:
Given that https://github.com/JuliaPackaging/WinRPM.jl/pull/177 will probably be merged, I think we can fix this issue without modifying the registry.

PR #8468, opened 2020-01-26
update URL after transfer to juliaclimate org

gaelforget:


gaelforget:
Based on this example (https://github.com/JuliaRegistries/General/pull/8323) I think the automerge fail is expected for this. 

I will simply wait for manual merge but please advise if there is anything else I should do

Thanks in advance

gaelforget:
Hi @DilumAluthge 

Thanks for merging this PR. 

Since the URL was updated via this PR, I have been trying periodically  https://github.com/JuliaClimate/MeshArrays.jl/issues/35 to submit the package patch release via `@JuliaRegistrator register()` but no luck -- I still get this error:

`Error while trying to register: Changing package repo URL not allowed, please submit a pull request with the URL change to the target registry and retry.`

I do not understand what else I could do, if anything; don't I have done what the message suggests already? is this just a matter of latency in some system?

Please advise

fredrikekre:
Maybe fix the capitalization?

DilumAluthge:
https://github.com/JuliaRegistries/General/pull/8500

gaelforget:
Thank you both. This fixed it

Did not occur to me that capitalization could matter for a URL (in the registrator pipeline)

PR #4169, opened 2019-10-08
remove packages and versions of packages that do not support Julia 1.0

KristofferC:

PR #8566, opened 2020-01-28
Fix Memento unbounded Compat

omus:
The change that [removed packages which didn't support Julia 1.0](https://github.com/JuliaRegistries/General/pull/4169) caused this compat file to no longer reflect the requirements stated by the packages Project.toml.

The reason I want this change is currently these bounds are stopping Memento from using TimeZones 1.0. I've spoken to the @rofinn about releasing a new version of Memento which should be happening later this week but this will fix the problem right away.

### DIFF
--- a/M/Memento/Compat.toml
+++ b/M/Memento/Compat.toml
@@ -1,15 +1,12 @@
 ["0.10"]
 Compat = "0.65-2"
-JSON = "0.16.1-0"
 Nullables = "0.0.3-1"
-Syslogs = "0"
-TimeZones = "0.7-0"
 julia = "0.6-1"
 
-["0.11-0.12"]
-JSON = "0.16.1-*"
-Syslogs = "0.0.1-*"
-TimeZones = "0.7.0-*"
+["0.10-0.12"]
+JSON = "0.16.1-0"
+Syslogs = "0"
+TimeZones = "0.7-1"
 
 ["0.11-0.12"]
 julia = ["0.7", "1"]
--- a/M/Memento/Compat.toml
+++ b/M/Memento/Compat.toml
@@ -1,12 +1,15 @@
 ["0.10"]
 Compat = "0.65-2"
-Nullables = "0.0.3-1"
-julia = "0.6-1"
-
-["0.10-0.12"]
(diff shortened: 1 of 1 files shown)


omus:
An alternative way of viewing the diff: `git diff --anchored='julia = "0.6-1"' ...`:
```diff
diff --git a/M/Memento/Compat.toml b/M/Memento/Compat.toml
index c89db34b8d..32d8e8497b 100644
--- a/M/Memento/Compat.toml
+++ b/M/Memento/Compat.toml
@@ -1,12 +1,15 @@
 ["0.10"]
 Compat = "0.65-2"
+JSON = "0.16.1-0"
 Nullables = "0.0.3-1"
+Syslogs = "0"
+TimeZones = "0.7-0"
 julia = "0.6-1"

-["0.10-0.12"]
-JSON = "0.16.1-0"
-Syslogs = "0"
-TimeZones = "0.7-0"
+["0.11-0.12"]
+JSON = "0.16.1-*"
+Syslogs = "0.0.1-*"
+TimeZones = "0.7.0-*"

 ["0.11-0.12"]
 julia = ["0.7", "1"]
```

omus:
For reference the Memento Project.toml files:
- [v0.11](https://github.com/invenia/Memento.jl/blob/v0.11.0/Project.toml)
- [v0.12](https://github.com/invenia/Memento.jl/blob/v0.12.0/Project.toml)

omus:
Posting part of the conversation from the Julia Slack which cleared up why having upper bounds is important:

> Curtis Vogt  12:06
> I'd like to adjust some upper bounds set in the registry for a package. The package currently doesn't set any upper bounds so should the registry be updated to reflect that or should I just bump up the current upper bounds? https://github.com/JuliaRegistries/General/pull/8566

> Kristoffer Carlsson  12:39
> You would most likely need to upper bound earlier releases. Otherwise, when the new upper bound actually bounds something, the resolver will just install  the earlier version with no bounds

DilumAluthge:
It seems to me the correct approach is to release a patch release of Memento that is compatible with TimeZones 1.0.

fredrikekre (approved):

PR #8645, opened 2020-01-30
Allow OpenSSL_jll to be installed on Julia 1.0+

staticfloat:
We've updated OpenSSL_jll to declare itself as installable on Julia 1.0+.  However, `RegistryTools.Compress` ignores compatibility bounds changes on a version number that is identical to a previous version except for build number.  Illustration:

```julia
julia> uncompressed = Dict(
           v"1.1.1+0" => Dict("julia" => "1.3.0-1"),
           v"1.1.1+1" => Dict("julia" => "1"),
       )
       RegistryTools.Compress.compress("registries/23338594-aafe-5451-b93e-139f81909106/O/OpenSSL_jll/Compat.toml", uncompressed)
Dict{Any,Any} with 1 entry:
  "1" => Dict{String,Any}("julia"=>"1.3.0-1")
```

This is a pretty reasonable policy, in my opinion; so we'll just work around this by manually editing the Registry.  The older version of `OpenSSL_jll.jl` is not very different from the current version, the main difference is that the latest version (which is what will get installed) throws an informative error if someone attempts to use it from Julia 1.0, whereas the old version throws a non-informative error.


staticfloat:
I've done the same for `Zlib_jll` and `Libiconv_jll`, as those are both needed by Jacob's project.

DilumAluthge (approved):

quinnj:
Thanks @staticfloat!!

PR #8647, opened 2020-01-30
Update Package.toml

DilumAluthge:
Let’s try this

@SimonDanisch


SimonDanisch:
Jeez, whats up with that queue? :D

SimonDanisch:
@DilumAluthge I think you need to restart travis - I also got tests stuck in the queue around that same time, and restart made it execute immediately

PR #9142, opened 2020-02-09
Remove `Libpciaccess_jll`, replaced by `Xorg_libpciaccess_jll`

giordano:
As far as I can see, `Libpciaccess_jll` isn't used by other packages (see [results](https://github.com/JuliaRegistries/General/search?q=libpciaccess_jll&unscoped_q=libpciaccess_jll) of GitHub search, but I also grepped locally the registry and Yggdrasil).  I'm removing it as it has been replaced by `Xorg_libpciaccess_jll` in #9137


DilumAluthge:
Is there any harm in leaving it in the registry?

DilumAluthge:
(I'm not oppose to removing it. Just curious.)

giordano:
> Is there any harm in leaving it in the registry?

No, apart from polluting the registry.

DilumAluthge:
If you are confident that no one is using it, I think it's fine to remove it.

With normal packages, I would say definitely don't remove it. But since JLL packages are such a special case...

DilumAluthge:
I'm writing this explicitly so that in the future if someone sees this, they don't think I support removing packages from the General registry.

**This is only because it is a JLL package.**

DilumAluthge (approved):

giordano:
> If you are confident that no one is using it

Of course I can't control all users out there, but I don't even know why I built this as it isn't used by any package.  Maybe I thought I could use it as dependency of another library, that however I never built.

DilumAluthge:
> I don't even know why I built this

Story of my life, buddy

PR #10058, opened 2020-02-25
Fix InteractBase bounds on Observables.

fredrikekre:
cc @piever


piever:
Looks good, thank you!

PR #10100, opened 2020-02-26
Fix JSExpr bounds on Observables.

fredrikekre:
cc @piever, @travigd


travigd:
I'm confused as to what this PR is meant to be doing. Does v0.3 of observables not work with JSExpr?

fredrikekre:
Maybe? Tests fail, but that might just be bad tests (see https://github.com/JuliaGizmos/JSExpr.jl/pull/34).

The current bounds claim that JSExpr < 0.5.1 is compatible with Observables 0.3, but JSExpr 0.5.1 has Observables 0.2 as an upper bound, so it is mostly that weirdness I want to fix.

travigd:
So, to be clear:
* This PR caps the existing compat for Observables to be `< 0.3`
* The open PR against JSExpr allows 0.3 for JSExpr v0.5.2

Yeah? Looks good to me then.

fredrikekre:
Yes. However, JSExpr probably works with Observables 0.3, but tests fail. Hence my PR to JSExpr.

fredrikekre:
So we could cap existing compat for Observables < 0.4, but should probably still release 0.5.2 with fixed tests then.

travigd:
My understanding is that we (whoever _we_ is) typically don't retroactively update compat to be _broader_ (the goal there is to minimize the differential between installations on existing machines and the state of the registry).

So the allowance for Observables v0.3 should happen in a new (patch release) version of JSExpr.

fredrikekre:
It is generally safer to broader. If you narrow it might be that someone has an "impossible" state.

This PR, however, broades for some releases, and narrows for other (e.g. JSExpr < 0.5.0 claims compatibility with too much).

DilumAluthge:
Bump @fredrikekre 

What’s your plan for this PR? Should I merge it?

fredrikekre:
I'll handle it.

DilumAluthge:
Bump @fredrikekre

fredrikekre:
Waiting on new registrator deployment.

DilumAluthge:
@fredrikekre Looks like there's been a new deployment of Registrator, right?

So this should be good to merge after a rebase.

PR #10324, opened 2020-02-29
Add compat upper bounds for Microbiome and friends

kescobo:
The open-ended `[0]` from #10262 was causing weird conflicts with the BioJulia registry. This adds upper bounds for the last version of each package that was registered in General.

### DIFF
--- a/B/BiobakeryUtils/Compat.toml
+++ b/B/BiobakeryUtils/Compat.toml
@@ -1,4 +1,4 @@
-[0]
+["0-0.1"]
 CSV = "0"
 DataFrames = "0"
 Microbiome = ["0.3", "0.4"]
--- a/M/Microbiome/Compat.toml
+++ b/M/Microbiome/Compat.toml
@@ -1,4 +1,4 @@
-[0]
+["0-0.4"]
 Clustering = "0"
 DataFrames = "0.15"
 Distances = "0.7"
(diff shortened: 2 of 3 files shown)


DilumAluthge:
I'm not sure what the source of the bug is, so I would appreciate it if someone with more expertise could review this PR (and explain the bug, if possible).

DilumAluthge:
Would it be better to just revert https://github.com/JuliaRegistries/General/pull/10262?

DilumAluthge:
@kescobo Just FYI since it is the weekend, this may not get reviewed before Monday.

kescobo:
@DilumAluthge Yeah, I suspected as much - no worries! Now that I know I can edit my local copies of the registry, I'm not in any great rush.

My impression from Kristoffer's comments on slack (in #pkg-usage) is that Pkg doesn't apply the compat entries in a registry to only versions found in that registry. So since this change capped the compat to julia 1.0, and I'm on julia 1.4, it just thinks there are no versions compatible with the julia version I was using.

kescobo:
> Would it be better to just revert #10262?

If for some reason these changes aren't or can't be approved, then I think yes. Or if it's easy for you to revert while we wait, that's fine too - I think the problems I had with the previous state were preferable to the current totally broken state. But if it's a hassle, it's ok with me if it takes a couple of days to sort it out.

StefanKarpinski:
Seems fine to me as long as it doesn't break things. Have you tried using this state of the registry to install packages? Does it work?

kescobo:
@StefanKarpinski I've used it locally to install my packages, and it works. I'm not aware of anything else that depends on these packages.

PR #10451, opened 2020-03-03
Fix DataFrames' bounds on Tables

davidanthoff:
This is a pretty minimal change, it fixes the situation that right now some of these older DataFrame versions declare themselves compatible with Tables 1.0.x, when they are actually not.

Right now this is creating havoc in a lot of places, so it would be good if we could fix/merge this ASAP.

CC @quinnj, @bkamins, @nalimilan


nalimilan (approved):

bkamins (approved):

bkamins:
Thank you!

quinnj:
Thanks!

PR #10472, opened 2020-03-03
Additional bounds on dependencies on old DataFrames.jl releases

### LINKED ISSUE #2137: Fix bounds in registry
It would be good if the [bounds for DataFrames.jl](https://github.com/JuliaRegistries/General/blob/master/D/DataFrames/Compat.toml) in the registry could be cleaned up. 

I started with the most obvious problem https://github.com/JuliaRegistries/General/pull/10451, but looking at the `Compat.toml` for DataFrames.jl, there seem to be other problems as well:
- There are other bounds that end in `-*`, those should probably all be bounded properly.
- Starting with DataFrames.jl 0.14, there are quite a number of versions that declare themselves compatible with any pre 1.0 Tables version. Is that right? Presumably Tables 0.2 was breaking? Not sure, but might be worth checking that as well.

bkamins:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2137

The bounds are added in a way that the bounds are less or equal than the bounds on 0.20.2. In this way - if I understand things correctly - there should be no risk that an old version of DataFrames.jl will be installed instead of latest release because it has looser bounds.

### DIFF
--- a/D/DataFrames/Compat.toml
+++ b/D/DataFrames/Compat.toml
@@ -1,22 +1,22 @@
 ["0-0.11"]
-CategoricalArrays = "0.3.6-0"
+CategoricalArrays = "0.3.6-0.7"
 julia = "0.6-1"
 
 ["0-0.17"]
 CodecZlib = "0.4-0"
 DataStreams = "0.3-0"
-Reexport = "0"
-SortingAlgorithms = "0"
+Reexport = "0.1-0.2"
+SortingAlgorithms = "0.1-0.3"
 TranscodingStreams = "0"
 WeakRefStrings = "0.4-0.5"
 
 ["0-0.18.1"]
 Compat = "0.59-2"
-Missings = "0.2.3-0"
-StatsBase = "0.11-0"
+Missings = "0.2.3-0.4"
+StatsBase = "0.11-0.32"
 
 ["0.12-0.13"]
-CategoricalArrays = "0.3.11-0"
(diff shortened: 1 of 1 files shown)


bkamins:
I understand that AutoMerge fails because this PR requires manual merging - right?

nalimilan:
AutoMerge says "Neither a new package nor a new version", which is correct. I guess it's expected that this kind of PR needs manual reviewing.

nalimilan (approved):

D/DataFrames/Compat.toml
```diff
@@ -50,24 +50,24 @@ Tables = "0.1.15-0"
...
 
 ["0.18.2-0.18"]
-CategoricalArrays = "0.5.2-*"
-Compat = "0.59.0-*"
+CategoricalArrays = "0.5.2-0.7"
+Compat = "0.59.0-2"
 
 ["0.18.2-0.19"]
-PooledArrays = "0.5.0-*"
+PooledArrays = "0.5.0"
```
fredrikekre: For future reference, this should be `"0.5"`, otherwise it excludes e.g. 0.5.1 etc.

fredrikekre: https://github.com/JuliaRegistries/General/pull/10521

bkamins: Thank you - I was not sure about that when I made a PR.

PR #11148, opened 2020-03-18
Limit DiffEqBase >=6.21 to >=1.3

ChrisRackauckas:


GunnarFarneback:
This change isn't effective. It's also necessary to modify
```
["4.22-6.23"]
julia = "1"
```

fredrikekre:
Indeed, see #11397 posted the same minute as you commented :stuck_out_tongue:

GunnarFarneback:
Convergent analysis.

fredrikekre:
Please don't merge changes like this without testing them.

ChrisRackauckas:
It was tested and it worked. It was just reversed by Registrator....

GunnarFarneback:
What Registrator does with conflicting ranges depends on Dict iteration order. I can't really follow the Pkg code but I wouldn't be surprised if the same holds there.

PR #11224, opened 2020-03-19
limit LLVM version to the one supported by Julia

vchuravy:
cc: @staticfloat @giordano @maleadt 

Installing LLVM 9 on Julia 1.3 can cause two separate version of LLVM to be loaded which makes LLVM.jl rather unhappy.
We should also fix this in BinaryBuilder.


DilumAluthge (approved):
It looks correct to me, but it would be great if someone else can also check it.

staticfloat:
This is a good first step; but it will be superseded by https://github.com/JuliaLang/julia/issues/33973

DilumAluthge:
@vchuravy Can you rebase this on master?

vchuravy:
> This is a good first step; but it will be superseded by JuliaLang/julia#33973

Right, but I am worried about Julia 1.3 & 14

DilumAluthge:
I think the syntax is correct on this, but I would appreciate if someone else can confirm.

@vchuravy Just to be safe, can you confirm that when using your branch of the registry, you are able to do things like `] add LLVM_jll`? If so, then it's probably safe to merge this.

vchuravy:
I tested it for LLVM_jll on 1.3

PR #11257, opened 2020-03-20
Widen BufferedStreams.jl's bound on Compat.jl

mcabbott:
This is related to https://github.com/JuliaRegistries/General/pull/8297, for which I made a PR two months ago https://github.com/BioJulia/BufferedStreams.jl/pull/60, which hasn't been merged. 

xref https://github.com/JuliaRegistries/General/pull/7094, a similar change.

BTW, today's path to this is that MAT.jl depends on BufferedStreams.jl:  https://github.com/JuliaIO/MAT.jl/blob/master/Project.toml


fredrikekre:
I assume you tested that it works with Compat 3?

mcabbott:
I even got it to pass on Travis, in that PR. This is my version with a Project.toml file, I did not go as far as editing the local registry to test the version with only a REQUIRE file with Compat@3. 

Compat version 3 also didn't drop anything, except support for Julia 0.6, as far as I know.

mcabbott:
And now I did `] dev Compat` and edited that to be version `2.999.0`. The tagged version of BufferedStreams.jl passes with that.

PR #11377, opened 2020-03-23
Retroactively apply monotonic compat bounds to MLJ repos

ablaom:
Context: https://discourse.julialang.org/t/julia-v1-4-0-has-been-released/36324/12?u=ablaom


ablaom:
@DilumAluthge Many thanks for this prompt attention. ❤️

PR #11477, opened 2020-03-24
Fix rename

mileslucas:
From PR #11253 I have, in less than a day, transferred the repo. I read through the official docs on renaming, but given the short lifetime and the guarantee that there are 0 downstream users of this, could the simple rename be authorized?

### DIFF
--- a/F/Firefly/Package.toml
+++ b/F/Firefly/Package.toml
@@ -1,3 +1,3 @@
 name = "Firefly"
 uuid = "38d273ff-5bca-42e1-8e07-0e7ae0d15fdd"
-repo = "https://github.com/juliahci/Firefly.jl.git"
+repo = "https://github.com/JuliaHCI/Firefly.jl.git"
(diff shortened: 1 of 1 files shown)


F/Firefly/Package.toml
```diff
@@ -1,3 +1,3 @@
 name = "Firefly"
 uuid = "38d273ff-5bca-42e1-8e07-0e7ae0d15fdd"
-repo = "https://github.com/mileslucas/Firefly.jl.git"
+repo = "https://github.com/juliahci/Firefly.jl.git"
```
fredrikekre:
```suggestion
repo = "https://github.com/JuliaHCI/Firefly.jl.git"
```

fredrikekre:
What do you mean with rename? The repo has just moved?

mileslucas:
I just transferred from `mileslucas/Firefly.jl` to `JuliaHCI/Firefly.jl`

fredrikekre:
Yea, that is not problematic at all, github even redirects the old URL.

fredrikekre:
The URL should still be updated as in this PR:

PR #11530, opened 2020-03-25
Update Compat.toml

yakir12:
To elevate a silly restriction on StaticArrays


D/DungBase/Compat.toml
```diff
@@ -1,4 +1,4 @@
 [0]
 IntervalSets = "0.3.2-0.3"
-StaticArrays = "0.11"
+StaticArrays = "0.11-0"
```
KristofferC:
```suggestion
StaticArrays = "0.11-0.12"
```

or?

PR #12098, opened 2020-04-03
Manual fix - VideoIO: Retroactively add upper bound to FFMPEG compat

IanButterworth:
This retroactively adds an upper bound to the FFMPEG compat that was in effect from 0.6.8-0.6.10 inclusive

Done because https://github.com/JuliaIO/VideoIO.jl/issues/227 is lingering


KristofferC:
Seems reasonable to me at least

galenlynch:
Would love to see this merged. The current lack of an upper bound is causing breakage for me.

PR #12117, opened 2020-04-03
New package: Clp_jll v1.16.11+0

jlbuild:

* Registering JLL package Clp_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/Clp_jll.jl
* Version: v1.16.11+0


ViralBShah:
@giordano is it just the name issue? In that case can we merge it?

PR #12136, opened 2020-04-03
Fix PhyloModels v0.3.2 release

jangevaare:
I made a mindless error in v0.3.2 that I registered last night. I would like to change the commit association on the registry to the fixed version please.


DilumAluthge:
Once a version has been registered, we do not allow it to be modified.

If there is a problem with this version, you can yank it. And then you can register a new version.

PR #12266, opened 2020-04-05
Widen GraphPlot's compat: Compose = ["0.7", "0.8"]

DilumAluthge:
Since GraphPlot 0.4 works with Compose 0.8, this is fine.

This should tide us over until the next registration of GraphPlot.

cc: @KristofferC


DilumAluthge:
@KristofferC Any objection to merging this now? We can re-narrow later if we want to, after they register a new version of GraphPlot.

PR #12812, opened 2020-04-12
Revert "New version: ModelingToolkit v1.5.0 (#12792)"

ChrisRackauckas:
This MTK v1.5 tag since it was more breaking than I thought, and is replaced with a 2.0 tag.


fredrikekre:
Please yank instead....

PR #13740, opened 2020-04-27
julia versions: remove all < v0.7, add all ≥ v0.7

StefanKarpinski:
Generated with:
```jl
versions = filter!(>=(v"0.7"), map(VersionNumber,
    filter!(contains(r"^v\d+\.\d+\.\d+$"), readlines(`git tag -l`))))

for v in versions
    sha1 = readchomp(`git rev-parse v$v^{tree}`)
    println("[\"$v\"]")
    println("git-tree-sha1 = ", repr(sha1))
    println()
end
```

### DIFF
--- a/J/julia/Versions.toml
+++ b/J/julia/Versions.toml
@@ -1,96 +1,3 @@
-["0.1.1"]
-git-tree-sha1 = "ef9dab73df8b626ccf8c882c57b39be5b381ede9"
-
-["0.1.2"]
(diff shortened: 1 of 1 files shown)


fredrikekre:
What is this used for?

StefanKarpinski:
Nothing, really. As one can tell by the fact that several of the versions were wrong 😂. But it is used by JuliaHub to know what Julia versions there are—it mostly treats Julia as "just another dependency".

DilumAluthge (approved):

PR #13747, opened 2020-04-27
Add a plain-text file that contains only the version number of the latest stable version of Julia

DilumAluthge:
Julia is installed by default in the GitHub Actions virtual environments.

Right now, when they build their virtual environments, the process for determining which version of Julia they use is:
1. Use the GitHub API to get the latest GitHub release, and then use that.

Right now that is fine because the latest GitHub release is 1.4.1.

But what happens when we release e.g. 1.0.6. Then the latest GitHub release will be 1.0.6. But we won't want 1.0.6 in the virtual environment, we want 1.4.1.

So we really just need a plain-text file that contains the version number of the latest stable release.

Then the GitHub Actions people can pull this plain-text file and get the version number they need.

Also, anyone else that needs the latest stable version number can use this.

cc: @SaschaMann 
cc: @davidanthoff 
cc: @fredrikekre 
cc: @StefanKarpinski 

---

Long term, we should add a simple GitHub Actions "bot" that automatically updates this text file. That should not be very difficult.


StefanKarpinski:
Can the other side not parse the J/julia/Versions.toml file and get the latest version from there?

fredrikekre:
Don't we host a JSON file with Julia versions? I thought Elliot set that up?

StefanKarpinski:
It also doesn't really seem like it belongs in the registry.

DilumAluthge:
> It also doesn't really seem like it belongs in the registry.

That's a good point. I think we need this feature, but there is no reason it should be inside the registry.

PR #13756, opened 2020-04-27
Manual update compat for LazyWAVFiles

baggepinnen:
A bug appeared when used with LazyArrays v0.16 so I think it's best to remove that compatability declaration.


baggepinnen:
Would someone be willing to take a look at this?

giordano:
Note that there is a merge conflict

DilumAluthge:
@baggepinnen Can you rebase on master and fix the merge conflict?

DilumAluthge:
This isn't quite right, which is why Travis is failing.

baggepinnen:
Hmm, then I'll just close this PR, it has been a month now so hopefully all related packages are updated and users will not end up with the incompatible versions

PR #13902, opened 2020-04-30
Update ProfileSVG URL

kimikage:


timholy:
[noblock] I fully support this transfer.

kimikage:
Thanks!

PR #14088, opened 2020-05-03
Update Versions.toml

ahojukka5:
Fixing missing dependency on version v0.3.0. See discussion here: https://github.com/JuliaFEM/FEMBase.jl/issues/68


ahojukka5:
The only change is to move version tag v0.3.0 to a next commit, and there the change is in Project.toml file to introduce a new dependency: (https://github.com/JuliaFEM/FEMBase.jl/commit/5011799fb6712359723fd6036756839055a70f18

```
[deps]
TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
```

ahojukka5:
cc: @KristofferC

DilumAluthge:
As a general rule we don't allow you to modify versions once they've been registered.

If this version is bad, you can yank it. Then you can register a new fixed version.

PR #14154, opened 2020-05-04
Revert "New package: NL2sol_jll v2.2.0+0"

giordano:
Reverts JuliaRegistries/General#14111

We can't actually distribute this package.  I'll also take down https://github.com/JuliaBinaryWrappers/NL2sol_jll.jl, there is little point in keeping a totally broken package in the registry.


giordano:
Put on hold until @macd clarifies the status of the package, whether we can distribute it or not.

macd:
So actually, we are in the clear here.  The ACM was only worried about posting the PDF of the original NL2SOL articles from TOMS which was never my intent. 
The original code is freely downloadable from Netlib.  From Netlib's FAQ:
"The Netlib repository contains freely available software, documents, and..."  and also "Most netlib software packages have no restrictions on their use".  The code itself does not contain any copyright mark.  But I have conservatively put ACM's license language in the license file of NL2sol.f even though I really do not think it is necessary.  Since NL2sol.jl has already been merged into the general registry, there is nothing more left to do so this pull request should just be closed without merging.  Thanks for your help Mose and sorry for the fire drill.

PR #14296, opened 2020-05-06
Yank package versions that depend on X11_jll

giordano:
In this way I _think_ `X11_jll` will disappear from the dependencies of the latest versions of all packages.  I tried to break (and yank) as few versions as possible, so for example for `Cairo_jll` I moved `X11_jll` only to v1.14, but please double or triple check that I did it correctly.


DilumAluthge:
So, it looks correct to me. But because there are so many weird quirks with version numbers that have build numbers, it would be good to have some more reviews.

staticfloat (approved):
looks about right to me.

PR #14368, opened 2020-05-07
Cache artifacts in AutoMerge action

giordano:
I'm not entirely sure this is a good idea, but it may be worth trying it out. Right now many jobs are failing because GitHub is having troubles delivering the artifacts in the US

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -18,6 +18,17 @@ jobs:
       - uses: julia-actions/setup-julia@082493e5c5d32c1fa68c35556429b0f1b2807453 # v1.0.1
         with:
           version: ${{ matrix.julia-version }}
+      - name: Cache artifacts
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/.ci/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
       - name: Install dependencies by running Pkg.instantiate()
         run: julia --project=.ci/ -e 'using Pkg; Pkg.instantiate()'
       - name: AutoMerge.run


DilumAluthge:
We might as well give this a try. If for some reason it causes problems we can always remove it later.

DilumAluthge (approved):

DilumAluthge:
@fredrikekre Any objections to merging this?

.github/workflows/automerge.yml
```diff
@@ -18,6 +18,17 @@ jobs:
       - uses: julia-actions/setup-julia@082493e5c5d32c1fa68c35556429b0f1b2807453 # v1.0.1
         with:
           version: ${{ matrix.julia-version }}
+      - name: Cache artifacts
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Registry.toml') }}
```
giordano, line 27: @SaschaMann do you think is correct to use `Registry.toml` here?  Is my understanding correct that this would invalidate the cache every time this file changes, which would be quite often?

SaschaMann: Yes, that's correct. But it falls back to the `restore-keys` if this key doesn't match an existing cache, so it might still have some benefit even if this file changes quite a lot.

I'm not familiar with AutoMerge/RegistryCI. Do you want to cache artifacts used by their dependencies or some package-dependent artifacts that are used by the to-be-registered package?

`.ci/Manifest.toml` and `.ci/Project.toml` might be a better choice if this is about caching AutoMerge/RegistryCI dependencies.

giordano: > Do you want to cache artifacts used by their dependencies or some package-dependent artifacts that are used by the to-be-registered package?

Ideally both, but mainly at least the dependencies of the AutoMerge pipeline

DilumAluthge: > some package-dependent artifacts that are used by the to-be-registered package

We add the package being registered inside a temporary depot. The path of that temporary depot is created by `mktempdir`. So it is going to be tricky to cache that.

giordano: I see, good point, so only dependencies of AutoMerge then.  I'm going to change this file to `.ci/Project.toml` then

giordano: Done!

DilumAluthge (approved):
@giordano You can merge this whenever it is ready to merge.

giordano:
Let's see how it goes

giordano:
Uhm, it doesn't seem to be working: https://github.com/JuliaRegistries/General/pull/14503/checks?check_run_id=660899894
```
Cache not found for input keys: Linux-test-cache-artifacts-b9e3b8686c362ce99f7c953cd23d3fdddf87e4f4742dc228ff50c29c85b197d6, Linux-test-cache-artifacts-, Linux-test-, Linux-.
```
Any idea why?

DilumAluthge:
@SaschaMann pls halp?

DilumAluthge:
@SaschaMann Can you help with this?

SaschaMann:
Might be a scope issue because it can't find a cache created on master: 

> For example, a cache created for the branch feature-a (with the base master) would not be accessible to a pull request for the branch feature-b (with the base master).

https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

Apparently scheduled builds are not supported by the caching action yet 🙄 https://github.com/JuliaRegistries/General/runs/661611221?check_suite_focus=true#step:4:11

PR #14547, opened 2020-05-11
Add DataFrames upper bounds to RDatasets

nalimilan:


pdeffebach:
does RData need one too? See [here](https://discourse.julialang.org/t/rdata-undefvarerror-identifier-not-defined/39276).

bkamins (approved):

PR #14557, opened 2020-05-11
Add upper bounds to older RData releases

nalimilan:


bkamins (approved):
The only comment is that TimeZones has a higher restriction in 0.6 than in 0.7 (but I think we can ignore this fact).

nalimilan:
> The only comment is that TimeZones has a higher restriction in 0.6 than in 0.7 (but I think we can ignore this fact).

Does it? I thought I took 0.10 for both.

bkamins:
The lower bound, not the upper bound, it is `0.7.3` and `0.7` respectively (it makes little sense, but this is what currently the registry seems to say if I understand correctly how to read it)

PR #14735, opened 2020-05-14
remove Markdown.jl

pfitzseb:
There really shouldn't be any packages in General that have the same UUID as stdlibs.


StefanKarpinski:
We should do this. I think we should just merge this and see what happens. Most likely nothing.

pfitzseb:
Also removed it from Registry.toml now.

pfitzseb:
Bump.

PR #15150, opened 2020-05-22
fix compat for Plots

BeastyBlacksmith:
ref: https://github.com/JuliaPlots/Plots.jl/pull/2727
Can someone confirm, that this is a valid correction?


BeastyBlacksmith:
Should be correct now, checked with RegistryTools

giordano:
I'm not an expert of compat bounds, but Travis doesn't seem to agree that the new bounds are correct

BeastyBlacksmith:
Is fixed

BeastyBlacksmith:
@DilumAluthge anything missing to get this merged?

PR #15457, opened 2020-05-27
CompatHelper does not work on Julia 1.1

DilumAluthge:
cc: @maleadt


DilumAluthge:
See also: https://github.com/bcbi/CompatHelper.jl/pull/211

PR #15503, opened 2020-05-28
New version: OCReract v1.0.1 - Fix: Added tesseract installation in automerge.yml

leferrad:
Since OCReract is a wrapper, the main dependency must be installed for proper testing


leferrad:
Try again

PR #15909, opened 2020-06-05
Migrate into org repo

aorchini:


aorchini:
Thanks - a pain you need to do this by hand? Wouldn't be best for automerge to check url updates too (not just version update)?

fredrikekre:
Maybe, but how often do you migrate repos? I would guess at most once.

PR #16211, opened 2020-06-11
Document the good-faith requirement

DilumAluthge:
Participation in the General registry must be in good faith.

Currently, this requirement is not documented.

This PR documents the good faith requirement.

I think most people would agree with these rules, so hopefully I'll be able to merge this relatively quickly.

cc: @StefanKarpinski 
cc: @sbromberger

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ We do not accept packages or package registrations that:
 2. Are made in bad faith and/or with bad intentions
 3. Waste the time and/or energy of registry maintainers and/or members of the community
 
-Pull requests that violate the above rules will be rejected.
+Pull requests that violate one or more of the above rules will be rejected.
 
 ### Automatic merging of pull requests
 
--- a/README.md
+++ b/README.md
@@ -37,7 +37,6 @@ Packages registered in the General registry must be made in good faith.
 We do not accept packages or package registrations that:
 1. Are obviously and intentionally incorrect
 2. Are made in bad faith and/or with bad intentions
-3. Waste the time and/or energy of registry maintainers and/or members of the community
 
 Pull requests that violate one or more of the above rules will be rejected.
 
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Packages registered in the General registry must be made in good faith.
 
 We do not accept packages or package registrations that:
 1. Are intentionally broken
-2. Are made in bad faith or with bad intentions
+2. Are made in bad faith and/or with bad intentions
 3. Waste the time and/or energy of registry maintainers and/or members of the community
 
 Pull requests that violate the above rules will be rejected.
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ repository after the new release of your package is merged into the registry.
 Packages registered in the General registry must be made in good faith.
 
 We do not accept packages or package registrations that:
-1. Are intentionally broken
+1. Are obviously and intentionally incorrect
 2. Are made in bad faith and/or with bad intentions
 3. Waste the time and/or energy of registry maintainers and/or members of the community
 
--- a/README.md
+++ b/README.md
@@ -30,6 +30,17 @@ requests need to be manually reviewed and merged by a human.
 It is ***highly recommended*** to also use [TagBot][tagbot], which automatically tags a release in your
 repository after the new release of your package is merged into the registry.
 
+### Requirements for packages in the General registry
+
+Packages registered in the General registry must be made in good faith.
+
+We do not accept packages or package registrations that:
+1. Are intentionally broken
+2. Are made in bad faith or with bad intentions
+3. Waste the time and/or energy of registry maintainers and/or members of the community
+
+Pull requests that violate the above rules will be rejected.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,6 +30,17 @@ requests need to be manually reviewed and merged by a human.
...
 repository after the new release of your package is merged into the registry.
 
+### Requirements for packages in the General registry
+
+Packages registered in the General registry must be made in good faith.
+
+We do not accept packages or package registrations that:
+1. Are intentionally broken
+2. Are made in bad faith and/or with bad intentions
+3. Waste the time and/or energy of registry maintainers and/or members of the community
```
Nosferican, line 40: The third point is not too clear.

README.md
```diff
@@ -30,6 +30,17 @@ requests need to be manually reviewed and merged by a human.
 It is ***highly recommended*** to also use [TagBot][tagbot], which automatically tags a release in your
 repository after the new release of your package is merged into the registry.
 
+### Requirements for packages in the General registry
+
+Packages registered in the General registry must be made in good faith.
+
+We do not accept packages or package registrations that:
+1. Are intentionally broken
+2. Are made in bad faith and/or with bad intentions
```
Nosferican, line 39: Maybe add examples?

README.md
```diff
@@ -30,6 +30,17 @@ requests need to be manually reviewed and merged by a human.
 It is ***highly recommended*** to also use [TagBot][tagbot], which automatically tags a release in your
 repository after the new release of your package is merged into the registry.
 
+### Requirements for packages in the General registry
+
+Packages registered in the General registry must be made in good faith.
+
+We do not accept packages or package registrations that:
+1. Are intentionally broken
```
Nosferican, line 38: Not sure what this means?

DilumAluthge: 🤷

DilumAluthge: Direct quote from Slack

DilumAluthge: I guess Seth's IsTwo.jl is an example

sbromberger: How about "obviously incorrect"?

README.md
DilumAluthge: Sure, if someone can come up with examples.

README.md
DilumAluthge: Again, quoting from Slack.

sbromberger: I don't like #3. It's too subjective.

README.md
DilumAluthge: I'll change it to "obviously and intentionally incorrect." Since this is documentation about intent.

StefanKarpinski: The problem was not that it was broken, it was that you clearly didn’t even mean for it to work. People will register packages that don’t work, we’re not going to be able to stop that.

DilumAluthge: I think "obviously and intentionally incorrect" covers that, right? It covers the case of authors not intending for their code to work correctly.

README.md
DilumAluthge: The whole thing is subjective 🤷

DilumAluthge: I'll take 3 out. Three people have disliked it.

README.md
sbromberger: How do we determine intent? That is, if I submit
```
"""
    istwo(x)

Return `true` if x is equal to two
"""
istwo(x::Number) = isapprox(x, 2)
```
does that pass muster?

DilumAluthge: Basically, it's at the discretion of the registry maintainers 🤷 

The main reason for documenting this is that now when I close a bad-faith PR, I can use these rules as justification.

DilumAluthge:
Okay, I think we now have a PR here that is very narrow.

Does anyone object to merging this PR in its current state?

DilumAluthge:
I will leave this open for a day, and then I will merge this if no one objects to merging it in its current state.

README.md
sbromberger: If it's at the discretion of the registry maintainers, then *put that in there*.

sbromberger: @StefanKarpinski has said in Slack that it is "absolutely at the discretion of the maintainers". I would suggest that we can scrap everything other reason and just use that.

README.md
Nosferican: Malicious backdoors?

README.md
```diff
@@ -30,6 +30,16 @@ requests need to be manually reviewed and merged by a human.
 It is ***highly recommended*** to also use [TagBot][tagbot], which automatically tags a release in your
 repository after the new release of your package is merged into the registry.
 
+### Requirements for packages in the General registry
+
+Packages registered in the General registry must be made in good faith.
```
sbromberger, line 35: "Packages registered..." should probably be "Pull requests for packages to be registered..." - since the criterion relates to the PR, not the package itself (unless that is what you're trying to convey, in which case we might want to clarify.)

Is it the pull request that should be made in good faith, the package that must be made in good faith, or both?

DilumAluthge:
Thanks everyone for your help on this.

Closing in favor of https://github.com/JuliaRegistries/General/pull/16234

PR #16214, opened 2020-06-11
Remove Org.jl

non-Jedi:
I had thought I would spend more time getting this package into a useful state, but it hasn't happened. No reason to take up a scarce 3-letter name for an unmaintained package. Nothing depends on this, and I am mostly certain it has no users.


fredrikekre:
https://github.com/JuliaRegistries/General/blob/master/README.md#how-do-i-remove-a-package-or-version-from-the-registry

PR #16234, opened 2020-06-12
Document the current policy: all decisions at the discretion of the registry maintainers

DilumAluthge:
This PR documents our current policy.

To be clear, this PR is not in any way creating, modifying, or removing policy. The policy documented in this PR is the policy that we currently use and have always been using.

The purpose of this PR is just to document how we do things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ registry, and code cannot be deleted.
 
 ## Registry maintenance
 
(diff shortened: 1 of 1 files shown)


sylvaticus:
I understand the meaning of this PR and I agree with it, but just you can't write "Different registry maintainers may choose to enforce different rules and requirements. Additionally, the same registry maintainer may use different criteria for different situations on a case-by-case basis."
That can be very badly interpreted, I'm not a native English speaker but I suggest that this should be somehow rephrased.

DilumAluthge:
For this PR, I have removed some potentially confusing text.

In a future PR, perhaps someone can make suggestions on how to rephrase it.

PR #16517, opened 2020-06-17
Limit GMP 6.2 to Julia 1.6

fingolfin:
See https://github.com/JuliaPackaging/Yggdrasil/pull/1159


odow:
Is it possible to get this merged ASAP? This has effectively broken most of the JuMP ecosystem using GLPK on mac.

fredrikekre:
Please use `RegistryTools.Compress.(load|save)` when modifying these files, you probably want:
```
["6-6.1"]
julia = "1-1.5"

["6.2-6"]
julia = "1.6"
```
Does GMP@6.1 not work on Julia 1.6? Is this still needed now that the version bump was reverted in JuliaLang/julia?

giordano:
> Edit: I guess this patch is more about forcing GMP@6.1 on pre-1.6 Julia?

Correct, that's the main point

fingolfin:
The GMP version in Julia and GMP *must* match. So if GMP 6.2 was reverted from Julia master (I wasn't aware) then actually the version restriction for GMP 6.1 should include Julia 1.6, until the day Julia master switches back to GMP 6.2

fingolfin:
And so GMP 6.2 should be Julia >= 1.7 for now

odow:
6.2 was reverted on julia `master`: https://github.com/JuliaLang/julia/pull/36328

Can't we just yank the `GMP_jll` tag for `v6.2` and revisit this issue when/if GMP upgraded on Julia?

fredrikekre:
We can merge https://github.com/JuliaRegistries/General/pull/16517#issuecomment-645929185 for now. Presumably the issues will be resolved before 1.6.

fingolfin:
Updated to match @fredrikekre's suggestion (thanks)

PR #16663, opened 2020-06-20
Split SnoopCompile into subpackages

timholy:
We recently split SnoopCompile into sub-packages, primarily to ensure that a tool that makes measurements of system state does not introduce large perturbations of that state (https://github.com/timholy/SnoopCompile.jl/issues/95). Because Registrator doesn't seem to let you register multiple packages simultaneously (https://github.com/JuliaRegistries/Registrator.jl/issues/298) and I didn't want to wait for a dependency-chain of registrations to be merged, I instead used LocalRegistry.jl to generate this combined PR.

Closes #16649
Closes #16650


DilumAluthge:
@timholy Can you make a new PR that changes all of the URLs to be `https` URLs?

timholy:
See #16812

PR #16876, opened 2020-06-24
Allow julia 1.0 for newer releases of XML2_jll

nicoleepp:
Ran into https://github.com/JuliaPackaging/Yggdrasil/issues/1203 

Versions of XML2_jll allow julia 1 since [release v2.9.9+2](https://github.com/JuliaBinaryWrappers/XML2_jll.jl/blob/XML2-v2.9.9+2/Project.toml).


nicoleepp:
Seems like `+` in semver for packages is not supported. Closing

PR #17272, opened 2020-07-01
Delete v0.5.11 due to error

chakravala:
The commit for v0.5.11 doesn't exist anymore and was overwritten by accident, so it must be removed from the registry.

I was talking on the phone and accidentally messed up my commits


chakravala:
@fredrikekre

DilumAluthge:
Instead of deleting it, just yank it.

PR #18713, opened 2020-07-30
Adjust PrettyPrint bound

andreasnoack:
The existing bound wasn't correct as the newly released version 0.2 of PrettyPrint broke NameResolution.

cc @thautwarm


thautwarm:
Thanks.

PR #18872, opened 2020-08-03
Add julia v1.5.0

vdayanand:


fredrikekre:
What are those entries used for?

StefanKarpinski:
There are some tools that use them. Pkg itself doesn't.

fredrikekre:
Maybe those tools can keep this list then? Or just use Elliot's JSON file?

StefanKarpinski:
If we're going to do that then I think we should just delete Julia from the registry entirely.

PR #18959, opened 2020-08-04
Allow LibGit2_jll 1.0 on Julia 1.6

nalimilan:
Needed by https://github.com/JuliaLang/julia/pull/35233.

@giordano Is there a particular reason why you required Julia 1.7 when you registered the package (as mentioned [here](https://github.com/JuliaPackaging/Yggdrasil/pull/638#issuecomment-658317060))?

### DIFF
--- a/L/LibGit2_jll/Compat.toml
+++ b/L/LibGit2_jll/Compat.toml
@@ -5,4 +5,4 @@ julia = "1.3.0-1"
 julia = "1"
 
 ["1"]
-julia = "1.7"
+julia = "1.6"


giordano:
> @giordano Is there a particular reason why you required Julia 1.7 when you registered the package (as mentioned here)?

Yes, because Julia v1.6 still doesn't support LibGit v1

nalimilan:
Ah, OK, I thought I needed to allow 1.6 before merging the PR, but indeed I just tested it and it's not the case. So let's keep this PR as a draft until JuliaLang/julia#35233 has been merged.

fredrikekre:
bump

nalimilan:
Since 1.0.1 we allow any Julia version above 1 AFAICT: https://github.com/JuliaRegistries/General/commit/8d478c3be64784cc266ac1ab6a681fa1cfb55a43

fredrikekre:
So this is not relevant? Or should 1.0.0 also be allowed on earlier Julia?

nalimilan:
Yeah, I guess. See https://github.com/JuliaRegistries/General/pull/18959.

giordano:
Is LibGit2 v1.0 API/ABI compatible with the old ones?

nalimilan:
No.

PR #18991, opened 2020-08-04
Change repo URLs to https://GitHub.com to break pattern matching

StefanKarpinski:
Attempting to unfuck this situation: https://github.com/JuliaLang/Pkg.jl/issues/1942.

PR #19004, opened 2020-08-05
Use custom shell to tidy up AutoMerge

SaschaMann:
The multiline string of the `run` key will be saved to a file. The filename will be inserted in place of the `{0}`.

ref. https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#custom-shell

~(needs to be tested, but I don't know how)~ Seems to run as expected on this PR

@DilumAluthge


DilumAluthge:
I'll merge this once we've reverted #18991

PR #19436, opened 2020-08-13
Fix some compat entry in Arb_jll

thofma:
@giordano


giordano:
I'm a bit confused by what this PR wants to achieve.  Can you please describe what's your goal?

thofma:
At the moment the two latest `Arb_jl` versions (2.18.1 and 2.18.2) require FLINT_jll with version _exactly_ 2.6.2. But now we have FLINT_jll 2.6.3. So they cannot be installed at the same time. I noticed that `_jll` packages usually don't have compat entries, so I just delete it.

giordano:
Ok, so the old versions still need FLINT 0.0.2, but the new ones are compatible with any version of FLINT?

thofma:
Yes.

PR #19557, opened 2020-08-15
New version: SnoopCompileBot 2

aminya:


aminya:
@DilumAluthge Could you help with this? I cannot seem to be able to make this work

DilumAluthge:
What are you trying to accomplish?

If you want to register a new version of a package, just use Registrator (either the GitHub App or the web interface at JuliaHub.com)

DilumAluthge:
Now I've figured out what's going on. You want to remove the old "SnoopCompileBot" package (currently a subdirectory of SnoopCompile.jl) and replace it with https://github.com/aminya/SnoopCompileBot.jl

Unfortunately, "SnoopCompileBot" already has multiple registered versions in the registry, and those `git-tree-sha1` entries point to git trees that live in the `git@github.com:timholy/SnoopCompile.jl.git`.

If we change the URL of "SnoopCompileBot" to point to the https://github.com/aminya/SnoopCompileBot.jl repo, this will break everything, because the required `git-tree-sha` will not be found in the https://github.com/aminya/SnoopCompileBot.jl repos.

In short, unless I am mistaken, it is not possible to make this change, because it would be equivalent to deleting a package from the General registry, which we do not allow.

You will need to rename https://github.com/aminya/SnoopCompileBot.jl to something else (e.g. SnoopCompileBot2) and register it with the new name.

cc: @aminya 
cc: @timholy 

cc: @KristofferC 
cc: @StefanKarpinski

aminya:
The old package does not work anymore because this was a package that was used mostly in GitHubActions, and I had forgotten to pin the version in the workflow file. So always the latest version is used. But recently SnoopCompile was published without SnoopCompileBot in it, and so all the packages are broken!

I just want to fix it using the new API.

DilumAluthge:
Can you point me to an example of a failed GitHub Actions run? This will help determine what the best course of action is.

aminya:
Here it is one of them for example: https://github.com/JuliaPlots/Plots.jl/pull/2898#issuecomment-673365763 Plots.jl needs this ASAP.

DilumAluthge:
I see. So the SnoopCompileBot functionality has been completely removed from SnoopCompile.

Unfortunately there is no way to fix this in the registry. You will need to register `SnoopCompileBot2` as a new package. Then, you will need to make pull requests to update the GitHub Actions workflow files.

DilumAluthge:
If I am incorrect about this, @StefanKarpinski or @KristofferC will be able to correct me.

aminya:
This is a side effect that I did not know about. Monorepo caused many issues and headaches for us, and we decided to move away from it. But now I am seeing that using a monorepo has a punishment that is not fixable!

fredrikekre:
Just mirror the trees in the new repo.

DilumAluthge:
> Just mirror the trees in the new repo.

Could you explain how to do that?

fredrikekre:
Check out the old trees in the new repo and then tag with the same version number for example.

aminya:
Thanks! That is a clever solution 🎉  I'll do that this week. I am quite busy right now.

PR #19731, opened 2020-08-18
remove upper bound on MozillaCACerts_jll

visr:
Since it doesn't use semver.

Is already adjusted in LibCURL with this commit: https://github.com/JuliaWeb/LibCURL.jl/commit/e25de0d03133f76fc51c9f825c560819290f4d0a#diff-910a7b3ca0075d545dcec45cb7335ca9

### DIFF
--- a/L/LibCURL/Compat.toml
+++ b/L/LibCURL/Compat.toml
@@ -16,5 +16,5 @@ BinaryProvider = "0.5"
 
 ["0.6-0"]
 LibCURL_jll = "7.68.0-7"
-MozillaCACerts_jll = "2020.0.0-*"
+MozillaCACerts_jll = ">= 2020"
 julia = "1.3.0-1"


giordano (approved):

GunnarFarneback:
I think this PR needs to be reverted.
```
(@v1.5) pkg> up
   Updating registry at `~/.julia/registries/CVRegistry`
   Updating git-repo `git@gitlab.contextvision.se:Julia/CVRegistry.git`
   Updating git-repo `https://github.com/JuliaLang/Example.jl.git`
   Updating git-repo `https://github.com/JuliaGPU/CUDAnative.jl.git`
   Updating git-repo `https://github.com/JuliaGPU/GPUArrays.jl.git`
ERROR: ArgumentError: invalid version range: ">= 2020"
```

giordano:
Should this be checked by Travis?

DilumAluthge:
It should have been caught by the Travis CI on the General registry, yes. But Travis passed on this PR. I'm not sure how it slipped through.

@fredrikekre

DilumAluthge:
I think we'll need to add a new test to catch this.

fredrikekre:
Why was this PR even needed? The entry that is there already means `>=2020`?

fredrikekre:
> Should this be checked by Travis?

> It should have been caught by the Travis CI on the General registry, yes. But Travis passed on this PR. I'm not sure how it slipped through.

The tests I added for exactly this was removed by this commit: https://github.com/JuliaRegistries/RegistryCI.jl/commit/9c6f8fc2ce7393695c9040cb0ec10bd1adc1ba5f#diff-c921fae36efa43624d14d15a33bc9b24

visr:
Sorry about this. So it's the space between `>=` and `2020` that broke it? If so we should fix it in LibCURL.jl as well, I copied it from there.

> The entry that is there already means >=2020?

Yeah I see now. I interpreted `2020.0.0-*` as any build number, but that doesn't make any sense.

fredrikekre:
No, everything is fine. `>= 2020` in the project file results in  `2020.0.0-*` in the registry, it has never been incorrect.

PR #20419, opened 2020-08-28
Fix Deps due to compat override on DiffEqBase DataStructures

ChrisRackauckas:
Fixes deps to match https://github.com/JuliaRegistries/General/pull/20415


fredrikekre:
Please just fix the Compat, don't add dependencies to versions.

ChrisRackauckas:
We have to because DataStructures's breaking change.

PR #20439, opened 2020-08-28
Load and compress TOML files after various manual edits.

fredrikekre:


L/LLVM_jll/Compat.toml
```diff
@@ -6,6 +6,3 @@ julia = "1.4"
 
 [9]
 libLLVM_jll = "9.0.1"
-
-[10]
-libLLVM_jll = "10.0.0"
```
fredrikekre, line 11: This looks strange, but there is no LLVM_jll version 10 so.

DilumAluthge (approved):

PR #20453, opened 2020-08-28
Fix some compat entry in Singular_jll

thofma:
I removed the compat bound here https://github.com/JuliaPackaging/Yggdrasil/pull/1528, but this does not update the compat entry here.

This is the same as https://github.com/JuliaRegistries/General/pull/19436.

@giordano

### DIFF
--- a/S/Singular_jll/Compat.toml
+++ b/S/Singular_jll/Compat.toml
@@ -1,3 +1,2 @@
 [4]
-FLINT_jll = "2.6.2"
 julia = "1"


giordano:
But 4.1.3+0 and 4.1.3+1 do depend on that package, this PR would effectively break those versions

thofma:
No, they won't break. FLINT 2.6.2 broke things and 2.6.3 is the fix.

giordano:
Yeah, ok, these are compat bounds, I was confused with dependencies, removing it would be breaking

thofma:
Yes, indeed. It is just about the compat bounds.

PR #20779, opened 2020-09-03
New package: SCS_GPU_jll v2.1.2+0

jlbuild:

* Registering JLL package SCS_GPU_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/SCS_GPU_jll.jl
* Version: v2.1.2+0


kalmarek:
[noblock]
loading of the package will fail, as the package dlopens `libscsgpuindirect.so` which is linked against `libcublas.so` (provided e.g. by `CUDA_jll`). However `CUDA_jll` is not in the deps of the package this will be hidden by `@require CUDA_jll="..."` in the `__init__()` of `SCS.jl`.

reason: `CUDA_jll` is too heavy to ship it to everyone using `SCS.jl`

PR #21214, opened 2020-09-11
Relax compatibility of DocStringExtensions for old Documenter versions

iamed2:
We are currently using Documenter v0.21.5 for docs for a package, while another dependency of the package is using DocStringExtensions 0.8. The retrocapping caused our documentation jobs to start breaking.

### DIFF
--- a/D/Documenter/Compat.toml
+++ b/D/Documenter/Compat.toml
@@ -5,7 +5,7 @@ julia = "0.6-1"
 Compat = "0.62-2"
 
 ["0-0.22.3"]
-DocStringExtensions = "0.2.0 - 0.7"
+DocStringExtensions = "0.2.0 - 0.8"
 
 ["0.19.3-0.19"]
 Compat = "0.70-2"


DilumAluthge (approved):

iamed2:
Thank you!

fredrikekre:
Why wasn't this bound to 0.8 in the capping? is it because https://github.com/JuliaRegistries/General/pull/21214/files#diff-20a35dc9970e804d703c758b68d1887bR23 ?

A bit strange that https://github.com/JuliaRegistries/General/pull/21214/files#diff-20a35dc9970e804d703c758b68d1887bR23 exists now, perhaps that should be relaxed too then.

DilumAluthge:
It was initially capped to 0.7 because of the monotonicity of the caps.

https://github.com/JuliaRegistries/General/blob/47da5c9f7d5ad184e6f83ccb3cc92f25853b5be7/D/Documenter/Compat.toml#L22-L26

Since 0.22.4 and 0.22.5 only supported up to 0.7, that was the ceiling for the cap that was applied to earlier versions.

PR #21273, opened 2020-09-12
Optim bounds once again

pkofod:
@fredrikekre I think maybe this is correct. I applied your diff, but now v1+ isn't installed on Julia v1.5. Can you verify this might be what I was going for? I just didn't want versions below v1 of Optim to be installed on Julia lower than v1.2. Thanks!


DilumAluthge:
@pkofod Can you rebase on master and squash?

@fredrikekre Can you review this?

DilumAluthge:
For what it's worth, I think that this is correct. That being said, it would be good for Fredrik to review it as well, just to be safe.

First column is the Julia version, second column is whether or not Optim 0.17 is installable on that Julia version, third column is whether or not Optim 1 is installable on that Julia version:
```julia
julia> for minor in [0, 1, 2, 3, 4]
       for patch in [0, 1, 2]
       ranges = Pkg.Types.VersionRange.(["1.0-1.1", "1.2-1"])
       ver = VersionNumber(1, minor, patch)
       println(ver, " ", ver in ranges[1], " ", ver in ranges[2])
       end
       end
1.0.0 true false
1.0.1 true false
1.0.2 true false
1.1.0 true false
1.1.1 true false
1.1.2 true false
1.2.0 false true
1.2.1 false true
1.2.2 false true
1.3.0 false true
1.3.1 false true
1.3.2 false true
1.4.0 false true
1.4.1 false true
1.4.2 false true
```

fredrikekre:
Sorry, I think I misunderstood what you wanted in the previous PR, or maybe I was tired. I can push a rebase, but why do you also want to limit Optim 0.17 to Julia < 1.2?

fredrikekre:
I think is the diff we need (compared to master) now after Optim 1.2:
```diff
diff --git a/O/Optim/Compat.toml b/O/Optim/Compat.toml
index 56eea5a..cbe61fd 100644
--- a/O/Optim/Compat.toml
+++ b/O/Optim/Compat.toml
@@ -80,11 +80,5 @@ StatsBase = "0.29-0.33"
 
 [1]
 FillArrays = "0.6.2-0.9"
-
-["1-1.1"]
-NLSolversBase = "7.7-7"
-julia = "1.2"
-
-["1.2-1"]
 NLSolversBase = "7.7.0-7"
 julia = "1.2.0-1"
```

PR #21615, opened 2020-09-19
SnoopCompileBot URL and git sha update

aminya:
Updated the URL and git sha

Based on @fredrikekre's suggestion in https://github.com/JuliaRegistries/General/pull/19557#issuecomment-674461349, I created branches resembling the original code: 
https://github.com/aminya/SnoopCompileBot.jl

### DIFF
--- a/S/SnoopCompileBot/Package.toml
+++ b/S/SnoopCompileBot/Package.toml
@@ -1,4 +1,4 @@
 name = "SnoopCompileBot"
 uuid = "1d5e0e55-7d74-4714-b8d8-efa80e938cf7"
-repo = "https://github.com/timholy/SnoopCompile.jl.git"
+repo = "https://github.com/aminya/SnoopCompileBot.jl.git"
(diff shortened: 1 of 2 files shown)


fredrikekre (changes requested):
The tree shas should not change.

aminya:
So how can I do this then?

fredrikekre:
If you check out the correct trees you shouldn't have to change anything here (except the URL)

aminya:
If I understand correctly, you are suggesting that I rewrite the history of my master, so it is built from the @timholy's repository. 
What I did here was to just create branches for the old versions, which means the git sha would become different.

fredrikekre:
The tree sha should not change (the commit sha will be different).

aminya:
@fredrikekre I am not still clear what to do. Could you please give me a git command to try?

aminya:
Closed in favor of #22253

PR #21627, opened 2020-09-19
Bring Singular_jll back to previous version

giordano:
Ref: https://github.com/oscar-system/Singular.jl/issues/271.  CC: @thofma


fredrikekre:
Could also just yank the previous one?

giordano:
Also, yes

PR #21903, opened 2020-09-24
Fix LibGit2_jll compat bounds

nalimilan:
Replaces https://github.com/JuliaRegistries/General/pull/18959.


giordano:
I don't think this is correct, libgit2 v1.0 has a different soversion than older versions, my understanding is that trying to load v1.0 in old Julia versions would fail badly

nalimilan:
Yeah, I wonder why the bounds were removed when we tagged 1.0.1. I've pushed a new commit which only allows LibGit2 0.x before Julia 1.6, and 1.x from 1.6 onwards.

nalimilan:
Though before 1.0 the LibGit2 API broke a few things every six months so it's hard to do anything with it...

giordano:
I can see only one JLL package in the registry that currently depends on `LibGit2_jll` and I higly doubt many people are using it, so I don't really expect anyone bumping into this issue, but it's better to keep these compatibilities in sync with what versions are used in Julia

nalimilan:
Yes, from LibGit2 1.0 and Julai 1.6 on I think it's the way to go. For earlier Julia versions that would require packaging LibGit2 versions prior to 0.28, which would take quite a bit of work.

DilumAluthge:
@nalimilan Needs a rebase?

nalimilan:
Done.

nalimilan:
I don't understand the failure:
```julia
LibGit2_jll: Test Failed at /home/travis/.julia/packages/RegistryCI/4vNNv/src/registry_testing.jl:167

  Expression: compressed == compat

   Evaluated: Dict("1" => Dict("julia" => "1.6"),"0" => Dict("julia" => "1-1.5"),"1.1-1" => Dict("JLLWrappers" => "1.1.0-1")) == Dict{String,Any}("0" => Dict{String,Any}("julia" => "1-1.5"),"1.0.0-1.0.1" => Dict{String,Any}("julia" => "1.6"),"1.1-1" => Dict{String,Any}("JLLWrappers" => "1.1.0-1","julia" => "1.6"))
```

DilumAluthge:
@nalimilan @giordano Is this good to merge now?

nalimilan:
I think so, thanks.

DilumAluthge:
Alright, let me know if this breaks anything, and we can revert it.

PR #21996, opened 2020-09-25
fix IS version

jd-lara:
The previous registration request I implemented incorrectly sem-ver and tagged a version that breaks packages with dependencies on InfrastrucutreSystems. The tag should have been `0.16.0` and `0.15.5` shouldn't be in the registry given that it isn't compatible with packages depending on `0.15.X`.


fredrikekre:
You can yank the 15.5 release if this is still needed? That will prevent it from beeing installed.

PR #22298, opened 2020-10-02
Moving repo to GitHub organization

eford:
Do I need to do anything else?

### DIFF
--- a/R/RvSpectML/Package.toml
+++ b/R/RvSpectML/Package.toml
@@ -1,3 +1,3 @@
 name = "RvSpectML"
 uuid = "1f61ac2c-3a1c-440a-925a-2707197041c8"
-repo = "https://github.com/eford/RvSpectML.jl.git"
+repo = "https://github.com/RvSpectML/RvSpectML.jl.git"


fredrikekre:
Ideally you would use GitHubs transfer functionality such that the old URL redirects to the new one. Any reason you didn't do that?

giordano:
In the new repository the history starts from what I understand was the last revision of the old repository, so I'm not sure old versions will be installable after this pull request?

eford:
> Ideally you would use GitHubs transfer functionality such that the old URL redirects to the new one. Any reason you didn't do that?

@fredrikekre Thanks.  Yes, I didn't know that existed (and a google search returned someone suggesting that submiting a PR here).  
I'd be happy to delete the one I created and try the GitHub transfer.  After I do that, will the Julia registries automatically update?  If not, should I update this PR or close it and start a new one?

@giordano Thanks the warning.  In this case, it's a small group of us preparing this software in advance of data going public.  So we haven't advertised it yet and it wouldn't matter if the previous versions were to be uninstallable.

eford:
> In the new repository the history starts from what I understand was the last revision of the old repository, so I'm not sure old versions will be installable after this pull request?

@giordano Thanks the warning.  In this case, it's a small group of us preparing this software in advance of data going public.  So we haven't advertised it yet and it wouldn't matter if the previous versions were to be uninstallable.

giordano:
Ok, however the [General registry is committed to *never* remove any package or any version](https://github.com/JuliaRegistries/General/blob/654bc9ef2660500e91943001ca26e32780d338dd/README.md#how-do-i-remove-a-package-or-version-from-the-registry), even for packages supposed to be used by few people.  As suggested by Fredrik, the easiest way forward is to use GitHub functionality to transfer the repository to the new organisation.  After that we can merge this pull request.

eford:
@fredrikekre & @giordano :  Ok.  I transfered ownership and github says:
```
@eford has transferred the eford/RvSpectML.jl repository to RvSpectML/RvSpectML.jl.

No further action is required at this time.
```
Please let me know if there's anything else I should do.

eford:
FYI, It might be useful to put the QA for "How do I transfer a package to an organization or another user?" in the Julia manual, so it's easier for people to find.  
And thanks for your helpful feedback.   As a long-time scientific programmer, I can get overwhealmed trying to make sense of all the the packaging/cloud stuff, so I appreciate the helpful and patient responces from people like you.

giordano:
Ok, looks good now!

PR #22358, opened 2020-10-03
Change package URL for Arrow.jl

quinnj:


fredrikekre:
Any reason this isn't a transfer? Are the old trees available in the new repo?

quinnj:
I'd didn't realize a transfer would have been easier, but I did replay all my new commits onto the existing package tree and then force pushed master to the new repo. That should ensure all the old trees are still intact, right?

giordano:
As a check, are you able to install old versions of the package using this branch of the registry?

quinnj:
Yes, I am able to

giordano:
Ok, then let's go

PR #22930, opened 2020-10-14
Update Package.toml

arturgower:
Changed URL from user to organization.


arturgower:
In case it helps, here is the package with the new url: https://github.com/JuliaWaveScattering/EffectiveWaves.jl
and I am sadly the only contributor

PR #22954, opened 2020-10-14
Make Artifacts.jl installable on Julia versions 1.0-1.3 as well

iamed2:
Fixes https://github.com/invenia/LibPQ.jl/issues/202

This allows newer Yggdrasil JLLs to continue to be installed by Pkg on older Julia versions.


staticfloat:
Hmmm..... yes this should be fine.

iamed2:
> We may eventually not support this anymore

The reason this came up is an existing package's version became uninstallable on CI, since it depended on OpenSSL_jll and the dependency on Artifacts.jl was added in a build version of OpenSSL_jll (It was not restricted down to 1.1.1+5, Pkg picked 1.1.1+6 and then couldn't resolve because it needed Artifacts on Julia <1.3. This is the future situation which I think it is very important to avoid. 

I think extra care needs to be paid to build releases as well, because Pkg doesn't seem to handle those properly.

staticfloat:
> This is the future situation which I think it is very important to avoid.
> 
> I think extra care needs to be paid to build releases as well, because Pkg doesn't seem to handle those properly.

This is a great point, and one we will need to be careful to avoid in the future.

PR #23207, opened 2020-10-19
New version: GMP_jll v6.2.0+3

jlbuild:

* Registering JLL package GMP_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/GMP_jll.jl
* Version: v6.2.0+3


fingolfin:
This failed to install because the package is for Julia >= 1.6 while the test bot runs Julia 1.5.2. Otherwise, I think this looks correct and can be merged.

PR #23334, opened 2020-10-20
Update julia to v1.5.2

vdayanand:


fredrikekre:
Can we do https://github.com/JuliaRegistries/General/pull/18872#issuecomment-668341969 instead?

DilumAluthge:
I agree. Now that we have https://julialang-s3.julialang.org/bin/versions.json, I don't think we need to have Julia in the registry.

PR #23421, opened 2020-10-22
Un-do registrator bungled compat bounds

staticfloat:
I'm not sure why Registrator assigned Julia 1.6+ to this.


fredrikekre:
Because that is what the compat bounds say: https://github.com/JuliaBinaryWrappers/GMP_jll.jl/blob/d2e7dfea280486796ead411980e3a08caea7833f/Project.toml#L13

fingolfin:
See https://github.com/JuliaPackaging/Yggdrasil/pull/1968

PR #23869, opened 2020-10-29
Enable TagBot issue comment triggers

christopher-dG:
Needs RegistryCI 4.3.0 to be released first. I'll update the `.ci` project and manifest once that's in.

### DIFF
--- a/.github/workflows/TagBotTriggers.yml
+++ b/.github/workflows/TagBotTriggers.yml
@@ -1,0 +1,16 @@
+name: TagBot Triggers
+on:
+  schedule:
+    - cron: 0 0 * * *
+  pull_request:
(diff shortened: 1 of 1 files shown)


DilumAluthge:
> I'll update the `.ci` project and manifest once that's in.

No need. Once RegistryCI 4.3.0 is registered, just go to:
- https://github.com/JuliaRegistries/General/actions?query=workflow%3Aremember_to_update_registryci

And run the workflow manually. It will open a PR to update the manifest.

DilumAluthge:
The CI manifest was updated in: https://github.com/JuliaRegistries/General/pull/23877

So this should be good to go.

cc: @christopher-dG

PR #23938, opened 2020-10-30
Add TagBot token to automerge

christopher-dG:
Needs RegistryCI 4.3.3 (https://github.com/JuliaRegistries/RegistryCI.jl/pull/289).

I also bumped the Julia version to a supported one, but really we could just get rid of the `setup-julia` step.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.5.2'
+          - 'nightly'
         os:
           - ubuntu-latest
         arch:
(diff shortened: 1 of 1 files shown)


DilumAluthge:
The `import PkgName` is broken on 1.5.2 for some packages. But it's fixed in nightly, so let's use that instead.

christopher-dG:
That seems... unwise. But it's your call. This should now wait for https://github.com/JuliaRegistries/RegistryCI.jl/pull/290 too.

DilumAluthge:
We can use 1.5.1 instead.

DilumAluthge:
I haven't been able to figure out what changed in 1.5.1 to 1.5.2 that made `import PkgName` break for some packages.

And since the breakage does not happen on Julia nightly, I'm not particularly motivated to investigate it further.

christopher-dG:
Hm, that is really odd. But yeah, hopefully 1.6 will come soon ™️

DilumAluthge:
Alright should be all set now.

PR #24055, opened 2020-11-02
Run the TagBot trigger cron every four hours

DilumAluthge:
Currently, we have thousands of packages running a TagBot cron once per day. We are getting rid of that, so I feel less guilty about an hourly TagBot cron on a single repository (General registry)

### DIFF
--- a/.github/workflows/TagBotTriggers.yml
+++ b/.github/workflows/TagBotTriggers.yml
@@ -1,7 +1,7 @@
 name: TagBot Triggers
 on:
   schedule:
-    - cron: 0 0 * * *
+    - cron: 0 * * * *
   pull_request:
     types:
       - closed
--- a/.github/workflows/TagBotTriggers.yml
+++ b/.github/workflows/TagBotTriggers.yml
@@ -1,7 +1,7 @@
 name: TagBot Triggers
 on:
   schedule:
-    - cron: 0 * * * *
+    - cron: 0 */4 * * *
   pull_request:
     types:
       - closed


christopher-dG:
I think hourly might be a bit excessive. How about like every four hours or something? And regardless of how often we should definitely update the comment message for cron jobs to avoid confusing people with the number of comments that will come out. Maybe we can wait until I finish the auto fixup feature before doing this too? I would feel better about spamming people if they had an easier way to make it stop.

DilumAluthge:
> I think hourly might be a bit excessive. How about like every four hours or something? And regardless of how often we should definitely update the comment message for cron jobs to avoid confusing people with the number of comments that will come out. Maybe we can wait until I finish the auto fixup feature before doing this too? I would feel better about spamming people if they had an easier way to make it stop.

For sure. So let's:
1. [ ] Do it every 4 hours. Can you push that change to this branch?
2. [ ] Not merge this until you have updated the comment message
3. [ ] Not merge this until you have the auto fix up

.github/workflows/TagBotTriggers.yml
```diff
@@ -1,7 +1,7 @@
 name: TagBot Triggers
 on:
   schedule:
-    - cron: 0 0 * * *
+    - cron: 0 * * * *
```
christopher-dG, line 4:
```suggestion
    - cron: 0 */4 * * *
```

DilumAluthge:
@christopher-dG Any update on this? The daily cron jobs seem to be failing frequently, which makes me worry - if both the PR trigger job and the cron job failed, we increase the probability that a package will miss a tag before the 3 day lookback period ends.

christopher-dG:
Sorry, it kind of slipped my mind after I put the PR up. I'll try to get it finished tomorrow.

christopher-dG:
Sorry again, I got kinda carried away with what I was doing today. Tomorrow for sure! 😅

DilumAluthge:
@christopher-dG is this good to merge now?

PR #24161, opened 2020-11-04
Restrict LibCURL_jll 7.68+ to v1.6+

staticfloat:
Because it's actually v7.66 that ships with Julia 1.5, let's restrict all newer versions to 1.6+


giordano:
Ref: https://github.com/JuliaPackaging/Yggdrasil/issues/2051

staticfloat:
Dilum, I'm requesting your review just because I'm not 100% certain I've gotten my hand-wrought compressed bounds quite right.

DilumAluthge:
The version range looks good.

```julia
julia> v"1.5.0" in Pkg.Types.VersionRange("1.6.0-1")
false

julia> v"1.5.1" in Pkg.Types.VersionRange("1.6.0-1")
false

julia> v"1.5.2" in Pkg.Types.VersionRange("1.6.0-1")
false

julia> v"1.6.0" in Pkg.Types.VersionRange("1.6.0-1")
true

julia> v"1.6.1" in Pkg.Types.VersionRange("1.6.0-1")
true

julia> v"1.6.2" in Pkg.Types.VersionRange("1.6.0-1")
true

julia> v"1.7.0" in Pkg.Types.VersionRange("1.6.0-1")
true

julia> v"1.7.1" in Pkg.Types.VersionRange("1.6.0-1")
true

julia> v"1.7.2" in Pkg.Types.VersionRange("1.6.0-1")
true
```

And it roundtrips under decompression/recompression (Travis checks this). So it should be good to go!

DilumAluthge (approved):

StefanKarpinski:
This broke Windows CI for Downloads because we did NOT ship libcurl on Windows and libcurl < 7.71 or so is buggy on Windows, at least the way that Downloads uses it. So I'm going to propose reverting this.

PR #24170, opened 2020-11-04
Fix some compat bounds for LibCURL_jll.jl

DilumAluthge:
I think I initially missed a mistake in https://github.com/JuliaRegistries/General/pull/24161. This should fix it.


DilumAluthge:
cc: @staticfloat

staticfloat:
There is no 7.67, so this should be fine.

DilumAluthge:
Ahhhhh. That's why this PR is failing under roundtrip.

PR #24188, opened 2020-11-05
Update Package.toml

ScottPJones:
ShortStrings.jl was moved by the owner to JuliaString


DilumAluthge:
Done. You should be able to register new versions of the package now.

cc: @ScottPJones @xiaodaigh

PR #24265, opened 2020-11-06
New version: LLVM_full_jll v11.0.0+4

jlbuild:

* Registering JLL package LLVM_full_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/LLVM_full_jll.jl
* Version: v11.0.0+4


staticfloat:
@dilum @fredrikekre so here's an interesting case.  We registered a new version that is `v11.0.0+4` that has a new compat entry that previous `11.0.0+X` releases didn't have.  I'm fine with us retroactively changing the compat entries of the previous `11.0.0+X` releases, but apparently the compressor gets confused here and doesn't properly limit the previous compat entries.  For the record, what it generated was this:

```
[11]
JLLWrappers = "1.1.0-1"
julia = "1.6.0-1"

[8-11]
julia = "1"
```

PR #24654, opened 2020-11-15
Create registrator_issues.yml

DilumAluthge:


fredrikekre:
What is this?

DilumAluthge:
I'm experimenting with a way of running Registrator without a server. The workflow would be:
1. User opens an issue on the General registry with a title such as `register MyOrganization/MyPackage`
2. We check if the user is either (a) the owner of the repo, or (b) a public member of the organization that owns the repo.
3. If the user meets either of the above conditions, we make the registration.

It's still experimental, so AutoMerge is will not merge any PRs made from this.

DilumAluthge:
Example PR is here: https://github.com/JuliaRegistries/General/pull/24658

PR #24908, opened 2020-11-18
Change RobotZoo url

bjack205:
Changing ownership of RobotZoo from my personal account to the account of my research lab.

### DIFF
--- a/R/RobotZoo/Package.toml
+++ b/R/RobotZoo/Package.toml
@@ -1,3 +1,3 @@
 name = "RobotZoo"
 uuid = "74be38bb-dcc2-4b9e-baf3-d6373cd95f10"
-repo = "https://github.com/RobotExplorationLab/RobotZoo.jl.git"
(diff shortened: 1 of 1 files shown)


R/RobotZoo/Package.toml
```diff
@@ -1,3 +1,3 @@
 name = "RobotZoo"
 uuid = "74be38bb-dcc2-4b9e-baf3-d6373cd95f10"
-repo = "https://github.com/bjack205/RobotZoo.jl.git"
+repo = "https://github.com/RobotExplorationLab/RobotZoo.jl.git"
```
rejuvyesh, line 3: Seems like a typo. Should be `Robotic`?

DilumAluthge: Bump @bjack205 

Seems like the new URL should be `https://github.com/RoboticExplorationLab/RobotZoo.jl`

bjack205: ah, yes, sorry about that. It's now fixed.

PR #24916, opened 2020-11-19
New version: Libtask_jll v0.4.0+0

jlbuild:

* Registering JLL package Libtask_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/Libtask_jll.jl
* Version: v0.4.0+0


devmotion:
[noblock]

This version of Libtask_jll is only compatible with Julia 1.3: https://github.com/JuliaPackaging/Yggdrasil/blob/a9bd57bc8e0899c26b1a1432cda9689f5c100ef7/L/Libtask/build_tarballs.jl#L68 I assume this causes the automerge problems?

fingolfin:
@devmotion BTW have you already tested whether it really works there? For libcxxwrap_julia_jll, our 1.3 binary does not actually work on macOS (see discussion on https://github.com/JuliaPackaging/Yggdrasil/pull/2106)

devmotion:
No, I was about to check it when I realized that it is not available in the registry. BTW that's also why I haven't opened any of the subsequent PRs to Yggdrasil yet.

fingolfin:
You can test it already now w/o needing to wait for the registry: you can `add URL` the JLL repository

devmotion:
Thanks for the tip! I just tested it and all tests passed :slightly_smiling_face:

devmotion:
I noticed that it is even possible to load it with Julia > 1.3. I would have assumed that this would error but instead Pkg just prints
```julia
julia version requirement for package `Libtask_jll [3ae2931a]` not satisfied
```
Maybe it's since I added the JLL repo manually?

PR #25252, opened 2020-11-24
Fix compat typo

kescobo:
I accidentally set compat to

```
[compat]
DataFrames = "0.19, 0.20, 0.21, 22"
```

When I tagged the last version. Sorry for the complication :facepalm:


kescobo:
Do I also need to do something on my repo end? I already updated the `Project.toml`, but do I need to change which commit that General points to?

M/Microbiome/Versions.toml
```diff
@@ -29,4 +29,4 @@ git-tree-sha1 = "a89979a590578dd9dfb372757522dc881a84db76"
 git-tree-sha1 = "274452475b32b4dd54f9abb6e2297e88d5fe9004"
 
 ["0.6.5"]
-git-tree-sha1 = "4e901f0569d9684cc02130d489098cca084848d4"
+git-tree-sha1 = "8653f05a1a4ee373cf13792833801b6a47e98876"
```
fredrikekre, line 32: You should not update this

kescobo: Even though that sha points to something with a `Project.toml` saying it's compatible with DataFrames v22?

fredrikekre: Yes, this is immutable. You can release 0.6.6 with the fix if you like. Can still merge this PR though, it shouldn't break anything.

kescobo: OK, got it, thanks

PR #25344, opened 2020-11-26
[FLINT_jll] yank borked version

thofma:
This broke everything, see https://github.com/JuliaPackaging/Yggdrasil/pull/2175. I hope I did it the right way.

@fingolfin

PR #25388, opened 2020-11-27
[LoadFlint] update compat

thofma:
To deal with https://github.com/JuliaRegistries/General/pull/25344 on julia 1.0.

I played around with `VersionSpec` and this seems to be the correct syntax for `[2.6.0 - 2.6.999)`.


thofma:
Would be good if we could merge this to unbreak packages on 1.0. @giordano

PR #25468, opened 2020-11-29
Fix compat entry for LibOSXUnwind_jll

fingolfin:
LibOSXUnwind_jll 0.0.6 is actually compatible with (and required by) Julia >= 1.5.1.
Since the Julia build system pins the JLL to 0.0.6+0 and doesn't use the registry, this
isn't obvious normally, but the incorrect compat entry prevents linking libjulia_jll 1.5
and derived JLLs correctly against libosxunwind.

For details see https://github.com/JuliaPackaging/Yggdrasil/pull/2190#issuecomment-735318854

Ping @giordano @staticfloat

### DIFF
--- a/L/LibOSXUnwind_jll/Compat.toml
+++ b/L/LibOSXUnwind_jll/Compat.toml
@@ -3,4 +3,4 @@ julia = "1"
 
 ["0.0.6-0"]
 JLLWrappers = "1.1.0-1"
-julia = "1.6.0-1"
+julia = "1.5.1-1"
--- a/L/LibOSXUnwind_jll/Versions.toml
+++ b/L/LibOSXUnwind_jll/Versions.toml
@@ -6,3 +6,4 @@ git-tree-sha1 = "cded5b3cf1ce542fb1d322e9d1c5c488ba5725b0"
 
 ["0.0.6+1"]
 git-tree-sha1 = "4ad86f1d058c211e8916ceddf2404ea75b002e47"
+yanked = true


L/LibOSXUnwind_jll/Compat.toml
```diff
@@ -3,4 +3,4 @@ julia = "1"
 
 ["0.0.6-0"]
 JLLWrappers = "1.1.0-1"
-julia = "1.6.0-1"
+julia = "1.5.1-1"
```
giordano, line 6: > LibOSXUnwind_jll 0.0.6 is actually compatible with (and required by) Julia >= 1.5.1.

That's not accurate, `LibOSXUnwind_jll` v0.0.6+1 has the infamous [`aarch64-apple-darwin` platform](https://github.com/JuliaBinaryWrappers/LibOSXUnwind_jll.jl/blob/6e33af9f1adcc6345dac4bbe76e2b92ee3faf5bf/Artifacts.toml#L2), which isn't compatible with Julia v1.5.1.  At very least this should be
```suggestion
julia = "1.5.2-1"
```

giordano (changes requested):

L/LibOSXUnwind_jll/Compat.toml
giordano: Alternatively, we can yank v0.0.6+1, fix the compat bounds for v0.0.6+0, and then merge https://github.com/JuliaPackaging/Yggdrasil/pull/2191 which will become v0.0.6+2, but this must share the same compat as v0.0.6+0.  Is it possible?

fingolfin: OK, the point is that I want to *yank* 0.0.6+1 and instead have 0.0.6+2 which does not include aarch64, see https://github.com/JuliaPackaging/Yggdrasil/pull/2191. How to do this correctly then?

giordano: To yank a version add `yanked = true` after it

giordano: But then you also need to fix the compat entries as appropriate

fingolfin: I don't understand-- there won't be aarch64 in the 0.0.6+2 JLL, so what is wrong about using >=1.5.1?

giordano: I don't know, I said "as appropriate" because I don't know what's the situation.  Is compat bound of 0.0.6+0 the same as future 0.0.6+2.

fingolfin: I think 0.0.6 had no compat bound at all (it may have predated the julia_compat kwarg); but as explained in the linked issue, it is used by Julia >= 1.5.1 (where the build system hardcodes 0.0.6+0)

fingolfin: I've now added `yanked = true` for 0.0.6+1; the compat bound is correct in so far that it reflects precisely those Julia versions which need 0.0.6; and once https://github.com/JuliaPackaging/Yggdrasil/pull/2191 is merged it'll also match what the Project.toml of LibOSXUnwind_jll says.

fingolfin: @giordano does that seem OK to you now? Is the situation a bit clearer now (sorry, but it simply is a bit confusing and explaining it clearly isn't super easy)

PR #25988, opened 2020-12-07
Add DateSelectors.jl with history

oxinabox:
Package: https://github.com/invenia/DateSelectors.jl

This is part of the Contribution2020 event,
where Invenia (and others) are open sourcing a bunch of stuff.

In this particular case we are taking the whole package from closed to open.
Including the history of 7 past releases.

I have semi-manually written this code.
It was copied out of our internal registry, but then I had to add the StatsBase compat and update Julia 1 compat.
This needs some careful checking probably.

Pretty sure it can't be automerged since it wasn't made by the bot


oxinabox:
this might also need a bunch of minor and patch version tags?

PR #26003, opened 2020-12-07
New package: AWSTools.jl w/ history

mattBrzezinski:


A/AWSTools/Compat.toml
```diff
@@ -0,0 +1,103 @@
+[0]
+julia = "0.6-1.3"
+
+["0-0.0.1"]
+Nulls = "0.0.6-0.1"
```
mattBrzezinski, line 5: Looks like this needs to be registered before we can go in with this package.

Looks to be the old version of:
> https://github.com/JuliaData/Missings.jl

PR #26021, opened 2020-12-08
Register Cliquing.jl

iamed2:
This package was previously registered in Invenia's private registry. Versions before 2 depended on a private package which is not being open-sourced, so those have been removed, but otherwise this is copied from Invenia's registry with the URL changed.


iamed2:
Closing for now as there are some unanswered questions around splitting registry history.

iamed2:
Questions have been answered; looks like history splitting will work.

DilumAluthge:
@iamed2 Can I merge this?

iamed2:
@DilumAluthge Yes, please! :)

PR #26489, opened 2020-12-16
Add fake `11.0.0+3` versions to our LLVM assert builds

staticfloat:
This forces BB to build a `v11.0.0+4` version that corresponds to the
appropriate non-assert LLVM builds


staticfloat:
@vchuravy this is my new favorite way to ensure that the assert builds line up nicely with the non-assert builds.  While that's not technically required (since we [can override version numbers in `deps/Versions.make`](https://github.com/JuliaLang/julia/blob/master/deps/Versions.make#L52)), it just helps reduce overall cognitive load.

Once this gets merged, I can then just do a normal build:

```
JULIA_PKG_SERVER= julia-1.6 build_tarballs.jl --verbose --debug --assert --deploy --register
```

in `libLLVM@11.0.0`, then once that gets merged, I can do the same in `Clang@11.0.0` and `LLVM@11.0.0`.

PR #26562, opened 2020-12-17
[Antic_jll] yank bad version

thofma:
We made an oopsie in https://github.com/JuliaPackaging/Yggdrasil/pull/2264.


N/Nemo/Compat.toml
```diff
@@ -65,7 +65,7 @@ Arb_jll = "2.17.0"
 LoadFlint = "0.2.1-0.2"
 
 ["0.18-0"]
-Antic_jll = "0.2.2-0.2"
+Antic_jll = "0.2.2-0.2.2"
```
giordano, line 68: Why do you need to change this if the broken version is yanked?

thofma: julia 1.0

giordano: Julia v1.0 wouldn't use the JLL anyway

thofma: I don't like that it will pull in 0.2.3 on julia 1.0 and 0.2.2 on julia >= 1.1. I don't like the inconsistent state.

thofma:
Thanks!

N/Nemo/Compat.toml
fredrikekre: That seems like a strange argument, basically no package install the same version on all Julia versions :shrug:

thofma: Which package with no restriction on the julia version in the `[compat]` section installs different versions on different julia versions?

thofma: Which package with no restriction on the julia version in the `[compat]` section installs different versions on different julia versions?

PR #26710, opened 2020-12-21
Add action to automatically close stale PRs

giordano:
This uses [`actions/stale`](https://github.com/actions/stale)

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -11,6 +11,7 @@ jobs:
     - uses: actions/stale@v3
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        stale-pr-message: 'This pull request has been inactive for 30 days and will be automatically closed in 3 days. Feel free to register your package or version again once you fix AutoMerge issues'
+        skip-stale-pr-message: true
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        skip-stale-pr-message: true
+        close-pr-message: 'This pull request has been inactive for more than 30 days and has been automatically closed. Feel free to register your package or version again once you fix AutoMerge issues'
+        days-before-stale: 30
+        days-before-close: 3
```
fredrikekre, line 17: Does this mean they are closed after 33 days?

fredrikekre (approved):

.github/workflows/stale.yml
giordano: In practice, yes.  According to https://github.com/actions/stale/issues/28#issuecomment-616489032, `days-before-close: 0` doesn't really work, so if we want to close after 30 days we need to do something like 29 + 1

giordano: Reduced `days-before-close` 1, which is the minimum value

PR #26760, opened 2020-12-22
Update PhysicalCommunications/Package.toml

ma-laforge:
Point to new location of package (JuliaTelecom).  I am told this is necessary.


ma-laforge:
I hope what I am doing is correct.

PR #27285, opened 2021-01-04
Test out the "stopwatch" functionality

DilumAluthge:
cc: @SaschaMann

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -66,7 +66,7 @@ jobs:
     if: github.event_name == 'pull_request'
     environment: stopwatch
     runs-on: ubuntu-latest
-    needs: AutoMerge
+    # needs: AutoMerge # TODO: uncomment this line
     steps:
       # TODO: this step should run the "cron" code path
       - run: echo "Hello world! This job runs 20 minutes after the first job."
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -62,3 +62,11 @@ jobs:
             additional_check_runs = String["CI (1, ubuntu-latest, x64)",],
           )
         shell: julia --color=yes --project=.ci/ {0}
+  AutoMerge-stopwatch:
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Here's what I have in the settings for the "stopwatch" environment:

Screen Shot 2021-01-04 at 04 47 30

DilumAluthge:
Screen Shot 2021-01-04 at 04 51 44

Awesome.

And while this is waiting, it doesn't count towards our 20 job limit, right?

.github/workflows/automerge.yml
```diff
@@ -62,3 +62,11 @@ jobs:
             additional_check_runs = String["CI (1, ubuntu-latest, x64)",],
           )
         shell: julia --color=yes --project=.ci/ {0}
+  AutoMerge-stopwatch:
+    if: github.event_name == 'pull_request'
+    environment: stopwatch
+    runs-on: ubuntu-latest
+    # needs: AutoMerge # TODO: uncomment this line
+    steps:
+      # TODO: this step should run the "cron" code path
```
SaschaMann, line 71: If the cron code path requires secrets, you may have to switch to the `pull_request_target` event and pull in the PR files manually in the first job, or separate the jobs into different workflows (and add a check to the merging step that the check run has actually finished).

DilumAluthge: We only run AutoMerge on PRs made by JuliaRegistrator, which has write access to this repo. So the `pull_request` event should have access to secrets, right? Since the PRs are not made from forks.

SaschaMann:
> And while this is waiting, it doesn't count towards our 20 job limit, right?

That's my understanding of it and at least during the tests I've done, it didn't seem to block any other jobs.

.github/workflows/automerge.yml
SaschaMann: Ah, yes. I didn't know about that, I thought it ran on all PRs.

PR #27616, opened 2021-01-09
New version: libLLVM_jll v11.0.0+7

jlbuild:

* Registering JLL package libLLVM_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/libLLVM_jll.jl
* Version: v11.0.0+7


L/libLLVM_jll/Versions.toml
```diff
@@ -19,12 +19,6 @@ git-tree-sha1 = "e30f4189412370d80b8833c755374f7533bba856"
 ["9.0.1+5"]
 git-tree-sha1 = "8bbcb2ac071438cd061ac210f6ba6754f45b8672"
 
-["10.0.0+0"]
-git-tree-sha1 = "4d429230b2707518d8653262b2c42188a59af9a6"
-
-["10.0.1+0"]
-git-tree-sha1 = "b4e9dd1b8fff9c3f207bbf1e622458c722bfa06b"
-
```
giordano, line 27: It may be better to yank these versions with `yanked = true`, rather than wiping them out

vchuravy: Well I did want to get rid of the Compat.toml entries as well

PR #27740, opened 2021-01-11
Fix AutoMerge

DilumAluthge:


.github/workflows/automerge.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           )
         shell: julia --color=yes --project=.ci/ {0}
   AutoMerge-stopwatch:
-    if: github.repository == github.event.pull_request.head.repo.full_name
+    if: github.event_name != 'pull_request' || github.repository == github.event.pull_request.head.repo.full_name
```
SaschaMann, line 73: This probably shouldn't run on scheduled events either. If you look at the Actions tab, there are scheduled jobs waiting 20 min to be run. It doesn't really matter if it's consistent but afaict there's no reason why it would have to wait, is there?

DilumAluthge: The scheduled job itself doesn't have to wait.

The only thing that waits is the "stopwatch". It waits 10 minutes, and then triggers a new job if there hasn't been a job in the last 8 minutes.

SaschaMann: Yes but the stopwatch is also started for the scheduled jobs currently

DilumAluthge: Yeah I think that's fine, right? It ensures that after a scheduled job ends, we have another job after that.

SaschaMann: It's fine, just noticed it may run more builds than intended.

PR #27752, opened 2021-01-11
Run CI on Julia 1.3, 1.4, 1.5, 1.6-beta1, and nightly

DilumAluthge:
The more Julia versions that we run CI on, the less likely it is that I'll merge a PR that breaks all the things.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,8 @@ jobs:
       matrix:
         version:
           - '1.3'
+          - '1.4'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -17,6 +17,8 @@ jobs:
       matrix:
         version:
           - '1.3'
+          - '1.4'
+          - '1.5'
```
giordano, line 21: Isn't `1.5` the same as `1` at the moment?

PR #27753, opened 2021-01-11
Register LazyArtifacts.jl with no registered versions

DilumAluthge:
For context, see the discussion in https://github.com/JuliaRegistries/General/pull/27666

Basically, I can't figure out how to register CUDA 2.5.0 in the General registry. CUDA 2.5.0 has a dependency on the LazyArtifacts stdlib, which is a stdlib that only exists in Julia >= 1.6.

So, when I registered CUDA 2.5.0, it broke all registry operations for users on Julia <1.6 - specifically, Pkg would complain that it did not know the UUID `4af54fe1-eca0-43a8-85a7-787d91b784e3` (the UUID for the LazyArtifacts stdlib in Julia >= 1.6).

I think the solution is to add LazyArtifacts.jl to the General registry (so that Julia < 1.6 will know the UUID `4af54fe1-eca0-43a8-85a7-787d91b784e3`), but without any registered versions.

### DIFF
--- a/L/LazyArtifacts/Package.toml
+++ b/L/LazyArtifacts/Package.toml
@@ -1,0 +1,3 @@
+name = "LazyArtifacts"
+uuid = "4af54fe1-eca0-43a8-85a7-787d91b784e3"
+repo = "https://github.com/JuliaPackaging/LazyArtifacts.jl.git"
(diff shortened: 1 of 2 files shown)


staticfloat:
I'd rather not explore our boundary conditions like this; let's just actually release a version from that psuedo-package.

DilumAluthge:
@staticfloat Can you register a version from LazyArtifacts.jl?

giordano:
I'm going to add some tests to see if it can at least be loaded

giordano:
https://github.com/JuliaRegistries/General/pull/27799.  Of course it was totally broken: https://github.com/JuliaPackaging/LazyArtifacts.jl/commit/6c3c9f90e0bcf64d9cd4e3fa5b4622be6af87a2c

PR #27974, opened 2021-01-14
Further decrease timeout of CI and AutoMerge jobs

giordano:
Also, disable PkgServer and remove redundant CI job (`1` is the same as `1.5`).


DilumAluthge:
I would remove the `'1.5'` job instead of the `'1'` job.

giordano:
Done

DilumAluthge (approved):

PR #28033, opened 2021-01-15
Delete branch when closing stale PRs

giordano:
I want to automatically delete branches when the pull requests are closed with [`actions/stale`](https://github.com/actions/stale).  Today https://github.com/actions/stale/pull/190 was finally merged, but as far as I can see there is no new release, yet.  @SaschaMann can I use commit https://github.com/actions/stale/commit/b12dccced80582bf8c52244f75838892ec5c1e82 (currently the latest commit on `main`) as reference?  I'm always a bit confused about which branch to look at for GitHub Actions.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -8,7 +8,7 @@ jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/stale@v3
+    - uses: actions/stale@b12dccced80582bf8c52244f75838892ec5c1e82
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-pr-message: 'This pull request is inactive and will be closed'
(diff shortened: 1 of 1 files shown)


SaschaMann:
>  I'm always a bit confused about which branch to look at for GitHub Actions.

The safest way is to click on the latest release in the sidebar and then copy the commit it points to:

grafik

grafik

grafik

---

The details of which branch that is on can vary between actions. Using the commit associated with the tag should always work because that's what people would access via `@v3.0.14` etc at the time of copying the commit (this association may change later, hence we use the hash, not the tag)

giordano:
> The safest way is to click on the latest release in the sidebar and then copy the commit it points to:

The problem is that there is no tag here to look at, that's why I want to reference the commit.  But good to point out I could have just looked what branch they tagged before, thanks!

SaschaMann:
> The problem is that there is no tag here to look at, that's why I want to reference the commit. 

Those screenshots were from the actions/stale repo. You can reference the commit but use the tag to find which commit it is.

SaschaMann:
Sorry I misunderstood you, I overread the part where you wanted to use a not-yet-released version.

DilumAluthge (approved):

giordano:
[It worked](https://github.com/JuliaRegistries/General/pull/26461#event-4215110621), yay!

PR #28171, opened 2021-01-17
Really fix CI this time

DilumAluthge:


giordano (approved):
Looks good as far as I can see, AutoMerge job at least installs RegistryCI v6.2.1

PR #28237, opened 2021-01-19
Increase the stale threshold to 60 days

DilumAluthge:
It's not uncommon for people to be away from GitHub for 30 days.

I think it should be fine to increase the stale threshold to 60 days.


giordano:
I mean, people may be away also for 60 days.  They can always register again (and retrigger registration is in many cases need anyway), but at least the list of open PRs is cleaner

PR #28266, opened 2021-01-19
update URL for Checkpoints.jl

oxinabox:
Looks like this got missed when @rofinn open sourced this


C/Checkpoints/Package.toml
```diff
@@ -1,3 +1,3 @@
 name = "Checkpoints"
 uuid = "b4a3413d-e481-5afc-88ff-bdfbd6a50dce"
-repo = "https://gitlab.invenia.ca/invenia/Checkpoints.jl.git"
+repo = "https://github.com/invenia/Checkpoints.jl"
```
nickrobinson251, line 3: Was removing `.git` suffix international?

oxinabox: It was not
```suggestion
repo = "https://github.com/invenia/Checkpoints.jl.git"
```

PR #28310, opened 2021-01-19
Restrict MbedTLS_jll 2.25 to Julia 1.7+

staticfloat:
MbedTLS broke compatibilty between 2.24 and 2.25, 2.24 is what ships with 1.6

X-ref: https://github.com/JuliaPackaging/Yggdrasil/pull/2421


giordano (approved):
I _think_ it's OK, but I always get compat wrong

PR #28451, opened 2021-01-22
Fix Singular_jll compat entry

fingolfin:
This is a stop-gap measure until we can properly fix the underlying problem, and teach BB to support VersionSpec for dependencies correctly. But it'll save me some headaches...

See also https://github.com/JuliaBinaryWrappers/libsingular_julia_jll.jl/pull/1

### DIFF
--- a/L/libsingular_julia_jll/Compat.toml
+++ b/L/libsingular_julia_jll/Compat.toml
@@ -21,7 +21,7 @@ julia = "1.5.0-1"
 
 ["0.5-0"]
 JLLWrappers = "1.2.0-1"
-Singular_jll = "402.0.1"
+Singular_jll = "402.0.1-402.0"
 
 ["0.5-0.5.3"]
 julia = "1.3.0-1"


giordano (approved):
I'm not super-versed in compat bounds in the registry, but I _think_ this is OK.

If we do this, I don't see the need for https://github.com/JuliaBinaryWrappers/libsingular_julia_jll.jl/pull/1

L/libsingular_julia_jll/Compat.toml
```diff
@@ -21,7 +21,7 @@ julia = "1.5.0-1"
 
 ["0.5-0"]
 JLLWrappers = "1.2.0-1"
-Singular_jll = "402.0.1"
+Singular_jll = "402.0.1-402.0"
```
fingolfin, line 24: @giordano As you know, the goal here is to replicate the effect of a Project.toml compat entry of the form `~402.0.1`. I believe it is, based on my believe that the string here is basically a valid argument for the `VersionSpec` constructor, and I see this:
```
julia> VersionSpec("402.0.1-402.0") == Pkg.Types.semver_spec("~402.0.1")
true
```
Perhaps somebody who really knows can confirm this is the correct way to do it, e.g. @fredrikekre ?

DilumAluthge:
Looks right at a glance

PR #30785, opened 2021-02-24
Interpolations v0.13.0+ compat StaticArrays 0.12+

mkitti:
Interpolations v0.13.0 introduced code that depends on StaticArrays 0.12 or later. https://github.com/JuliaMath/Interpolations.jl/blob/e8165ec2d1859a2e67ba89e8a17b4e321c004410/src/lanczos/lanczos.jl#L115


DilumAluthge:
Looks like CI is failing.

DilumAluthge:
Looks like a problem with the compression.

DilumAluthge:
Instead of doing this PR, here is what I would recommend:
1. Make a new patch release of Interpolations.jl (e.g. Interpolations 0.13.2) that has a correct compat entry for StaticArrays.
2. Yank old versions of Interpolations (e.g. Interpolations 0.13.0 and Interpolations 0.13.1) that do not have the compat entry for StaticArrays.

PR #30956, opened 2021-02-28
Update URL for LiveServer.jl

tlienart:
The repo ownership was transferred to me back in November https://github.com/tlienart/LiveServer.jl (I just hadn't realised until now that I needed to update the URL 😱)

Btw: I tagged two patch releases since then; I hadn't installed the Registrator.jl app properly on that repo so I didn't realise that these releases failed to registre (up until today). Should I reset the patch version to 0.6.1 to avoid Registrator complaining about skipped version? Thanks!


giordano:
>  Should I reset the patch version to 0.6.1 to avoid Registrator complaining about skipped version?

I think that's the easiest and cleanest option, yes

PR #30982, opened 2021-02-28
Update the repo URL for MKL.jl

ViralBShah:
Changing the URL since we moved MKL.jl to a new org. Is this the only place this is necessary? I am mainly doing this so that I can then use the web registrator on juliahub.com (which otherwise complains).

cc @giordano @DilumAluthge


DilumAluthge:
Yeah this should be the only place you need to change it.

ViralBShah:
What does this error mean?

DilumAluthge:
Oh yeah that's just because we merged it manually. 

It shouldn't even error in this case. I'll try to fix the config.

PR #31523, opened 2021-03-08
add license automerge check to README

ericphanson:
for https://github.com/JuliaRegistries/RegistryCI.jl/pull/344


ericphanson:
Note we don't only check those two names, but I thought giving examples would help clarify what the requirement is talking about for a new dev. (We either check every single plaintext top-level file if there are less than 100 of them, or a hardcoded list of ~500 possible names).

DilumAluthge (approved):

PR #31669, opened 2021-03-10
Change package source repository

pdimens:
PopGen.jl is back into BioJulia, and we would like future releases to occur on that repo instead of my fork.

### DIFF
--- a/P/PopGen/Package.toml
+++ b/P/PopGen/Package.toml
@@ -1,3 +1,3 @@
 name = "PopGen"
 uuid = "af524d12-c74b-11e9-22a8-3b091653023f"
-repo = "https://github.com/pdimens/PopGen.jl.git"
+repo = "https://github.com/BioJulia/PopGen.jl.git"


giordano:
Are all registered already available in the BioJulia repository?  As far as I can see, https://github.com/pdimens/PopGen.jl is 85 commits ahead of BioJulia

pdimens:
@giordano I haven't merged all the branches yet, but that's the next to follow. There are a few repo things I'd like to do but don't have the right permissions to do so (like removing `master` and making `dev` the default branch). All the future releases will occur off of BioJulia. I'm not sure how to migrate previous released back up to the source repo though.

giordano:
But if we merge this _before_ you move everything, versions will become uninstallable for the users

pdimens:
That makes sense. I tried googling how merge releases back to the base repo and am coming up empty. Maybe I'm not phrasing the question correctly. Would you happen to know how I could move the tagged releases from my fork to the original one in BioJulia?

pdimens:
That would definitely be easier, but BioJulia has the original PopGen.jl, wouldn't it need to be deleted from that account? If that happens, my understanding is that Github will [somehow] choose one of the forks to be the origin, and transferring my fork back to BioJulia may inadvertently lead to the main package repo being incorrectly pinged as a fork.

giordano:
> That would definitely be easier, but BioJulia has the original PopGen.jl, wouldn't it need to be deleted from that account?

Yes

> and transferring my fork back to BioJulia may inadvertently lead to the main package repo being incorrectly pinged as a fork.

Is that bad?  At the moment your repository is anyway identified as fork of something else when it's the upstream.  You can contact GitHub support to fix it.  You can for example ask to make your repository the upstream and BioJulia a fork, delete BioJulia's fork, move your repository there

pdimens:
Thanks for the guidance on this. pdimens/PopGen.jl is now the upstream and has been transferred to BioJulia/PopGen.jl. IIRC, all releases can safely be registered and tagged in BioJulia/PopGen.jl

PR #31881, opened 2021-03-13
Move VectorizationBase to JuliaSIMD

chriselrod:


chriselrod:
I wanted to tag a new release:
https://github.com/JuliaSIMD/VectorizationBase.jl/commit/707216c90ac81b38ef75dbf0523ee8fc7e8958ec
But no response from the registrator. Is there a step missing from the README?

Or can I expect it to work after step 2?

I've now made a PR for each of the 7 repos I've moved to JuliaSIMD:
https://github.com/JuliaSIMD

PR #31558, opened 2021-03-09
New version: Cbc_jll v2.10.5+2

jlbuild:
Autogenerated JLL package registration

* Registering JLL package Cbc_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/Cbc_jll.jl
* Version: v2.10.5+2
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/adaac1c3d733beacacaf59988bdf52e44242f92c

PR #31977, opened 2021-03-15
Fix version requirements for Cbc_jll

odow:
# The problem

The issue arose because Cbc_jll@2.10.5+1 was previously released,
but due to some compatibility issues, the next version released
was 2.10.3.

We recently re-compiled Cbc pointing at the upstream 2.10.5, which
got released as 2.10.5+2
https://github.com/JuliaRegistries/General/pull/31558
https://github.com/JuliaPackaging/Yggdrasil/pull/2668
We deemed this okay because the only consumer of Cbc_jll is Cbc.jl, 
which pinned Cbc_jll@2.10.3.

However, we didn't account for the version incompatibilities (it seems like
`[Compat]` can't handle `+N` version revisions).

I think this PR makes `2.10.5+0` and `2.10.5+1` un-installable, but no one 
should be using it.

# Alternatives

If modifying versions like this is unacceptable, an alternative is to release new 
versions of the dependencies using the "multiply the version number by 100" trick 
so we can install `Clp@100.1700.600` etc.

I've verified that this allows Cbc.jl to work on Linux: https://github.com/jump-dev/Cbc.jl/pull/160

cc @giordano

### DIFF
--- a/C/Cbc_jll/Compat.toml
+++ b/C/Cbc_jll/Compat.toml
@@ -1,17 +1,12 @@
 [2]
 julia = "1"
-
-["2-2.10.3"]
 Cgl_jll = "0.60.2"
 Clp_jll = "1.17.6"
 CoinUtils_jll = "2.11.3"
-JLLWrappers = "1.1.0-1"
 Osi_jll = "0.108.5"
 
+["2-2.10.3"]
+JLLWrappers = "1.1.0-1"
+
 ["2.10.5-2"]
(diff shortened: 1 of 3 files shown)


giordano:
> I think this PR makes 2.10.5+0 and 2.10.5+1 un-installable, but no one
> should be using it.

For extra safety, you should yank those versions, i.e. add
```toml
yanked = true
```
below each version

odow:
I've added the `yanked = true`

PR #32842, opened 2021-03-25
[GHA] Run CI also for Julia v1.5

giordano:


.github/workflows/ci.yml
```diff
@@ -24,9 +24,8 @@ jobs:
         version:
           - '1.3'
           - '1.4'
-          - '1'
-          - '^1.6.0-0' # TODO: remove this line once 1.6 is released
-          # - '1.6'    # TODO: uncomment this line once 1.6 is released
+          - '1.5'
+          - '1.6'
```
DilumAluthge, line 28:
```suggestion
          - '1'
```

DilumAluthge (approved):

.github/workflows/ci.yml
giordano: But why?

DilumAluthge: So, I guess the question is this: When Julia 1.7 comes out, suppose that we forget to immediately add `- '1.7'` to the list. Which would we rather have be the situation:
1. We are testing on 1.6 but not testing on 1.7.
2. We are testing on 1.7 but not testing on 1.6.

I think that situation 2 is preferable. If for some reason we forget to immediately update the workflow file, I'd rather be testing on 1.7 but not testing on 1.6.

giordano: But we have been testing on v1.6 for months.  If this had been
```yaml
- "1.5"
- "^1.6.0-0"
```
we should have only change it to
```yaml
- "1.5"
- "1.6"
```
with no practical difference

DilumAluthge: Yeah, we were kind of in a weird spot since we were using the 1.6 RC. The only reason I had this:
```yaml
- "1.3"
- "1.4"
- "1.5"
- "^1.6.0-0"
```

Instead of this:
```yaml
- "1.3"
- "1.4"
- "1"
```

is that I specifically wanted to also use the 1.6 RC. But this is kind of a unique situation in which I started using the 1.6 RC for AutoMerge because of some issues with 1.5.3.

Normally, before 1.6 was released, we would have had this:
```yaml
- "1.3"
- "1.4"
- "1"
```

And then after 1.6 was released, it would change to this:
```yaml
- "1.3"
- "1.4"
- "1.5"
- "1"
```

DilumAluthge: Idk why I had `"1.5"` instead of `"1"` previously. That was an oversight on my part.

DilumAluthge: Normally I do not run CI or AutoMerge on a release candidate. I usually just run AutoMerge on `"1"`, and CI on everything up to `"1"`. 

But AutoMerge would OOM or hang with certain packages on 1.5.3, so I made the not-necessarily-wise decision to run AutoMerge on a release candidate.

PR #34270, opened 2021-04-14
Disable AutoMerge

DilumAluthge:


DilumAluthge:
A temporary change due to a change in Registrator's security guarantees. Context here: https://github.com/JuliaRegistries/Registrator.jl/issues/329

PR #34335, opened 2021-04-15
Update urls of two repos transferred from ablaom to JuliaAI

ablaom:
The repositories are: [EarlyStopping.jl](https://github.com/JuliaAI/EarlyStopping.jl) and [IterationControl.jl](https://github.com/JuliaAI/IterationControl.jl)


ablaom:
@fredrikekre Thank you.

PR #24779, opened 2020-11-17
Remove Julia from the General registry

DilumAluthge:
We now have the https://julialang-s3.julialang.org/bin/versions.json file, which is auto-updated. So we don't need to have Julia in the General registry.

Closes #23334

PR #35320, opened 2021-04-26
Julia 1.6.1

pfitzseb:


fredrikekre:
image

pfitzseb:
Yeah dunno, we should either automate this or get rid of Julia in the registry altogether. 
I can't quite remember what breakage caused https://github.com/JuliaRegistries/General/pull/25234, but IIRC it wasn't just JuliaHub/DocumentationGenerator.

DilumAluthge:
Yeah I really wish we could take it out of the registry and just use https://github.com/SaschaMann/Julia-versions.json instead

Unfortunately, I don't know how to fix the failures that https://github.com/JuliaRegistries/General/pull/24779 caused

PR #35487, opened 2021-04-27
Only run the Manifest Updater one time per week

DilumAluthge:
Before this PR, we run the Manifest Updater one time per day.

After this PR, we run the Manifest Updater one time per week.

We also have a `workflow_dispatch` trigger that allows users with commit access to manually trigger a run if they want to get immediate manifest updates.

### DIFF
--- a/.github/workflows/update_manifests.yml
+++ b/.github/workflows/update_manifests.yml
@@ -1,7 +1,7 @@
 name: Update Manifests
 on:
   schedule:
-    - cron: 0 0 * * *
+    - cron: 0 0 * * 0
   workflow_dispatch:
 jobs:
   update_manifests:


ericphanson:
Can we do it as a workflow trigger? Then we can just run it when we want to bump versions, eg new RegistryCI release

DilumAluthge:
> Can we do it as a workflow trigger? Then we can just run it when we want to bump versions, eg new RegistryCI release

Yep, we also have a `workflow_dispatch` trigger.

But I do think that there is value in also having a scheduled (cron) job that runs e.g. weekly. That way, we make sure that we update to e.g. the latest bug fix releases of all of our indirect (recursive) dependencies.

ericphanson:
Ah good, the trigger will be useful then.

Yeah, I can see it being be bad to get way behind on releases and then run into accumulated issues doing a big update to get some urgent bugfix.

ericphanson (approved):

PR #35519, opened 2021-04-27
Deploy `RegistryCI#master` to a staging environment with a read-only AutoMerge job (i.e. an AutoMerge job that does not merge PRs, post comments, or update commit statuses)

### LINKED ISSUE #35514: RegistryCI#master workflow for testing on General
I originally proposed this on #pkg-registration in the Julia slack:

> The idea that RegistryCI could break General’s automerge is kinda scary, and while we have integration tests, it’s always a bit of worry that when we bump General we might hit weird edge cases with new package registrations and things could break. Especially because RegistryCI does things like load arbitrary packages, clone git repos, etc, and it can be hard to test every edge case (esp since writing the integration tests isn’t always super easy either).

> So I have an idea: what if we sometimes add another workflow to General’s CI which runs RegistryCI#master but does not write comments or influence merging. Then if we want to test something, we can just run that for a few days and then go through and see if we have any spurious failures before cutting a release.

@DilumAluthge mentioned that we need to ensure it doesn't post a comment or create the `automerge/decision` commit status.

DilumAluthge:
Fixes #35514

### DIFF
--- a/.github/workflows/automerge_staging.yml
+++ b/.github/workflows/automerge_staging.yml
@@ -1,0 +1,88 @@
+# INSTRUCTIONS: To enable or disable this workflow, please go to 
+# https://github.com/JuliaRegistries/General/actions/workflows/automerge_staging.yml
+# and click on the "···" button on the right-hand side.
+# Then click on "Enable workflow" or "Disable workflow".
+
+name: AutoMerge (staging)
+on:
+  pull_request:
+
+# Make sure that the `GITHUB_TOKEN` only has read-only permissions
+# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions
+permissions: read-all
+
+jobs:
+  AutoMerge_staging:
+    timeout-minutes: 60
+    if: github.event_name != 'pull_request' || github.repository == github.event.pull_request.head.repo.full_name
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        version:
+          - '1'
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
+    steps:
+      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
+      - uses: julia-actions/setup-julia@0b9b1d2cd24245f151902702d8e73b3f6b910014 # v1.6.0
+        with:
+          version: ${{ matrix.version }}
+      - run: rm -rf .ci/JuliaManifest*
+      - run: rm -rf .ci/Manifest*
+      - name: Cache artifacts
+        uses: actions/cache@1a9e2138d905efd099035b49d8b7a3888c653ca8 # v2.1.5
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/.ci/Manifest.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - run: julia --color=yes -e 'import Pkg; Pkg.Registry.add("General")'
+        env:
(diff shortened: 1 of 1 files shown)


DilumAluthge:
https://github.com/JuliaRegistries/RegistryCI.jl/pull/383 needs to be merged before we can merge this PR.

.github/workflows/automerge_staging.yml
```diff
@@ -0,0 +1,85 @@
...
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/.ci/Manifest.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - run: julia --color=yes -e 'import Pkg; Pkg.Registry.add("General")'
+        env:
+          JULIA_PKG_SERVER: ""
+      - run: julia --color=yes -e 'import Pkg; Pkg.Registry.update()'
+      - run: .ci/instantiate.sh
```
ericphanson, line 47: I think maybe we should just add `RegistryCI#master` instead of using a manifest, so we always get the latest one. That way if we find a bug on staging, we can just fix it and future PRs will use the new version without a lot of manual version-bumping

DilumAluthge: Take a look at the commit I just pushed. It deletes all the `Manifest.toml` files at the beginning. So we won't use a manifest.

ericphanson: hm, that's good but we still need to add the development version or else we'll just get the release one

ericphanson: ahh you do do that already, sorry!

ericphanson: but actually, do we need the `run: .ci/instantiate.sh` step at all?

DilumAluthge: It has a bounded retry loop, which is useful because the network is flaky. CI used to fail frequently just because the network is unreliable, but the CI failures stopped once we added the bounded retry loop.

DilumAluthge: E.g. CI would fail because we download a lot of dependencies and artifacts, and sometimes one of those downloads would fail due to an intermittent network issue, and then the whole CI or AutoMerge job would fail.

ericphanson: ah ok, sounds good

DilumAluthge:
BTW, before we merge this PR, make sure that we do the two TODOs on lines 54 and 55:
https://github.com/JuliaRegistries/General/blob/db8d4a9a3ea15627d53e83a726131dc166908104/.github/workflows/automerge_staging.yml#L54-L55

.github/workflows/automerge_staging.yml
```diff
@@ -0,0 +1,87 @@
...
+          version: ${{ matrix.version }}
+      - run: rm -rf .ci/JuliaManifest*
+      - run: rm -rf .ci/Manifest*
+      - name: Cache artifacts
+        uses: actions/cache@1a9e2138d905efd099035b49d8b7a3888c653ca8 # v2.1.5
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/.ci/Manifest.toml') }}
```
ericphanson, line 40: Will this clash with the main CI cache? And we aren’t using this manifest here, right? 

I’m not totally sure how these are supposed to be used

ericphanson (approved):
Besides the caching issue, this seems good to me. I think @fredrikekre has a point that a separate repo would be nice, but given that this should be non-disruptive (with a read-only token) and there’s a need for it, it seems worth doing here at this point.

.github/workflows/automerge_staging.yml
DilumAluthge: Hmmm. Maybe @giordano or @SaschaMann can weigh in.

PR #35616, opened 2021-04-29
Run TagBot Triggers inside a bounded retry loop

### LINKED ISSUE #35614: The TagBot Triggers cron job keeps failing intermittently
The TagBot Triggers cron job keeps failing intermittently. See for example: https://github.com/JuliaRegistries/General/runs/2463530364?check_suite_focus=true

Re-running the job usually succeeds. See for example: https://github.com/JuliaRegistries/General/runs/2463776699?check_suite_focus=true

Perhaps we could put the TagBot Triggers cron job inside a bounded retry loop?

We currently use a bounded retry loop for instantiating the CI environment (https://github.com/JuliaRegistries/General/blob/master/.ci/instantiate.sh). So we could probably use a very similar script to implement a bounded retry loop for the step in the TagBot Triggers cron job that fails intermittently.

cc: @christopher-dG

DilumAluthge:
Fixes #35614

### DIFF
--- a/.ci/tagbot_triggers.sh
+++ b/.ci/tagbot_triggers.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# This script runs `TagBot.main()` inside a bounded retry loop.
+
+try_count=0
+while [ $try_count != 9 ]; do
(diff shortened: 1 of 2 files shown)


DilumAluthge:
I'm going to go ahead and merge this. If this looks wrong, @christopher-dG please revert it.

PR #35965, opened 2021-05-03
Move RLBase, RLCore, RLEnvs, RLZoo into RL

findmyway:
Hi @GunnarFarneback ,

Based on @StefanKarpinski 's suggestion, I used the git-filter-repo tool according to the answer here: https://stackoverflow.com/a/10548919/3401312 to update the git tree hash in the meanwhile. Please let me know if did anything wrong here.

Thanks!

### DIFF
--- a/R/ReinforcementLearningBase/Package.toml
+++ b/R/ReinforcementLearningBase/Package.toml
@@ -1,3 +1,4 @@
 name = "ReinforcementLearningBase"
 uuid = "e575027e-6cd6-5018-9292-cdc6200d2b44"
-repo = "https://github.com/JuliaReinforcementLearning/ReinforcementLearningBase.jl.git"
+repo = "https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl.git"
+subdir = "src/ReinforcementLearningBase"
--- a/R/ReinforcementLearningBase/Versions.toml
+++ b/R/ReinforcementLearningBase/Versions.toml
@@ -1,89 +1,89 @@
 ["0.1.0"]
-git-tree-sha1 = "f548233ca31a278818454e549406ac8f34135ba2"
+git-tree-sha1 = "937076c70633bced57129c87fded3abcc7b85403"
 
 ["0.2.0"]
-git-tree-sha1 = "b98eb69967fecfb6bec06d271ecbf2d5f42fe76b"
+git-tree-sha1 = "ce77348d2e6580ffcdb402ca327de63eef2b10f8"
 
 ["0.2.1"]
-git-tree-sha1 = "f75d10578b8d71ef394ffbb0fd18edcbc365da04"
+git-tree-sha1 = "17a7d78eb0657d9ad0fb70b26091810d9876664b"
 
 ["0.2.2"]
-git-tree-sha1 = "f85fac6aa9cd8621c234797a0735eb799ebda29e"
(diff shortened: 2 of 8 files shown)


GunnarFarneback:
I don't know what you did more exactly but changing the stored content hashes in the repository is totally out of the question. That would imply that there is now something else in those versions than when they were released.

StefanKarpinski:
Julia package versions are associated with specific git trees and this association is immutable — once you've published a version, that's what it is. You're never allowed to change it. In order not to break the ability to git clone a package and checkout older versions, you need to make sure that however you've rewritten the repo, those exact trees are still there in the repo history somewhere. You can move the history of a package to a subdirectory of a new repo and things will be fine since the original trees are still there, they've just been moved to a subdirectory — but _nothing_ in that subdirectory can change, just the location of it in the repo. You can also rewrite all the commit history between released versions: the version is associated with a tree, not a commit (which includes the entire development history of how it was created). It is unclear to me what you've done here and what you're trying to do. Can you explain a little what you're attempting to accomplish?

DilumAluthge:
Maybe it would be good if we had a tutorial on how to use `git-filter-repo` to move the history of a package into a subdirectory in a way that preserves the tree hashes. Then we can point package authors to that tutorial.

StefanKarpinski:
Ok, I checked out both ReinforcementLearning.jl and ReinforcementLearningBase.jl and the tree rewriting seems to have been done correctly. The latest tree for ReinforcementLearningBase.jl matches ReinforcementLearning.jl/src/ReinforcementLearningBase. So that's good. And ReinforcementLearning.jl has the right tree objects in it. For example:
```sh
$ git show f548233ca31a278818454e549406ac8f34135ba2
tree f548233ca31a278818454e549406ac8f34135ba2

.gitignore
.travis.yml
Project.toml
README.md
REQUIRE
src/
test/
```
Now for some reason this change to General tries to replace the trees for versions with commits. I'm not entirely sure what those commits are, but since versions are associated with trees rather than commits, that's clearly wrong. I think this change might be correct if you just drop all the hash changes.

findmyway:
> Now for some reason this change to General tries to replace the trees for versions with commits. I'm not entirely sure what those commits are, but since versions are associated with trees rather than commits, that's clearly wrong. I think this change might be correct if you just drop all the hash changes.

Those changes were commit ids associated with version tags. I had confused the commit id with tree id. Now I understand why it's wrong.  They are restored now. Thanks!

findmyway:
Hi @GunnarFarneback , @fredrikekre , @DilumAluthge ,

I've tested my fork locally and it seems all the previous versions of ReinforcementLearning.jl can be installed correctly. So I think it is good to merge this now? 😃

GunnarFarneback:
I've confirmed that all trees can be found by running
```
git clone https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl.git
cd ReinforcementLearning.jl
julia
```
and then this code:
```
using TOML
for pkg in ["ReinforcementLearningBase",
	    "ReinforcementLearningCore",
	    "ReinforcementLearningEnvironments",
	    "ReinforcementLearningZoo"]
    println(pkg)
    versions = TOML.parsefile(joinpath(first(DEPOT_PATH), "registries",
                              "General", pkg[1:1], pkg, "Versions.toml"))
    for version in sort(collect(keys(versions)))
        tree_sha = versions[version]["git-tree-sha1"]
        print(version)
        try
            println(" found")
        catch
            println(" missing")
        end
    end
end
```

PR #36047, opened 2021-05-04
remove 1.4.3 version

jd-lara:
version 1.4.3 introduced a sem ver mistake.


fredrikekre:
Not allowed, but you can yank it if you like.

jd-lara:
Thanks.

PR #36683, opened 2021-05-13
[README] Add FAQ about how to make the package load with proprietary deps

giordano:
I link to https://github.com/JuliaRegistries/RegistryCI.jl/issues/283 every other day, so I guess this earned the right to be a _frequently_ asked question:

image

The text is maybe too verbose, please suggests improvements

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,22 @@ If you for some reason can't (or won't) adhere to the guidelines you will have
 to wait for a human to review/merge the pull request. You can contact a human
 in the `#pkg-registration` channel in the official Julia Slack to expedite this process.
 
+#### My package fails to load because it needs proprietary software/additional setup to work, what can I do?
+
+Before merging a pull request, AutoMerge will check that your package can be installed and
+loaded.  It is OK for your package to not be fully functional, but making it at least load
+successfully would streamline registration, as it does not require manual intervention from
+the registry maintainers.  This would also let other packages depend on it, and use its
+functionalities only when the proprietary software is available in the system, as done for
+example by the [`CUDA.jl`](https://github.com/JuliaGPU/CUDA.jl) package.  If you are not
+able or willing to make your package always loadable without the proprietary dependency
+(which is the preferred solution), you can check if the environment variable
+`JULIA_REGISTRYCI_AUTOMERGE` is equal to `true` and make your package loadable during
+AutoMerge at least, so that it can be registered without manual intervention.  Examples of
+packages with proprietary software that use the environment variable check include
+[`Gurobi.jl`](https://github.com/jump-dev/Gurobi.jl) and
+[`CPLEX.jl`](https://github.com/jump-dev/CPLEX.jl).
+
 #### My pull request has a merge conflict, what do I do?
 
 Retrigger Registrator.
@@ -197,6 +213,7 @@ As long as the package is not yet registered, renaming the package from
   in the old pull request that it can be closed, linking to the new one.
 
 #### How do I rename an existing registered package?
+
 Technically, you can't rename a package once registered, as this would break existing users.
 But you can re-register the package again under a new name with a new UUID.
 Which has basically the same effect.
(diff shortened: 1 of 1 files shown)


christopher-dG (approved):

fredrikekre:
Worth mentioning that it can be useful to have the package load for other purposes too. For example package A, with proprietary deps, might be installed as a dependency when I install package B. Perhaps will only use the parts of B that doesn't require A, and in such a case it is useful if `using B` doesn't error.

giordano:
Good point, I added it, mentioning also `CUDA.jl`, so hopefully people understand how this is useful.  Thanks!

README.md
```diff
@@ -163,6 +163,18 @@ If you for some reason can't (or won't) adhere to the guidelines you will have
 to wait for a human to review/merge the pull request. You can contact a human
 in the `#pkg-registration` channel in the official Julia Slack to expedite this process.
 
+#### My package fails to load because it needs proprietary software/additional setup to work, what can I do?
+
+Before merging a pull request, AutoMerge will check that your package can be installed and
+loaded.  It is OK for your package to not be fully functional, but making it at least load
+successfully would streamline registration, as it does not require manual intervention from
+the registry maintainers.  This would also let other packages depend on your one, and use
```
ericphanson, line 171:
```suggestion
the registry maintainers.  This would also let other packages depend on it, and use
```

README.md
```diff
@@ -163,6 +163,18 @@ If you for some reason can't (or won't) adhere to the guidelines you will have
...
+its functionalities only when the proprietary software is available in the system, as done
+for example by the [`CUDA.jl`](https://github.com/JuliaGPU/CUDA.jl) package.  You can at
+least make your package load successfully with AutoMerge by allowing the necessary software
+to be not available when the environment variable `JULIA_REGISTRYCI_AUTOMERGE` is equal to
+`true` (e.g., `get(ENV, "JULIA_REGISTRYCI_AUTOMERGE", "false") == "true"`).
```
ericphanson, line 176:
```suggestion
for example by the [`CUDA.jl`](https://github.com/JuliaGPU/CUDA.jl) package.  If you aren't
able or willing to make your package always loadable without the proprietary dependency
(which is the preferred solution), you can check if the environment variable
`JULIA_REGISTRYCI_AUTOMERGE` is equal to`true` (e.g.,
`get(ENV, "JULIA_REGISTRYCI_AUTOMERGE", "false") == "true"`) and try to make your package
loadable during AutoMerge at least, so that it can be registered without manual intervention.
```

ericphanson (commented):
I think it would be good to emphasize more that always-loadable is the preferred solution

README.md
```diff
@@ -163,6 +163,20 @@ If you for some reason can't (or won't) adhere to the guidelines you will have
...
+loaded.  It is OK for your package to not be fully functional, but making it at least load
+successfully would streamline registration, as it does not require manual intervention from
+the registry maintainers.  This would also let other packages depend on it, and use
+its functionalities only when the proprietary software is available in the system, as done
+for example by the [`CUDA.jl`](https://github.com/JuliaGPU/CUDA.jl) package.  If you aren't
+able or willing to make your package always loadable without the proprietary dependency
+(which is the preferred solution), you can check if the environment variable
+`JULIA_REGISTRYCI_AUTOMERGE` is equal to`true` (e.g.,
+`get(ENV, "JULIA_REGISTRYCI_AUTOMERGE", "false") == "true"`) and try to make your package
+loadable during AutoMerge at least, so that it can be registered without manual intervention.
```
odow, line 178: How about adding something like
````
Examples of packages with proprietary software that use the environment variable
check include:
 - [Gurobi.jl](https://github.com/jump-dev/Gurobi.jl)
 - [CPLEX.jl](https://github.com/jump-dev/CPLEX.jl)
Consult their source code for examples on how to write this check, including as
part of a `Pkg.build` step.
```

giordano: Ok, thanks, I mentioned those packages so readers can go and see them, but I didn't mention `Pkg.build` explicitly to not give them too many ideas :wink:

ericphanson (approved):

PR #36722, opened 2021-05-14
Remove the SuiteSparse.jl package

DilumAluthge:
If I understand correctly, this should never have been registered in the first place.

Also, if you do `Pkg.add("SuiteSparse")` in any Julia 1.x version, it will not install this package, because `SuiteSparse` is currently a stdlib in all Julia 1.x versions.


ViralBShah:
I think this really should be removed while it ships as an stdlib.

PR #36875, opened 2021-05-17
The AutoMerge guideline documentation is now automatically generated

DilumAluthge:
The AutoMerge guideline documentation is now automatically generated. Therefore, we don't need to repeat that information here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 [TagBotTriggers-url]: https://github.com/JuliaRegistries/General/actions/workflows/TagBotTriggers.yml
 [TagBotTriggers-img]: https://github.com/JuliaRegistries/General/actions/workflows/TagBotTriggers.yml/badge.svg "TagBot Triggers status"
 [UpdateManifests-url]: https://github.com/JuliaRegistries/General/actions/workflows/update_manifests.yml
-[UpdateManifests-img]: https://github.com/JuliaRegistries/General/actions/workflows/update_manifests.yml/badge.svg "Update Manifests status" 
+[UpdateManifests-img]: https://github.com/JuliaRegistries/General/actions/workflows/update_manifests.yml/badge.svg "Update Manifests status"
 
 General is the default Julia package registry. Package registries are used by Julia's
 package manager [Pkg.jl][pkg] and includes information about packages such as versions,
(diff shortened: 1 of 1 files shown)


ericphanson:
I’m not totally sure about this. I think RegistryCI is more general than General and the RegistryCI docs might evolve to reflect that (eg we’ve talked about making all the checks configurable so other registries can pick and choose). So if we change those docs to reflect that (eg already I think we should mention there that the license check can be turned off) then it might not make as much sense for General.

I also think the list is less visible now and already few people read the readme. Maybe making it shorter by removing these will help with that though.

But I can see why it’s annoying to have it in two places and to need to manually keep it in sync.

DilumAluthge:
> But I can see why it’s annoying to have it in two places and to need to manually keep it in sync.

Yeah, I find it inconvenient to have to maintain two separate sources of documentation. Also, the old system meant that it was easy to have incorrect documentation. For example, once I restored the "URL ends with `/PackageName.jl.git`" check in RegistryCI, I forgot to update the General registry README. So for a long time, the README completely omitted any mention of that check.

DilumAluthge:
> I also think the list is less visible now and already few people read the readme. Maybe making it shorter by removing these will help with that though.

Yeah, it certainly seems that not everyone reads the README before registering a package. That's why, if AutoMerge fails, the comment includes a link to the README.

And in https://github.com/JuliaRegistries/RegistryCI.jl/pull/406, I also added a link to the RegistryCI docs. So it looks like this:

Screen Shot 2021-05-17 at 20 37 04

In theory, this makes it easy for people to get to both the General README and also the RegistryCI docs.

DilumAluthge:
> I’m not totally sure about this. I think RegistryCI is more general than General and the RegistryCI docs might evolve to reflect that (eg we’ve talked about making all the checks configurable so other registries can pick and choose). So if we change those docs to reflect that (eg already I think we should mention there that the license check can be turned off) then it might not make as much sense for General.

One option would be to auto-generate the docs for General as well. E.g. use Documenter instead of having a long README. And in the Documenter docs, we'd auto-generate the list of automerge guidelines.

So we'd autogenerate the docs for General, and we'd autogenerate the docs for RegistryCI. Those would be separate. So in the General docs, we'd just show all the checks. And in the RegistryCI docs, we could include additional information on which checks can be disabled in private registries.

ericphanson:
> One option would be to auto-generate the docs for General as well. E.g. use Documenter instead of having a long README. And in the Documenter docs, we'd auto-generate the list of automerge guidelines.

If we did want to go this route, we could maybe do something so that you don't need to repeat/sync-up settings between `AutoMerge.run` to actually run the automerge and between the docs. It would be cool to have all checks be individually configurable, maybe with Preferences.jl, and then have a docs script that reads those preferences and generates the appropriate docs for the registry (for General or for any other registry). And of course `AutoMerge.run` would use those preferences too, to choose which checks to apply.

I think the other benefit of Preferences is that we wouldn't need to have a zillion keyword arguments to `AutoMerge.run` or thread them through everywhere, since we could just read them in the function that needs to decide which checks to apply.

PR #36880, opened 2021-05-17
remove the misnamed package FreudenthalTriangulation

KristofferC:
This was accidentally registered with the wrong name and no one depends on it (new one is called FreudenthalTriangulations.jl). The old repository is also renamed which means this package might have problems getting installed in the future. Doesn't seem to be much point in keeping this package and it was also requested by the authors that it be removed.

cc @SidhartK @mykelk


SidhartK:
Yep, I approve of removing the old package FreudenthalTriangulation.jl.

DilumAluthge:
Also should be removed from `Registry.toml`?

KristofferC:
Yes

PR #37667, opened 2021-05-27
Migrate BioFetch to BioJulia org

BioTurboNick:


giordano:
Why didn't you simply move the repository?

BioTurboNick:
I did. JuliaRegistrator said I needed to do this.

BioTurboNick:
Unless I misunderstood the problem: https://github.com/BioJulia/BioFetch.jl/commit/e3a12a02dcf0e9f69b519800f9776ae93e68f272#commitcomment-51391189

giordano:
But then you forked that repo to your own account breaking the redirect

giordano:
Your repository already diverged from upstream

BioTurboNick:
Yes? The point is to keep doing development in mine and then PR into BioJulia.

giordano:
You use different branches for that, not the same default branch in a fork.  In general, it's a bad idea to do a pull request from the same branch in a fork

BioTurboNick:
Alright. Not understanding why that's a problem here.

giordano:
Because at the moment the package is being served from your fork instead from BioJulia and since I saw the two `main` branches diverged (your `main` branch is 11 commits ahead, 1 commit behind `BioJulia:main`) I had to check you actually didn't break previous versions.

It looks like previously registered versions are still there so I'm going to merge this

BioTurboNick:
Alright, noted for the future. Thanks!

PR #38298, opened 2021-06-07
Move back DistributedReinforcementLearning

findmyway:
This is a following up of https://github.com/JuliaRegistries/General/pull/35965

I've tested it locally based on https://github.com/JuliaRegistries/General/pull/35965#issuecomment-832721704 and the only tagged version of DistributedReinforcementLearning@v0.1.0 is found.


findmyway:
The `DistributedReinforcementLearning.jl` was only experimental before. Recently I get some  time to work on it again and prepare to integrate it into  RL.jl.
[noblock]

PR #38626, opened 2021-06-11
Yank all registered versions of StatPlots.jl

### LINKED ISSUE #5518: Don't fix now: Eventually unregister StatPlots (when 0.6 is completely out of use)
Is it possible to "unregister" `StatPlots`? Everytime I have to install the package I get confused about whether I should install `StatPlots` or `StatsPlots`.

DilumAluthge:
Fixes https://github.com/JuliaPlots/Plots.jl/issues/5518

It's far too confusing to have both `StatPlots.jl` and `StatsPlots.jl` in the registry.

### DIFF
--- a/S/StatPlots/Versions.toml
+++ b/S/StatPlots/Versions.toml
@@ -1,17 +1,23 @@
 ["0.8.0"]
 git-tree-sha1 = "6216b81c79889743254af713d8cb5f21a18efacd"
+yanked = true
 
 ["0.8.1"]
 git-tree-sha1 = "4555d876c4c2db293dc661c94f26679b788ccb13"
+yanked = true
 
 ["0.8.2"]
 git-tree-sha1 = "e35b968df3937eaee1c8df698bcfd7d047c45110"
+yanked = true
 
 ["0.9.0"]
 git-tree-sha1 = "0f741f9854516b6d8aec55c8c6b6b01805029eb0"
+yanked = true
 
 ["0.9.1"]
 git-tree-sha1 = "2087cd367b8481ace3841d67af708262234134e3"
+yanked = true
 
 ["0.9.2"]
 git-tree-sha1 = "245c50f8a6534bb16ada031e064363f8298b61b9"
+yanked = true


DilumAluthge:
If there are no objections, I'll merge this in a day or two.

fredrikekre:
Do the package maintaners agree and approve of this? I am sure they do, but just good to get it "on paper".

DilumAluthge:
> Do the package maintaners agree and approve of this? I am sure they do, but just good to get it "on paper".

Good point! Let me find out who the best people to ping are.

DilumAluthge:
@mkborregaard @piever @daschw Are any of you maintainers of StatPlots.jl and/or StatsPlots.jl?

If not, would you happen to know who the maintainers are?

BeastyBlacksmith:
can't speak for the others, but I do approve this

daschw (approved):
I also approve

piever:
I also think it's fine to yank releases with the old name

DilumAluthge:
From Slack, Kristoffer seems to lean against this.

oxinabox:
This is the issue on adding something to unlist things
https://github.com/JuliaLang/Pkg.jl/issues/2194

One thing suggested in that issue is the this deprecation/unlisting could also display a message (configured in the registry) to the user if they did type it in rather than tab-completing.
So in this case we would for StatPlot display a message something like "StatPlot was renamed to StatsPlot in 1901, while the code StatPlot will install works, it is a long way behind on new features/bugfixes etc"

KristofferC:
> From Slack, Kristoffer recommends against this.

From my understanding, the reason for doing this is that someone copy-pasted some old code, Pkg installed StatPlots, and downgraded some other packages because of compatibility. There are probably hundreds of packages that haven't been updated in a long while and therefore would do roughly the same with installing. I just don't want it to become a norm that if a package is not updated for a while all its versions get yanked. There are also many packages that depend on this package and it is a little bit weird what happens with all those packages when their dependency has all of its version yanked.

Yanking originally was made to surgically remove individual buggy/unsafe releases so using it for something like this feels like a bit of an abuse of the functionality.

I have, however, done something similar to this when I got asked (emailed) by the authors to do so in https://github.com/JuliaRegistries/General/pull/36880 but that was a package with absolutely zero dependencies which is why it felt more ok.

DilumAluthge:
> I just don't want it to become a norm that if a package is not updated for a while all its versions get yanked

Yeah, I agree that we don't want this to become a norm.

BeastyBlacksmith:
This would close https://github.com/JuliaPlots/StatsPlots.jl/issues/225

giordano:
An alternative is to cap compatibility of the package to, say, Julia v1.6, so that future versions won't see the package.  It's much less disruptive than yanking

DilumAluthge:
> An alternative is to cap compatibility of the package to, say, Julia v1.6, so that future versions won't see the package. It's much less disruptive than yanking

https://github.com/JuliaRegistries/General/pull/39292

DilumAluthge:
I'm closing this PR in favor of https://github.com/JuliaRegistries/General/pull/39292

PR #38675, opened 2021-06-11
Fix TagBot script

christopher-dG:


christopher-dG:
This will reduce the retries to 10 as it's supposed to, but the retry behaviour is still a bit problematic because whichever repos get processed first will get duplicates while the ones last in line will never get reached.

DilumAluthge:
> This will reduce the retries to 10 as it's supposed to, but the retry behaviour is still a bit problematic because whichever repos get processed first will get duplicates while the ones last in line will never get reached.

We should probably just not run TagBotTriggers in a loop.

Really, the only error I want to retry on is the 503 we get when trying to list all General registry PRs. Any other error shouldn't be retried.

PR #38732, opened 2021-06-12
Don't run TagBotTriggers in a bounded retry loop, and set a 30-minute timeout

DilumAluthge:
Follow-up to https://github.com/JuliaRegistries/General/pull/38675

We shouldn't actually be running the entire TagBot Triggers job in a bounded retry loop.

The part we want to run in a bounded retry loop is the step where we try to list all PRs on the General registry, because sometimes the GitHub API returns 5xx for no reason. That part should be in a bounded retry loop, but the rest of the TagBot Triggers job should not be.


giordano:
Can we also add some [`timeout-minutes`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) for good measure?

DilumAluthge:
> Can we also add some [`timeout-minutes`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) for good measure?

Yes! Actually, we should edit all GHA workflows in this repo and set a timeout. What do you think? 60 minutes?

giordano:
60 minutes is definitely better than the default 360, but many jobs here are fairly quick that much lower timeouts (10-20 minutes) make also sense.  I don't know what's the average runtime for the specific job in question.

DilumAluthge:
What we really need is something like the 90th percentile, right? E.g. most AutoMerge jobs finish in 10-20 minutes, but some jobs will take 30-40 minutes because they have a lot of artifacts to download, or a lot of `precompile` statements. We want to minimize the number of legitimate jobs that time out.

DilumAluthge:
Ah I see. Yeah the cron jobs for TagBot Triggers seem to never take more than 20 minutes. Maybe 30 minutes is a good timeout? We can always increase it later if need be.

giordano:
Yeah, 30 minutes sounds like a good timeout then

DilumAluthge:
I've set `timeout-minutes` to 30. If we find that we are hitting the timeout frequently, we can increase it.

PR #38736, opened 2021-06-12
CI: Fix a major typo in the `.ci/instantiate.sh` shell script

DilumAluthge:
See also: #38675


DilumAluthge:
I think this is correct, but it would be good if someone can confirm.

giordano (approved):

PR #38775, opened 2021-06-13
Update urls for owner-transferred repos: ScientificTypes, MLJScientificTypes

ablaom:
cc @DilumAluthge


DilumAluthge (approved):

ablaom:
@DilumAluthge I wonder if we should revert these changes in light of the fact that we also want to change the names of the repos; see https://github.com/JuliaAI/ScientificTypes.jl/issues/125#issue-909967441 . 

Now that I see how name ambiguities are resolved with ownership sub-directories, we should do the name changes, 

ScientificTypes -> ScientificTypesBase
MLJScientificTypes -> ScientificTypes

at the same time as updating the urls?

What is your suggestion?

cc @juliohm

DilumAluthge:
Hmmm. In the General registry, renaming doesn't really work like that. We never actually remove a package name from the registry.

Here is what you would want to do:
1. Create a new package and call it `ScientificTypesBase`. 
2. Copy all of the content of `ScientificTypes` into `ScientificTypesBase`. Now register `ScientificTypesBase` as a new package.
3. Now, delete all contents of the `ScientificTypes` package and replace them with the contents of the `MLJScientificTypes` package. Then make a new breaking release of `ScientificTypes`.
4. Now you can deprecate the `MLJScientificTypes` package.

Does that make sense?

ablaom:
Yes. That makes sense and sounds like the safest.

PR #39444, opened 2021-06-23
Add explicit compat bounds for packages that use Lua_jll

ararslan:
Lua on Windows has the major and minor version as part of the DLL name, so updating the JLL to use a newer Lua version can break downstream packages that expect particular names. Currently the only registered users of Lua_jll are flux_core_jll and mpv_jll.


giordano (approved):
Perfect, thanks!

PR #40088, opened 2021-07-02
Fix ExprTools compat

omlins:
This is a workaround for the indirect dependency ExprTools.jl which in version 0.1.4 has a compat issue with CUDA.jl/GPUCompiler.jl, c.f. JuliaGPU/GPUCompiler.jl#214


I/ImplicitGlobalGrid/Compat.toml
```diff
@@ -1,5 +1,5 @@
 [0]
 CUDA = ["1", "3.1.0-3"]
-ExprTools = "0.1"
+ExprTools = "0.1.0, 0.1.1, 0.1.2, 0.1.3"
```
giordano, line 3: I don't understand, `"0.1"` already allows all of `0.1.0`, `0.1.1`, `0.1.2`, `0.1.3`, there is _**no**_ functional difference between the old compat and the new one.

omlins: thanks for reviewing @giordano .
The issue is that just yesterday ExprTools 0.1.4 was released and this version causes troubles with CUDA.jl/GPUCompiler.jl, c.f. JuliaGPU/GPUCompiler.jl#214 . So, the objectiv is to exclude 0.1.4. That's why I explictly noted `"0.1.0, 0.1.1, 0.1.2, 0.1.3"`

giordano: Well, what you did doesn't do what you expect: `0.1.0` means "0.1.0 _and all subsequent versions in the 0.1 series_", `0.1.1` means "0.1.1 _and all subsequent versions in the 0.1 series_", `0.1.2` means "0.1.2 _and all subsequent versions in the 0.1 series_", etc...  See https://pkgdocs.julialang.org/v1/compatibility/#Version-specifier-format-1.  That's why, as I said, there is _**no**_ functional difference between
```toml
ExprTools = "0.1"
```
and
```toml
ExprTools = "0.1.0, 0.1.1, 0.1.2, 0.1.3"
```

Maybe you wanted to do something like
```suggestion
ExprTools = "0.1.0-0.1.3"
```
instead

omlins: Oh yes, right, only if the mayor and minor version digit was 0, then different patch versions are considered incompatible. So, yes, what you suggest is what I wanted. Thanks!

giordano:
Actually, this pull request isn't really needed: if you fix it in https://github.com/eth-cscs/ImplicitGlobalGrid.jl/blob/a3e1efd2287df45d1436d8014c5fbfcc656ad3cb/Project.toml#L10 and submit registration again without changing version number, then #40070 will be updated automatically

giordano:
But note that if you want to support Julia 1, in your `Project.toml` you want to do
```toml
ExprTools = "=0.1.0, =0.1.1, =0.1.2, =0.1.3"
```

omlins:
> 
> 
> Actually, this pull request isn't really needed: if you fix it in https://github.com/eth-cscs/ImplicitGlobalGrid.jl/blob/a3e1efd2287df45d1436d8014c5fbfcc656ad3cb/Project.toml#L10 and submit registration again without changing version number, then #40070 will be updated automatically

Done

PR #40100, opened 2021-07-02
Change `compat` for DiffEqFlux

dannys4:
Address [583](https://github.com/SciML/DiffEqFlux.jl/issues/583) in DiffEqFlux.jl

### DIFF
--- a/D/DiffEqFlux/Compat.toml
+++ b/D/DiffEqFlux/Compat.toml
@@ -177,12 +177,10 @@ Optim = "1"
 ["1.4-1.12.0"]
 DiffEqSensitivity = "6.7.0-6"
 
-["1.40.0-1"]
-DiffEqSensitivity = "6.52.0"
-
 ["1.40-1"]
 Cassette = "0.3.7-0.3"
 Reexport = ["0.2", "1"]
+DiffEqSensitivity = "6.52.0"
 
 ["1.41-1"]
 GalacticOptim = "1-2"
--- a/D/DiffEqFlux/Compat.toml
+++ b/D/DiffEqFlux/Compat.toml
@@ -77,7 +77,7 @@ Optim = "0.20-0.21"
 ["1.12-1.17"]
 DiffEqBase = "6.36.0-6"
 
-["1.12.1-1"]
+["1.12.1-1.39"]
 DiffEqSensitivity = "6.19.0-6"
 
 ["1.13"]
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Can you explain the specific `[compat]` change that you want to make?

dannys4:
Sure-- DiffEqFlux v1.40.0 and v1.40.1 should be relying on at least DiffEqSensitivity 6.52.0. Currently they don't and it's causing problems downstream.

DilumAluthge:
Alright, so it sounds like for DiffEqFlux v1.40.0, we need to "add a [compat] entry" that would have looked like this in a `Project.toml` file:
```toml
DiffEqSensitivity = "6.52.0"
```

And similarly for DiffEqFlux v1.40.1.

Is that correct?

And are DiffEqFlux v1.40.0 and DiffEqFlux v1.40.1 the only two versions that we need to do this for?

dannys4:
According to @ChrisRackauckas, yes, all of that is correct.

DilumAluthge:
Alright, I've made the desired changes.

Here is the code that I used to generate these changes:
```julia
import Pkg
import RegistryTools

tmp_dir = mktempdir(; cleanup = true)

cd(tmp_dir)

run(`git clone git@github.com:JuliaRegistries/General.git`)

cd("General")

compat_filename = "D/DiffEqFlux/Compat.toml"

compat_dict = RegistryTools.Compress.load(compat_filename)

for ver in [v"1.40.0", v"1.40.1"]
    old_compat_string = compat_dict[ver]["DiffEqSensitivity"]
    old_compat_spec = Pkg.Versions.VersionSpec(old_compat_string)
    additional_compat = Pkg.Versions.semver_spec("6.52.0")
    new_compat_spec = intersect(old_compat_spec, additional_compat)
    compat_dict[ver]["DiffEqSensitivity"] = string(only(new_compat_spec.ranges))
end

RegistryTools.Compress.save(compat_filename, compat_dict)

println("# begin contents of the $(compat_filename) file")
println()
println(read(compat_filename, String))
println()
println("# end contents of the $(compat_filename) file")
```

dannys4:
If you don't mind, do I have your permission to modify this and put it up as a gist? It could come in handy for me/others in the future.

DilumAluthge:
Sure!

DilumAluthge:
Feel free to link the gist in this thread.

dannys4:
I just put up the gist [here](https://gist.github.com/dannys4/6daf9e59be77a2bdfcc6be900dc554df). Thanks for your help!

giordano:
Did this really fix the issue in https://github.com/SciML/DiffEqFlux.jl/issues/583?  #40104 is failing with the same error: https://github.com/JuliaRegistries/General/pull/40104/checks?check_run_id=2974674882#step:16:1489.  It doesn't look like this solved much (or all)

ChrisRackauckas:
It looks like the lower bound was wrong on v6.41 as well.  https://github.com/JuliaRegistries/General/pull/40100/commits/09278df710c6e2305859e0379f7c01b3f5d38437#diff-206c28ae05ab601be87e5c995f5cab97c2494fec64c98fa17313aa7e9d1fef81R188 should get changed. Oof, sorry. I thought the latest was tagged with the new lower bound.

PR #40170, opened 2021-07-03
rm tips for registry maintainers

ericphanson:
This doesn't belong in the General registry's README and the functionality is already covered in RegistryCI's API docs (https://juliaregistries.github.io/RegistryCI.jl/stable/public/#RegistryCI.AutoMerge.run).

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,19 +174,6 @@ not a curated list of Julia packages. In particular this means that:
    you install through the General registry -- you are responsible for reviewing your
    code dependencies.
 
-## Tips for registry maintainers
-
-### Enabling/disabling AutoMerge
-
-To enable/disable automerge, make a pull request to edit the
-[`.github/workflows/automerge.yml`](.github/workflows/automerge.yml) file. Specifically, you want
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -174,19 +174,6 @@ not a curated list of Julia packages. In particular this means that:
    you install through the General registry -- you are responsible for reviewing your
    code dependencies.
 
-## Tips for registry maintainers
```
giordano, line 177: Maybe add a pointer to RegistryCI docs?

ericphanson: Hm, they are linked earlier in the automerge section, though that's kind of a different context. I'm not totally sure General is the right place though-- maybe the [Pkg.jl docs on registries](https://pkgdocs.julialang.org/v1/registries/) should be updated to mention RegistryCI (and probably LocalRegistry.jl too)

ericphanson: I made a PR to do that here: https://github.com/JuliaLang/Pkg.jl/pull/2659

DilumAluthge (approved):

PR #40171, opened 2021-07-03
Update docs on renaming existing package

ericphanson:
To mention the old package's URL should be updated. cc @oxinabox who wrote the original version

PR #40193, opened 2021-07-03
Update Slack.jl repo URL

ericphanson:
since it was moved to a different GitHub organization. This way we don't need to rely on GitHub redirects. cc @logankilpatrick

ref https://github.com/JuliaRegistries/General/pull/40171#issuecomment-873480346


DilumAluthge:
I'm loving all of this clean-up!

logankilpatrick:
Thanks for this!

PR #40199, opened 2021-07-04
Update URL for ProximalOperators, ProximalAlgorithms, StructuredOptimization, AbstractOperators

lostella:
Updating URLs not to rely on GitHub redirect. cc @ericphanson (apologies for the quick follow-up after #40192, I was waiting for a second change of org, which happened today, before opening the PR to the registry)


ericphanson:
Ah no problem, thanks for the update here!

PR #40225, opened 2021-07-04
Change compat for ForwardDiff -- upper bound DiffRules

mcabbott:
This adds an upper bound so that all existing versions of ForwardDiff will not use DiffRules 1.1, and stop at 1.02. This applies to ForwardDiff v0.10.8 to v0.10.18 (latest). 

The reason to do so is that DiffRules supports N-ary rules, but ForwardDiff throws an error if given one, and DiffRules v1.1 added a few. This isn't even a new export from DiffRules, it's a documented feature, although the package contained no rules using it. That makes me think it's not a breaking change to DiffRules, but just a bug in some versions of ForwardDiff. So the aim of this PR is to prevent them from installing a version which triggers that.

(The immediate fix is that https://github.com/JuliaDiff/DiffRules.jl/pull/60 and DiffRules 1.2.1 has removed these rules, temporarily. So installing the latest should not at present break things.)


mcabbott:
Did I get this right? When I try installing on a fresh environment, I seem to get DiffRules v1.2.1 which is outside "1-1.02". 
```
(@v1.8) pkg> activate --temp
  Activating new project at `/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_ecsvaR`

(jl_ecsvaR) pkg> up
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
  No Changes to `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_ecsvaR/Project.toml`
  No Changes to `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_ecsvaR/Manifest.toml`

(jl_ecsvaR) pkg> add ForwardDiff
   Resolving package versions...
    Updating `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_ecsvaR/Project.toml`
  [f6369f11] + ForwardDiff v0.10.18
    Updating `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_ecsvaR/Manifest.toml`
  [d360d2e6] + ChainRulesCore v0.10.9
  [bbf7d656] + CommonSubexpressions v0.3.0
  [34da2185] + Compat v3.31.0
  [163ba53b] + DiffResults v1.0.3
  [b552c78f] + DiffRules v1.2.1
...
```

PR #40252, opened 2021-07-05
Change compat for ForwardDiff -- upper bound DiffRules -- II

mcabbott:
This corrects a typo in #40225, which said

> This adds an upper bound so that all existing versions of ForwardDiff will not use DiffRules 1.1, and stop at 1.02. 

Except that it should have been 1.0.2, which excludes 1.2, not 1.02, which includes 1.2.


F/ForwardDiff/Compat.toml
```diff
@@ -43,7 +43,7 @@ DiffRules = ["0.0.4-0.0.10", "0.1"]
 
 ["0.10.8-0"]
 DiffResults = ["0.0.1-0.0.4", "1.0.1-1"]
-DiffRules = ["0.0.4-0.0.10", "0.1", "1-1.02"]
+DiffRules = ["0.0.4-0.0.10", "0.1", "1-1.0.2"]
```
giordano, line 46: Why not
```suggestion
DiffRules = ["0.0.4-0.0.10", "0.1", "1-1.0"]
```
?

mcabbott: I can believe that would work equally well, didn't think of it.

PR #40707, opened 2021-07-11
 [libcgal_julia_jll] Fix compat for version 0.16.x

rgcv:
Sorry, I botched JuliaRegistries/General#40705, was trying to reset my fork's master branch for when I was testing the registry locally.  Rookie mistake, I know.


L/libcgal_julia_jll/Compat.toml
```diff
@@ -4,18 +4,19 @@ julia = "1"
 ["0.10-0.11"]
 libcxxwrap_julia_jll = "0.7.1"
 
+["0.16-0"]
+CGAL_jll = "5.2.0-5.2"
```
giordano, line 8: Just to make sure there is nothing wrong: why don't you specify `CGAL_jll` for v0.10-0.11?  Apart from that, this looks ok to me

rgcv: I'll have to recall which versions they were compatible too.  I was just afraid I'd tamper with more than I needed to.  I'll get back to this ASAP.

rgcv: Pushed a change that added a compat entry for `CGAL_jll` for version up to 0.15 according to deployments to JuliaBinaryWrappers.

PR #41992, opened 2021-08-02
[MiniLoggers] Transfer to JuliaLogging organization

Arkoniak:


giordano:
It looks like you broke the redirect already

PR #42250, opened 2021-08-05
[DEDataArrays] Remove `DiffEqBase` from dependencies

giordano:
This isn't actually used and it causes a circular dependency in
https://github.com/SciML/DiffEqBase.jl/pull/690.  See also
https://github.com/SciML/DEDataArrays.jl/pull/4.  CC: @ChrisRackauckas.


giordano:
Ok, let's wait to fix this in the package

PR #42597, opened 2021-08-11
fix compat for WeakRefStrings 1.0.0

KristofferC:
Claims to support Julia 1.0 but don't. Yanking the release doesn't help because yanking was implemented in Pkg in 1.1.


KristofferC:
Confirmed this installs the correct versions on the various Julia versions.

PR #43737, opened 2021-08-29
add `feed` workflow

ericphanson:
This adds a workflow that runs whenever a PR is labelled. If the label is "new package", then it uses a Slack endpoint kindly supplied by @logankilpatrick to update the `#new-packages-feed` channel (currently I believe the endpoint is pointed at `#stackoverflow-feed` but it's being updated). I want to also add a Zulip feed as a second job to this workflow (in a future PR).

I think this is useful for two reasons:

1. It provides a simple way for users on Slack/Zulip to keep up with all the awesome new packages being registered
2. It might help more folks get involved with the registry, to help chime in with ideas about names, contribute when they think a name-distance clash is OK or not, and help point new package devs towards [best practices](https://github.com/JuliaRegistries/General#should-i-register-my-package). This is why I want to trigger the update when the PR is created, not when it is merged.

### DIFF
--- a/.github/workflows/feed.yml
+++ b/.github/workflows/feed.yml
@@ -27,7 +27,12 @@ jobs:
                           "title_link" => pr_url,
                           "text" => pr_url,
                       )])
-        response = sendattachmenttoslack(data, ENV["SLACK_ENDPOINT"])
+
+        endpoint = ENV["SLACK_ENDPOINT"]
+        if !startswith(endpoint, "/")
+            endpoint = string("/", endpoint)
+        end
+        response = sendattachmenttoslack(data, endpoint)
         @info "Slack said: $response"
       env:
         SLACK_ENDPOINT: ${{ secrets.JULIALANGSLACKENDPOINT }}
--- a/.github/workflows/feed.yml
+++ b/.github/workflows/feed.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     types: [ labeled ]
 jobs:
-  build:
+  update-feed:
     if: ${{ github.event.label.name == 'new package' }}
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/feed.yml
+++ b/.github/workflows/feed.yml
@@ -1,0 +1,35 @@
+# This workflow updates feeds when the label `new package` is applied to PRs.
+name: Send to feeds
+on:
+  pull_request:
+    types: [ labeled ]
+jobs:
+  build:
+    if: ${{ github.event.label.name == 'new package' }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: julia-actions/setup-julia@latest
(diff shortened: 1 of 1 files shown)


DilumAluthge:
You could also add "opened" to the `types`, just to be safe. I'm not sure if `labeled` gets called when the label is part of the original PR creation.

It would be good for @logankilpatrick to review the specifics, as I'm not too familiar with the Slack bots.

DilumAluthge:
Also, I don't think it worked.

ericphanson:
Wow, Slack replied with a whole webpage: https://github.com/JuliaRegistries/General/pull/43737/checks?check_run_id=3455592925#step:3:77

DilumAluthge:
Can you add code that prepends `/` to the endpoint environment variable? I don't think I included a leading `/` when I input the secret.

ericphanson:
This is the webpage it returned, rendered:

[Slack API _ Slack.pdf](https://github.com/JuliaRegistries/General/files/7072920/Slack.API._.Slack.pdf)

So just a generic page, not saying anything about our request.

DilumAluthge:
I think this PR doesn't have access to secrets, because it was made from a fork. So the secret is empty, and we are just hitting hooks.slack.com.

DilumAluthge:
@ericphanson Can you make a new PR that is not from a fork?

ericphanson:
> You could also add "opened" to the types, just to be safe. I'm not sure if labeled gets called when the label is part of the original PR creation.

Hm, good point. I'd rather keep it to just `labeled` for now and see if it works, and then update it if it doesn't. The reason is that if it's not triggered by a `labeled` event, I don't know if there's an easy way to get the label name.

logankilpatrick:
I also think the secret is wrong right now, still trying to workout what it should be

DilumAluthge:
> > You could also add "opened" to the types, just to be safe. I'm not sure if labeled gets called when the label is part of the original PR creation.
> 
> Hm, good point. I'd rather keep it to just `labeled` for now and see if it works, and then update it if it doesn't. The reason is that if it's not triggered by a `labeled` event, I don't know if there's an easy way to get the label name.

Sounds good to me.

ericphanson:
> @ericphanson Can you make a new PR that is not from a fork?

Ohh, maybe the fork doesn't get the secret, right. Oops, will do that

DilumAluthge:
From the log:

```
env:
24
    SLACK_ENDPOINT: 
25
    PR_NAME: add `feed` workflow
26
    PR_NUMBER: 43737
    ```

If we were getting the secret, it would have three asterisks next to `SLACK_ENDPOINT`. We're not getting the secret because the PR is from a fork.

We should add a conditional that only runs this workflow if the PR is not from a fork. That still includes all Registrator.jl PRs, which is what we want.

PR #43740, opened 2021-08-29
add `feed` workflow, take 2

ericphanson:
Redo of #43737 but from a branch, not a fork.

### DIFF
--- a/.github/workflows/feed.yml
+++ b/.github/workflows/feed.yml
@@ -15,7 +15,7 @@ jobs:
       shell: julia --color=yes {0}
       run: |
         endpoint = ENV["SLACK_ENDPOINT"]
-        if isempty(ENDPOINT)
+        if isempty(endpoint)
           @info "Endpoint not specified; missing secret? Exiting."
           exit(0)
         end
--- a/.github/workflows/feed.yml
+++ b/.github/workflows/feed.yml
@@ -14,6 +14,12 @@ jobs:
     - name: Send to Slack
       shell: julia --color=yes {0}
       run: |
+        endpoint = ENV["SLACK_ENDPOINT"]
+        if isempty(ENDPOINT)
+          @info "Endpoint not specified; perhaps this PR was from a fork. Exiting."
+          exit(0)
+        end
+        
         using Pkg
         Pkg.activate(mktempdir())
         Pkg.add(name="Slack", version="0.1")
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Long term, it will probably be better to use a manifest specific for the `feed` workflow? Just so that if things break, it's easy to roll back to the last good manifest?

Doesn't need to be done in this PR.

DilumAluthge (changes requested):
Can you add a conditional that only runs this action if the PR is not made from a fork?

DilumAluthge:
@logankilpatrick would be good for you to double check the code.

DilumAluthge (approved):
This should have been an approval. LGTM as long as you add the "don't run on forks" logic before merging.

ericphanson:
> Can you add a conditional that only runs this action if the PR is not made from a fork?

I'm not sure how to do this :(. I added a julia-side conditional to bail if the endpoint is empty, but I think ideally it would be earlier in the workflow like where I check if the label is the right one.

logankilpatrick (approved):
Approved once the endpoint secret is updated.

logankilpatrick:
Ideally Slack.jl would check this, if you add a PR I can accept that

DilumAluthge:
> > Can you add a conditional that only runs this action if the PR is not made from a fork?
> 
> I'm not sure how to do this :(. I added a julia-side conditional to bail if the endpoint is empty, but I think ideally it would be earlier in the workflow like where I check if the label is the right one.

This conditional gives you "PR is not from a fork":
> if: github.event_name == 'pull_request' && github.repository == github.event.pull_request.head.repo.full_name

Stick that in your existing `if` with an `&&` in between.

DilumAluthge:
Secret updated.

DilumAluthge:
You can take the Julia-side conditional out once you've assembled the massive unholy YAML conditional.

ericphanson:
> You can take the Julia-side conditional out once you've assembled the massive unholy YAML conditional.

😄. I took out the `github.event_name == 'pull_request'` since we only run on PRs, and left in the Julia-side one but updated the language to not mention forks. I thought it was still helpful since if we get the secret's name wrong it will tell us more clearly than those giant webpage dumps we were getting in the other PR.

.github/workflows/feed.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: julia-actions/setup-julia@latest
+      with:
+        version: 1.6.2
+    - name: Send to Slack
+      shell: julia --color=yes {0}
+      run: |
+        endpoint = ENV["SLACK_ENDPOINT"]
+        if isempty(ENDPOINT)
```
DilumAluthge, line 18:
```suggestion
        if isempty(endpoint)
```

PR #43823, opened 2021-08-30
Yank TriangularSolve version 0.1.4

DilumAluthge:
TriangularSolve 0.1.4 has an incorrect `[compat]` entry.


DilumAluthge:
@chriselrod Can you confirm that this is correct?

chriselrod:
Yes, and then what's the next step?
I just tag another 0.1.4?

DilumAluthge:
> Yes, and then what's the next step?
> I just tag another 0.1.4 with the correct compat?

You'll tag 0.1.5 with the correct compat.

PR #44231, opened 2021-09-05
AlgebraicNumbers - change package location

fkastner:
Point the registry to the original repository instead of a fork.


fkastner:
I just pushed the contents of the fork to a new branch on the original repo, so all previous versions are available from that repo. Thanks to  @GunnarFarneback for providing the feedback on Zulip.
This should be good, now.

giordano:
I confirm on this branch I could install all versions 0.1.2-5

PR #44388, opened 2021-09-07
Update package URL for OutlierDetection.jl

davnn:
The package now belongs to an org instead of an individual contributor.


davnn:
@StefanKarpinski @DilumAluthge would it be possible to merge this PR so that we can update the PKG? Thanks!

StefanKarpinski:
Sure. Running CI to make sure. But looks good to me.

PR #45426, opened 2021-09-23
Add a stream to Zulip bot that tracks breaking changes to packages

rayegun:
I'm not sure how to check that this works, nor am I a GH actions expert (happy to learn ofc).

@DilumAluthge @ericphanson 

E: Also not sure if something has to be done on the Zulip end (create stream?).

Testing data:

- Registering package: TestPackage
- Repository: https://github.com/ericphanson/DoesNotExist.jl
- Version: v0.1.0
- Description: This is a test description.
- Commit: 976c26fc6165d5501fb765048fd18d2e30181a94
- Git reference: v0.1.0
- Release notes:

> This is my awesome package!
> My release notes have multiple lines.

### DIFF
--- a/.github/workflows/breaking_change_feed.yml
+++ b/.github/workflows/breaking_change_feed.yml
@@ -1,4 +1,4 @@
-# This workflow updates feeds when the label `new package` is applied to PRs.
+# This workflow updates feeds when the label `BREAKING` is applied to PRs.
 name: Send breaking to Zulip
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


ericphanson:
ah, it's hard to test it because the PR comes from a fork, so it doesn't get access to the secrets. Let me just give a good look then and we can test it live 😈 . (Really should be hard to break anything big from a separate workflow not touching any automerge stuff etc).

.github/workflows/breaking_change_feed.yml
```diff
@@ -0,0 +1,41 @@
+# This workflow updates feeds when the label `new package` is applied to PRs.
```
ericphanson, line 1:
```suggestion
# This workflow updates feeds when the label `BREAKING` is applied to PRs.
```

ericphanson (approved):
LGTM

PR #45588, opened 2021-09-27
limit JET's compatibilities to v1.6

aviatesk:
I'd like to limit Julia version compatibilities of currently released JET versions.
And future versions of JET will only support v1.7.

/cc @timholy does this look reasonable to you ?


timholy:
LGTM, but I am not an expert on these matters so perhaps @KristofferC?

KristofferC:
I think this should work. But it is always good to test these things out locally by checking out this branch and trying out the package manager to see that things behave as expected.

Also, is JET completely broken on 1.7? Otherwise, it might make sense to wait with this until the 1.7 release for JET is out.

aviatesk:
JET 0.5.0 is ready and it should work with v1.7 and higher.
I will make a minor release and then update this branch.

aviatesk:
@KristofferC can we move this forward ? I confirmed this worked locally.

PR #47659, opened 2021-10-28
Re-enable CI on Julia nightly

### LINKED ISSUE #47158: CI fails for Nightly build of Julia due to compat requirement for OpenBLAS
It is not blocking, but CI fails for Nightly build of Julia when I registered packages today. It comes from compat requirements of OpenBLAS

```
ERROR: Unsatisfiable requirements detected for package OpenBLAS_jll [4536629a]:
 OpenBLAS_jll [4536629a] log:
 ├─possible versions are: 0.3.17 or uninstalled
 └─restricted to versions 0.3.13 by an explicit requirement — no versions left
```

Can be found in #47157 and #47155 (at least)

DilumAluthge:
Closes #47158

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,9 +40,9 @@ jobs:
       - run: mv .ci/Manifest.${{ steps.manifest_version.outputs.manifest_version }}.toml .ci/Manifest.toml
       - run: rm -rf .ci/Manifest.*.toml
       - run: chmod 400 .ci/Project.toml
-        if: matrix.version != "nightly"
+        if: ${{ matrix.version != 'nightly' }}
       - run: chmod 400 .ci/Manifest.toml
-        if: matrix.version != "nightly"
+        if: ${{ matrix.version != 'nightly' }}
       - name: Cache artifacts
         uses: actions/cache@1a9e2138d905efd099035b49d8b7a3888c653ca8 # v2.1.5
         env:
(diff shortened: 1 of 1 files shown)


IanButterworth:
Hopefully fixed on nightlies now. Can someone rerun the nightly job?

DilumAluthge:
@IanButterworth Why is `precompile` trying to write to the project file?

IanButterworth:
> Why is precompile trying to write to the project file?

I tried locally and there were no changes made to the Project (confirmed by git)

I then set the project as read-only and hit
```
(.ci) pkg> resolve
ERROR: SystemError: opening file "/Users/ian/Documents/GitHub/General/.ci/Project.toml": Permission denied
```

I guess the heuristic here isn't actually detecting that there's no changes
https://github.com/JuliaLang/Pkg.jl/blob/6f8e36db88ed0c7727b5616b2baa58787cdc4e95/src/Types.jl#L1014

IanButterworth:
Why is `Pkg.precompile()` directly called? 
`Pkg.instantiate()` is called so `precompile` will be automatically called. And that `precompile` won't hit the `resolve` that's now called when it's run manually

IanButterworth:
Actually, I think `Pkg.instantiate()` should still allow `Pkg.precompile()` to resolve so that versioned stdlibs can be updated, so my question is moot and I need to fix that on Pkg

PR #47879, opened 2021-11-01
Add explicit compat bounds for packages that use GEOS_jll

visr:
Addresses https://github.com/JuliaGeo/GDAL.jl/issues/126

I used #39444 as an example for this edit.


giordano:
Perfect, thanks a lot!  After this, a PR to addd the same compat bounds in Yggdrasil would be the great (`Dependency("GEOS_jll"; compat="~3.9")`), but please use `[skip ci] [skip build]` in the commit message because we don't want to rebuild the packages, it's useless

PR #48533, opened 2021-11-10
try strengthen the idea that General is not really meant for personal utility packages

KristofferC:
Based on some recent happenings.  This is very vague for now since it was quite hard to specify it more.  Suggestions for improvements are very welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,8 @@ to setup **tests** (see
 [the Pkg.jl docs](https://pkgdocs.julialang.org/v1/creating-packages/#Adding-tests-to-the-package)
 and the [Test stdlib docs](https://docs.julialang.org/en/v1/stdlib/Test/)), which
 can be automatically run by free **continuous integration** services such as GitHub Actions.
+* Also, note that the General registry is not a place for for "personal packages" that consists of
+collections of "utility functions".
 
 Packages like [PkgTemplates.jl](https://github.com/invenia/PkgTemplates.jl) or
 [PkgSkeleton.jl](https://github.com/tpapp/PkgSkeleton.jl) provide easy ways to setup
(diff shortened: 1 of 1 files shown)


DilumAluthge (approved):

fredrikekre (approved):

README.md
```diff
@@ -97,6 +97,8 @@ to setup **tests** (see
 [the Pkg.jl docs](https://pkgdocs.julialang.org/v1/creating-packages/#Adding-tests-to-the-package)
 and the [Test stdlib docs](https://docs.julialang.org/en/v1/stdlib/Test/)), which
 can be automatically run by free **continuous integration** services such as GitHub Actions.
+* Also, note that the General registry is not a place for for "personal packages" that consists of
```
DilumAluthge, line 100:
```suggestion
* Also, note that the General registry is not a place for "personal packages" that consist of
```

GunnarFarneback:
I think it could be worth pointing out that for packages which are not useful to the general public, you have the option of using Pkg with unregistered packages or to create your own registry. Sorry, don't have any good formulations to suggest.

ericphanson:
> I think it could be worth pointing out that for packages which are not useful to the general public, you have the option of using Pkg with unregistered packages or to create your own registry. Sorry, don't have any good formulations to suggest.

Yeah, I think this would be good too. We mention it just above, but we could link that here. Maybe something like

```
Remember that you can [use a package without registering it](#do-i-need-to-register-a-package-to-install-it),
or even [create your own registry](https://pkgdocs.julialang.org/dev/registries/#Creating-and-maintaining-registries)
for personal packages.
```

On Slack @mcabbott wrote

> “The right time to register is usually once you have enough functionality up and running for it to be useful to others. It need not be completely finished work, but registering an empty shell which you might fill in, or a class project you are sure to abandon, is inappropriate. Doing so imposes work on the community, without benefit to others.”

which could be something else useful to put into the guidance, although it's a slightly different point than this one.

KristofferC:
Did one more take, tried to add a hint about LocalRegistry.jl but maybe I should just have directly linked to https://pkgdocs.julialang.org/dev/registries/#Creating-and-maintaining-registries

DilumAluthge:
Let's go ahead and at least get this in in its current form. We should continue to make improvements and add further additions in subsequent PRs.

PR #48581, opened 2021-11-10
Upper bound Memento to 1.2 for CloudWatchLogs

rofinn:
I'm not sure if I'm doing this right, but my understanding is that retroactively upperbound dependencies for a package requires manually editing the Compat.toml?

### DIFF
--- a/C/CloudWatchLogs/Compat.toml
+++ b/C/CloudWatchLogs/Compat.toml
@@ -42,7 +42,7 @@ Memento = "0.12"
 MbedTLS = "0.5-0.7"
 
 ["1.2-1"]
-Memento = ["0.13", "1"]
+Memento = ["0.13", "1-1.2"]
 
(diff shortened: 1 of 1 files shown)


fredrikekre:
Memento made a breaking release without bumping major version or what happened here? Wouldn't it be better to revoke the faulty version of Memento then?

rofinn:
More that CloudWatchLogs.jl made a faulty assumption about the timezone of an internal datetime field. This was never getting hit because we previously defaulted to an explicit `ZonedDateTime`. Memento is changing the default behaviour to add timezone information later in the pipeline and assume all datetimes are localzone.

https://github.com/invenia/CloudWatchLogs.jl/blob/master/src/event.jl#L10

rofinn:
Does the above response make sense or do you recommend just updating CWL.jl and maybe backporting the fix to previous releases?

PR #48840, opened 2021-11-16
Revert not setting Project.toml as read-only on nightly

IanButterworth:
@DilumAluthge this should be fixed now as https://github.com/JuliaLang/Pkg.jl/pull/2826 has hit the nightlies


DilumAluthge (approved):

DilumAluthge:
I guess we need to keep allowing the `Manifest.toml` to be read-write on nightly, right? So that it can update stdlib versions in the manifest when they are updated in nightly?

IanButterworth:
Yeah exactly

PR #49260, opened 2021-11-23
Make NDTensors a subdir package of ITensors repository

mtfishman:
Hopefully this is correct. I bumped the version of NDTensors from the last time it was registered, should I include that change in this PR as well?


mtfishman:
@DilumAluthge (not sure if you are the right person to ping for this kind of thing).

PR #49308, opened 2021-11-24
[KaHyPar_jll] Fix compat of `boost_jll`

giordano:
Boost breaks ABI in every single version, so we must always specify the compat
bounds with it.  See https://github.com/kahypar/KaHyPar.jl/issues/19.


SebastianSchlag:
Thanks for fixing this @giordano! 👍

GiggleLiu:
Thank you @giordano !

PR #49565, opened 2021-11-29
Yank ECOS@0.12.4

odow:
ECOS 0.12.4 was tagged as a patch release but it contained breaking changes. A new 0.13 has been released which should mean that this version isn't installed, but yanking for posterity.

x-ref: https://github.com/jump-dev/ECOS.jl/issues/123


odow:
Thanks @DilumAluthge

PR #49611, opened 2021-11-30
Update lower bounds for EllipsisNotation version 1.1.2

Tokazama:
Update lower bounds of 1.1.2 to ArrayInterface 3.1 and below. @ChrisRackauckas, I'm assuming this is the lower bounds you referred to [here](https://github.com/ChrisRackauckas/EllipsisNotation.jl/issues/34).


ChrisRackauckas:
No, it needs a lower bound of ArrayInterface 3.2 otherwise the method you're using is not defined. So v1.1.2 should essentially get a registry update to have the same bounds as v1.1.3. 

@DilumAluthge created tools to help for this you might want to use.

E/EllipsisNotation/Compat.toml
```diff
@@ -11,4 +11,4 @@ julia = "1.5.0-1"
 ArrayInterface = "3"
 
 ["1.1.3-1"]
```
ChrisRackauckas, line 13:
```suggestion
["1.1.2-1"]
```

E/EllipsisNotation/Compat.toml
```diff
@@ -11,4 +11,4 @@ julia = "1.5.0-1"
 ArrayInterface = "3"
```
ChrisRackauckas, line 11: then this would need to get changed to 1.1.1

Tokazama: Shouldn't we change the bounds of the last entry to "1.1.3-1" instead, because 1.1.2 was the last version where we needed to be below ArrayInterface=3.2

ChrisRackauckas: No, the issue is that it requires being above 3.2. That's why 1.1.3 was made.

ChrisRackauckas:
Did you check this registry works well locally? If so, I'll merge.

fredrikekre:
Is the intention to make EllipsisNotation 1.1.2 have the same compat on ArrayInterface as EllipsisNotation 1.1.3? In that case I think:
```diff
diff --git a/E/EllipsisNotation/Compat.toml b/E/EllipsisNotation/Compat.toml
index dab59d0223..25e80c634f 100644
--- a/E/EllipsisNotation/Compat.toml
+++ b/E/EllipsisNotation/Compat.toml
@@ -7,8 +7,8 @@ julia = "1"
 ["1.1-1"]
 julia = "1.5.0-1"
 
-["1.1-1.1.2"]
+["1.1-1.1.1"]
 ArrayInterface = "3"
 
-["1.1.3-1"]
+["1.1.2-1"]
 ArrayInterface = "3.2.0-3"
 ```

**Comment by ChrisRackauckas** [comment]
Yes. What happened was 1.1.2 had a bad lower bound that effected hundreds of downstream packages, so 1.1.3 was an immediate hotfix. This is now cleanup to make sure the version with the bad bound no longer exists in some sense. Still a bit time sensitive

**Comment by fredrikekre** [comment]
Okay. The diff above does that. Instead of manually trying to figure out things like this, have a look at `RegistryTools.Compress.load` which loads the compat and `RegistryTools.Compress.save` which recompresses.

File: E/EllipsisNotation/Compat.toml [hunk 3]
```diff
@@ -7,8 +7,8 @@ julia = "1"
 ["1.1-1"]
 julia = "1.5.0-1"
 
-["1.1-1.1.2"]
+["1.1.1"]
```
fredrikekre, line 10:
```suggestion
["1.1-1.1.1"]
```

E/EllipsisNotation/Compat.toml
```diff
@@ -7,8 +7,8 @@ julia = "1"
...
 ArrayInterface = "3"
 
-["1.1.3-1"]
-ArrayInterface = "3.2.0-3"
+["1.1.2-1"]
+ArrayInterface = "3.2"
```
fredrikekre, line 14:
```suggestion
["1.1.2-1"]
ArrayInterface = "3.2.0-3"
```

ChrisRackauckas:
Awesome, thanks a ton. Yeah I was out yesterday but needed a bit of a team effort to handle the breakage emails. This looks great.

PR #51268, opened 2021-12-27
Use CLI git in CI jobs

giordano:
This should at least speed up jobs running with Julia v1.7+ when libgit2 is the culprit, although we see that even the initial checkout step is occasionally horribly slow, which also uses CLI git.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -9,6 +9,8 @@ on:
     # We will choose to run the fallback job every 4 hours.
     - cron:  '0 */4 * * *'
   workflow_dispatch:
+env:
+    JULIA_PKG_USE_CLI_GIT: true
 jobs:
   AutoMerge:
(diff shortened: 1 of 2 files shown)


DilumAluthge:
I also think that GitHub might be having some issues, because I'm seeing on multiple platforms (GitHub Actions, and also our self-hosted Buildkite agents) that network traffic to/from GitHub seems to be slow over the last 24 hours.

DilumAluthge (approved):

giordano:
image

Sigh.......

DilumAluthge:
Seems like further evidence of an issue on GitHub's side, right?

fredrikekre (approved):

DilumAluthge:
If only we could restart a single failing job (or all failing jobs) without restarting any of the passed jobs.......

giordano:
This is getting more and more ridiculous
image

PR #51273, opened 2021-12-27
[CI] Symlink registry to current directory

giordano:
We already cloned the repository in the checkout step, avoid it cloning
over and over again.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -47,10 +47,10 @@ jobs:
             ${{ runner.os }}-test-${{ env.cache-name }}-
             ${{ runner.os }}-test-
             ${{ runner.os }}-
-      - run: julia --color=yes -e 'import Pkg; Pkg.Registry.add("General")'
-        env:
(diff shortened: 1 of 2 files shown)


DilumAluthge (changes requested):
I don't think we should do this. We don't know that the PR's copy of the registry is internally valid, or even trustworthy.

giordano:
Agreed, although it's unfortunate that we have already 99.999% of the repo locally and we need to du a full-blown clone once again.  Also, what's the rational of doing `add` + `update`?  Do you ever expect to be there any difference?  Today there are jobs getting stuck on the update step...

giordano:
Also, shouldn't the CI job for Julia v1.6 be required?  It's failing here, but this PR can be merged anyway

DilumAluthge:
> Also, shouldn't the CI job for Julia v1.6 be required? It's failing here, but this PR can be merged anyway

Yeah we should update the branch protection settings to make it required.

DilumAluthge:
https://github.com/JuliaRegistries/General/issues/51321

PR #51969, opened 2022-01-09
Upper bound some VectorizationBase-julia versions

chriselrod:
Some VectorizationBase versions are broken on Julia 1.7, creating potentially confusing bugs.

Unfortunately, version compatibility may require someone use an older VectorizationBase, e.g. if you're using Turing, you'll currently need VectorizationBase 0.16.2.

Thus, actually preventing someone from using a broken VectorizationBase may render some packages (e.g. Turing) un-installable on Julia 1.7, even though Turing isn't even using VectorizationBase directly.
So while I'd rather set 1.6 -- the latest actually working version -- as the upper bound, it seems like this could cause worse problems than what it is solving.
Hence, why this PR sets 1.7.1 for now.

This also means that once 1.7.2 is released, it'll be incompatible. That gives the package ecosystem some time to adjust, while still not breaking an otherwise working install for any current users.


chriselrod:
@DilumAluthge

PR #52844, opened 2022-01-20
Bump Missings dep on Unmarshal

DhairyaLGandhi:
It would be great to have this compat bound be verified, but it should be okay. Unamrshal is using a non-stable version of Missings and that can hold back much of the ecosystem. Now, I have made a PR https://github.com/lwabeke/Unmarshal.jl/pull/38 to do that in the parent repo, but I figured this would be faster. I saw that the registry has a version 0.4.3 whereas the parent repo only has the 04.2 tag.

### DIFF
--- a/U/Unmarshal/Compat.toml
+++ b/U/Unmarshal/Compat.toml
@@ -15,6 +15,6 @@ julia = "1"
 ["0.4-0"]
 JSON = "0.21"
 LazyJSON = "0.2"
-Missings = "0.4.3-0.4"
+Missings = "0.4.3-0.4, 1"
 Nullables = "1"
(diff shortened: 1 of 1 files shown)


DilumAluthge:
We usually prefer that the package author merges the compat bump and tags a new version of their package.

DhairyaLGandhi:
Yup, me too. The activity on the repo is low, and I noticed that there is a version that exists in the registry but not in the repo for Unmarshal already. It would be awesome if someone who has merge rights on Unmarshal to help out here.

DilumAluthge:
https://github.com/JuliaRegistries/General/pull/52893

PR #52871, opened 2022-01-20
Change package URL for NFFT.jl after move to JuliaMath

tknopp:
The repository has three packages and therefore the URL needed to be changes in all project files.


DilumAluthge (approved):

PR #52957, opened 2022-01-21
[LLVM] Limit compatibility of old versions

vchuravy:
Limit LLVM prior to v4 to Julia 1.6 since they are dependent on exports
from the runtime, that got removed in Julia 1.7


maleadt (approved):
LGTM

PR #52978, opened 2022-01-22
Fix "stub"-JET [compat]

timholy:
CC @aviatesk


aviatesk:
Thanks for the fix, @timholy .

PR #52996, opened 2022-01-22
[LLVM] Limit compatibility of -1.1 to Julia@1.2

vchuravy:
Latest Julia version where LLVM 1.1 is still passing test.


maleadt (approved):
LGTM, thanks!

PR #53824, opened 2022-02-03
Upper bound AWSS3 compat with FilePathsBase to 0.9.15

omus:
Sibling to: https://github.com/JuliaCloud/AWSS3.jl/pull/232


ericphanson:
I think we can merge this independently of https://github.com/JuliaCloud/AWSS3.jl/pull/232, because this fixes a broken incompatability that exists either way.

ericphanson (approved):

PR #54523, opened 2022-02-12
Fix ABI compatibility between Enzyme and Enzyme_jll

vchuravy:
- Enzyme@0.8 only support Enzyme_jll@0.0.25
- Fix ABI compatibility for Enzyme@0.6 and Enzyme@0.7


E/Enzyme/Compat.toml
```diff
@@ -72,16 +72,13 @@ GPUCompiler = "0.12"
...
+Enzyme_jll = "0.0.16-0.0.24"
 
 ["0.7.1-0"]
 GPUCompiler = "0.13"
 
-["0.8-0.8.4"]
-Enzyme_jll = "0.0.25-0.0"
-
-["0.8.5-0"]
+["0.8"]
```
vchuravy, line 83:
```suggestion
["0.8-0"]
```

PR #55916, opened 2022-03-03
remove broken CPUSummary versions

chriselrod:
@DhairyaLGandhi @giordano 

These versions of CPUSummary are broken and will cause errors in any package trying to use them. 0.1.9 will throw during precompilation of any dependent, while 0.1.10 will cause errors during precompilation of several dependents (including LoopVectorization.jl).

### DIFF
--- a/C/CPUSummary/Versions.toml
+++ b/C/CPUSummary/Versions.toml
@@ -27,9 +27,11 @@ git-tree-sha1 = "849799453de85b55e78550fc7b0c8f442eb497ab"
 
 ["0.1.9"]
 git-tree-sha1 = "282844544d64a8e8c860f206385ebd736a0bc60d"
+yanked = true
 
 ["0.1.10"]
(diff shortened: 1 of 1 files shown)


C/CPUSummary/Versions.toml
```diff
@@ -25,11 +25,5 @@ git-tree-sha1 = "ba19d1c8ff6b9c680015033c66802dd817a9cf39"
 ["0.1.8"]
 git-tree-sha1 = "849799453de85b55e78550fc7b0c8f442eb497ab"
 
-["0.1.9"]
-git-tree-sha1 = "282844544d64a8e8c860f206385ebd736a0bc60d"
```
fredrikekre, line 29: Just put `yanked=true` here instead.

chriselrod: Thanks, didn't know how.

PR #56805, opened 2022-03-18
Update Compat for JuliaInterpreter for compatibility with Julia 1.8+

George9000:
JuliaInterpreter versions before 0.9.11 cause segfaults on Julia 1.8. See [discussion](https://github.com/timholy/Revise.jl/issues/675#issuecomment-1071513603). An attempt at updating compat to reflect this requirement. In collaboration with @giordano, trying to clean up prior entries as well to reflect [compat documentation](https://github.com/JuliaLang/Pkg.jl/issues/1190).


George9000:
JuliaInterpreter versions released:

```
0.1.1
0.2.0
0.2.1
0.3.0
0.3.1
0.3.2
0.4.0
0.4.1
0.5.0
0.5.1
0.5.2
0.6.0
0.6.1
0.7.0
0.7.1-0.7.26
0.8.0-0.8.21
0.9.0-0.9.11
```
and CodeTracking versions:

```
0.1.0
0.2.0
0.2.1
0.3.0
0.3.1-0.3.4
0.4.0
0.5.0
0.5.1-0.5.12
1.0.0-1.0.8
(note 1.0.0 was released before 0.5.12)
```

J/JuliaInterpreter/Compat.toml
```diff
@@ -1,20 +1,39 @@
-["0-0.7"]
-julia = "1"
+["0 - 0.7"]
+julia = "1.0"
```
giordano, line 2: We want to prevent all old versions to install on Julia v1.8.  So this should be
```suggestion
julia = "1.0 - 1.7"
```

J/JuliaInterpreter/Compat.toml
```diff
@@ -1,20 +1,39 @@
...
+# assuming exact match was wanted
+# otherwise, use ~ or ^ specifier
```
giordano, line 4: For the same reasons, remove all comments.  The script that automatically generates this file doesn't know anything about comments.
```suggestion
```

J/JuliaInterpreter/Compat.toml
```diff
@@ -1,20 +1,39 @@
...
 
-["0.2-0.7.13"]
-CodeTracking = "0.3.2-0.5"
+["0.2 - 0.7.13"]
+CodeTracking = "0.3.2 - 0.5"
+# note this allows
+# CodeTracking [0.3.2 - 0.6)
+# otherwise use
+# CodeTracking = "0.3.2 - 0.5.8"
```
giordano, line 11:
```suggestion
["0.2-0.7.13"]
CodeTracking = "0.3.2-0.5"
```

J/JuliaInterpreter/Compat.toml
```diff
@@ -1,20 +1,39 @@
...
+CodeTracking = "0.5.9 - 0.5-1"
+# unclear what is meant here
+# or if this is a mistake
+# CodeTracking 0.5.9 to 1 or to 0.5.10?
+# CodeTracking latest version is now at 1.0.8
 
-["0.8-0.8.4"]
-julia = ["1.0", "1.5.0-1"]
+["0.8 - 0.8.4"]
+julia = "1.0 - 1.5.0"
```
giordano, line 24: This is _**not**_ the same as before.  Plus, we want to prevent it from being installable on v1.8
```suggestion
 julia = ["1.0", "1.5.0-1.7"]
```

J/JuliaInterpreter/Compat.toml
```diff
@@ -1,20 +1,39 @@
...
 
-["0.8.5-0.8"]
-julia = "1"
+["0.8.5 - 0.8"]
+julia = "1.5"
+# unclear what is meant
+# but thinking this is what is meant
+# compat says this is the interval
+# julia [1.5 1.6)
```
giordano, line 31: Same comments as above
```suggestion
["0.8.5-0.8"]
julia = "1.0 - 1.7"
```

J/JuliaInterpreter/Compat.toml
```diff
@@ -1,20 +1,39 @@
...
 
-["0.9-0"]
-julia = "1.6.0-1"
+["0.9 - 0.9.10"]
+julia = "1.6 - 1.7"
+# specifies julia up to 1.8
+# 1.8 requires Interpreter 0.9.11
```
giordano, line 36:
```suggestion
["0.9-0.9.10"]
julia = "1.6.0 - 1.7"
```

J/JuliaInterpreter/Compat.toml
```diff
@@ -1,20 +1,39 @@
...
+CodeTracking = "0.3.2 - 0.5"
+# note this allows
+# CodeTracking [0.3.2 - 0.6)
+# otherwise use
+# CodeTracking = "0.3.2 - 0.5.8"
 
-["0.7.14-0.7.20"]
-CodeTracking = "0.5.9-0.5"
+["0.7.14 - 0.7.20"]
+CodeTracking = "0.5.9 - 0.5"
```
giordano, line 14:
```suggestion
["0.7.14-0.7.20"]
CodeTracking = "0.5.9-0.5"
```

J/JuliaInterpreter/Compat.toml
```diff
@@ -1,20 +1,39 @@
...
 
-["0.7.21-0"]
-CodeTracking = ["0.5.9-0.5", "1"]
+["0.7.21 - 0"]
+CodeTracking = "0.5.9 - 0.5-1"
+# unclear what is meant here
+# or if this is a mistake
+# CodeTracking 0.5.9 to 1 or to 0.5.10?
+# CodeTracking latest version is now at 1.0.8
```
giordano, line 21:
```suggestion
["0.7.21-0"]
CodeTracking = ["0.5.9-0.5", "1"]
```

J/JuliaInterpreter/Compat.toml
George9000: Since 0.5 allows versions up to but not including 0.6, doesn't this allow Interp 0.7.13 to be used with CodeTracking 0.5.9?

Line 13 implies  CodeT 0.5.9 requires Interp 0.7.14

J/JuliaInterpreter/Compat.toml
George9000: I must be missing something. If ["0.5.9-0.5, "1"] means a union of the versions specified, why not just "0.5.9-1"?

giordano: Because `["0.5.9-0.5, "1"]` doesn't include v0.6 or the other infinite many breaking releases between 0.6 and 1.0`. `"0.5.9-1"` does it instead.

giordano:
Alright, it _should_ be good now, at least CI is happy!  Thanks a lot for doing this!

PR #57155, opened 2022-03-23
SmartTensors/PhysicsInformedML.jl.git rename

montyvesselinov:


montyvesselinov:
@JuliaRegistrator register

montyvesselinov:
@DilumAluthge is this going to be merged?! Am i doing the right thing?!

giordano:
> Am i doing the right thing?!

It'd be better to follow the documentation, which suggests to use GitHub transfer option:
https://github.com/JuliaRegistries/General/blob/7c9afaff058161907662f02123f9593b7b65fbc1/README.md#how-do-i-transfer-a-package-to-an-organization-or-another-user  You can delete https://github.com/SmartTensors/PhysicsInformedML.jl and then move https://github.com/TensorDecompositions/PhysicsInformedML.jl to the new URL, instead of creating duplicates.

montyvesselinov:
@giordano i do not plan to delete https://github.com/TensorDecompositions/PhysicsInformedML.jl 

i prefer to keep both https://github.com/TensorDecompositions/PhysicsInformedML.jl  and https://github.com/SmartTensors/PhysicsInformedML.jl because both have been referenced in papers and websites.

can you merge this?!

PR #57386, opened 2022-03-26
Revert "New version: OutlierDetection v0.2.6"

davnn:
Somehow TagBot cannot deal with this release, possibly because the last commit contains `[no ci]`? Reverts JuliaRegistries/General#57299

Would like to register again with an ammended commit message.


fredrikekre:
You can just `git tag` this release manually then.

davnn:
I don‘t think the git tree sha is valid then because I ammended the last commit message? Would prefer a merge of this revert + register @fredrikekre, or does anything speak against this?

davnn:
> You can just `git tag` this release manually then.

I did manage to fix the problem by undoing the amend and re-running tagbot manually afterwards.

PR #58407, opened 2022-04-12
change location of NMFk.jl to SmartTensors/NMFk.jl

montyvesselinov:


giordano:
I believe I said this already, but the canonical way to do this is to use the GitHub transfer functionality, not to create a brand new repository: https://github.com/JuliaRegistries/General/blob/6149b8c74e6122bb195408bbda312b749675c013/README.md#how-do-i-transfer-a-package-to-an-organization-or-another-user

PR #58501, opened 2022-04-14
Fix package information

thechampagne:


D/DogAPI/Versions.toml
```diff
@@ -1,2 +1,2 @@
 ["1.0.0"]
-git-tree-sha1 = "4ce09a6f533629106581683627ba990df00f9731"
+git-tree-sha1 = "7cc90eb9872302897550442f3ea0e726e901c8c8"
```
fredrikekre, line 2: This is not allowed.

fredrikekre (changes requested):

D/DogAPI/Versions.toml
thechampagne: There is no such a repo "https://github.com/thexxiv/DogAPI.jl.git"

thechampagne: It was my old account, new repo have new git hash

thechampagne: So If you didn't change the hash It won't be a problem to download the package ?

thechampagne:
.

giordano:
> .

That won't change the fact we can't accept this PR

thechampagne:
> 
> That won't change the fact we can't accept this PR

is there a solution?

giordano:
> is there a solution?

If the old git history isn't available anywhere, not really.

PR #59395, opened 2022-04-29
Yank Arpack_jll v3.8.0

IanButterworth:
According to @giordano and https://github.com/JuliaLinearAlgebra/Arpack.jl/issues/138 and https://github.com/JuliaLinearAlgebra/Arpack.jl/commit/d014b07135ce907c0d9ad536074a66af0eea6cb1 this release appears to be broken.

@ViralBShah is this right?

Without this being yanked I'm getting v3.8.0 resolved, and Arpack.jl held down at v0.5.0.
And seeing issues on aarch64 machines

### DIFF
--- a/A/Arpack_jll/Versions.toml
+++ b/A/Arpack_jll/Versions.toml
@@ -22,3 +22,4 @@ git-tree-sha1 = "cc34a702b843b26db6f8a5e7579762448829548d"
 
 ["3.8.0+0"]
 git-tree-sha1 = "4c31b0101997beb213a9e6c39116b052e73ca38c"
+yanked = true


ViralBShah:
Yes we can yank this.

ViralBShah:
I noted that Arpack_jll 3.5 also suffers from the same issue, and that this may not have to do with Arpack at all. https://github.com/JuliaLinearAlgebra/Arpack.jl/issues/138#issuecomment-1286760090

We should un-yank this - if that is a term.

ViralBShah:
We should revert this PR.

PR #59593, opened 2022-05-03
Yank arpack_jll 3.7

ViralBShah:
This build has bugs with Arpack.jl and seems to cause other issues.


giordano:
For the record, v3.7 of Arpack_jll is a very old build and https://github.com/JuliaPackaging/Yggdrasil/pull/254 suggests that Arpack.jl had problems with v3.7 of the library. I can't see any other packages depending on Arpack_jll, so I think it's safe to yank this build from the registry.

PR #60303, opened 2022-05-15
Update repo link for BinningAnalysis

ffreyer:
The current link is incorrect because @carstenbauer changed his name


carstenbauer:
Thanks Frederic! Would have expected it to work anyways (there is a redirect: https://github.com/crstnbr/BinningAnalysis.jl still works) but it's probably an issue / limitation on the JuliaRegistrator side of things.

ffreyer:
Yea the bot complained about it:
> Error while trying to register: Changing package repo URL not allowed, please submit a pull request with the URL change to the target registry and retry.

PR #60675, opened 2022-05-20
Yank Legolas 0.3.4

omus:
A breaking change was accidentally added in https://github.com/beacon-biosignals/Legolas.jl/pull/40 and left on the default branch. When another change was introduced the 0.3.4 was created accidentally releasing the breaking change to a non-breaking release.


omus:
Registering the change as breaking: https://github.com/JuliaRegistries/General/pull/60677

PR #60739, opened 2022-05-21
Add in sent to twitter action for JuliaPackages bot account

logankilpatrick:
This code should send the package updates to Twitter now as well: https://twitter.com/JuliaPackages

### DIFF
--- a/.github/workflows/feed.yml
+++ b/.github/workflows/feed.yml
@@ -90,7 +90,7 @@ jobs:
       - uses: julia-actions/setup-julia@latest
         with:
           version: 1.6.2
-      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
       - name: Send to Twitter
         shell: julia --color=yes {0}
         run: |
--- a/.github/workflows/feed.yml
+++ b/.github/workflows/feed.yml
@@ -103,7 +103,7 @@ jobs:
           using Twitter
           twitterauth(ENV["API_Key"], ENV["API_Key_Secret"], ENV["Access_Token"], ENV["Access_Token_Secret"])
           
-          myStatus = string("New #JuliaLang package announced: ", ENV["PR_NAME"],  "\n $text \n Learn more: $pr_url")
+          myStatus = string("New #JuliaLang package announced: ", ENV["PR_NAME"],  "\n$text \nLearn more: $pr_url")
           
           response = post_status_update(status=myStatus)
           
--- a/.github/workflows/feed.yml
+++ b/.github/workflows/feed.yml
@@ -82,3 +82,39 @@ jobs:
         type: 'stream'
         topic: 'New package registrations'
         content: ${{ steps.zulip.outputs.content }}
+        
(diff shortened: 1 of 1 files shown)


logankilpatrick:
I have already added the secrets to this repo

logankilpatrick:
I tested the API keys and such locally: https://twitter.com/JuliaPackages/status/1528077265895948288?s=20&t=_4vEvexzE-5F40PfQglOUQ

The general logic is the same as: https://github.com/JuliaLangSlack/StackOverflowBot/blob/main/src/TwitterBot.jl

.github/workflows/feed.yml
```diff
@@ -82,3 +82,39 @@ jobs:
         type: 'stream'
         topic: 'New package registrations'
         content: ${{ steps.zulip.outputs.content }}
+        
+  update-twiter:
+      if: ${{ github.event.label.name == 'new package' && github.repository == github.event.pull_request.head.repo.full_name }}
+      runs-on: ubuntu-latest
+      steps:
+      - uses: julia-actions/setup-julia@latest
```
giordano, line 90: It's usually more secure to specify a version, rather than using `latest`

.github/workflows/feed.yml
```diff
@@ -82,3 +82,39 @@ jobs:
...
+        with:
+          version: 1.6.2
```
giordano, line 92: `1.6.2` is a bit odd, it isn't even the latest LTS (that'd be `1.6.7`).  But presumable `1.7.2` (latest stable version` would also work?

.github/workflows/feed.yml
```diff
@@ -82,3 +82,39 @@ jobs:
...
+      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
```
giordano, line 93: Latest stable version is 3.0.2:
```suggestion
      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
```

.github/workflows/feed.yml
logankilpatrick: Works for me!

.github/workflows/feed.yml
logankilpatrick: This is just copied from the other parts of the action, should we update it for all of them ?

DilumAluthge:
I'll let @giordano and @ericphanson take the lead on reviewing this.

.github/workflows/feed.yml
giordano: Yes, it'd probably be better.  I think we can do
```suggestion
      - uses: julia-actions/setup-julia@ce353a9bc8bd2a581d0533eb1d0ffc6f60563eb0 # v1.7.1
```

logankilpatrick:
@giordano Okay, updated per the suggestions, I will make a separate PR to update the version for the other jobs so that we can easily revert if something goes wrong.

ericphanson (approved):
Looks fine to me!

giordano (approved):
Can't comment about the code inside the step (I'm not familiar with the twitter package), but the setup looks good to me me, as far as I can tell

> I will make a separate PR to update the version for the other jobs so that we can easily revert if something goes wrong.

Sounds good 👍

PR #60880, opened 2022-05-23
Update send to twitter with new formatting

logankilpatrick:
It looks like the parsed text already has the URL in it: https://twitter.com/JuliaPackages/status/1528796254783696896?s=20&t=PU0OPO_0hdi13CKdT7QzSg so this update condenses what we tweet to remove redundant info.

### DIFF
--- a/.github/workflows/feed.yml
+++ b/.github/workflows/feed.yml
@@ -103,7 +103,7 @@ jobs:
           using Twitter
           twitterauth(ENV["API_Key"], ENV["API_Key_Secret"], ENV["Access_Token"], ENV["Access_Token_Secret"])
           
-          myStatus = string(ENV["PR_NAME"],  " announced #JuliaLang\n$text")
+          myStatus = string(ENV["PR_NAME"],  " is being registered #JuliaLang\n$text")
           @info "Tweet to be sent: $myStatus"
           @info "Tweet length: $(length(myStatus))"
 
--- a/.github/workflows/feed.yml
+++ b/.github/workflows/feed.yml
@@ -103,7 +103,7 @@ jobs:
           using Twitter
           twitterauth(ENV["API_Key"], ENV["API_Key_Secret"], ENV["Access_Token"], ENV["Access_Token_Secret"])
           
-          myStatus = string("New #JuliaLang package announced: ", ENV["PR_NAME"],  "\n$text \nLearn more: $pr_url")
+          myStatus = string(ENV["PR_NAME"],  " announced #JuliaLang\n$text")
           @info "Tweet to be sent: $myStatus"
           @info "Tweet length: $(length(myStatus))"
 


.github/workflows/feed.yml
```diff
@@ -103,7 +103,7 @@ jobs:
           using Twitter
           twitterauth(ENV["API_Key"], ENV["API_Key_Secret"], ENV["Access_Token"], ENV["Access_Token_Secret"])
           
-          myStatus = string("New #JuliaLang package announced: ", ENV["PR_NAME"],  "\n$text \nLearn more: $pr_url")
+          myStatus = string(ENV["PR_NAME"],  " announced #JuliaLang\n$text")
```
giordano, line 106: So what will be the new test now?

logankilpatrick: This should fix the issue with the tweet being too long as well

giordano: I saw the tweet,  but there is a lot of text, can't easily guess what is what :slightly_smiling_face:

logankilpatrick: "New package: MultiAgentPathFinding v0.1.0" is `ENV["PR_NAME"]` and `text` is everything else. The other parts of the tweet are the text I removed in this PR

ericphanson: On Slack we pass the "title" as `ENV["PR_NAME"]` and the body of the message as `text`. It looks like this:
Screen Shot 2022-05-24 at 01 26 37
FWIW

But the "announced" bit is weird- this is for prospective regisrations, NOT new packages. We send these to Slack/zulip at registration time exactly so the community can comment. I think an "announcement" should wait until the package is registered.

ericphanson: So actually I think putting a twitter feed here is a bad idea. We should just add another GHA triggered on merges or something, to do this once the package is actually registered

logankilpatrick: I will switch to registered

logankilpatrick: I am not opposed to getting more eyes on these packages before they are merged

DilumAluthge: @ericphanson Would it be as simple as changing the trigger for the Twitter feed to be "merged" instead of "created"?

logankilpatrick: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges @DilumAluthge yes we could, I don't feel strongly about posting before or after the merge, so I'll defer to @ericphanson

ericphanson: I made https://github.com/JuliaRegistries/General/pull/60897 if we want to go the "announce" route

ericphanson: I would be careful about auto-tweeting unmoderated user-generated package names and in anyway implying they are endorsed somehow... If they are registered then they at least have gone through a 3 day waiting period and probably do not contain racial slurs in the package name. If not... who knows what people will do esp. if it will bring publicity through twitter.

logankilpatrick: Agreed, let's go with #https://github.com/JuliaRegistries/General/pull/60897

logankilpatrick: Approved there, thanks for the feedback!

PR #60897, opened 2022-05-23
Move Twitter workflow to upon PR merge instead of PR open

ericphanson:
using https://stackoverflow.com/a/59588725/12486544 and https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges

I'm not 100% sure this will work correctly... hard to test these things.


logankilpatrick (approved):

ericphanson:
Let's give it a try

giordano (approved):
Looks good as far as I can tell 👍

ericphanson:
Sorry @logankilpatrick looks like I didn't copy over the formatting change from https://github.com/JuliaRegistries/General/pull/60880 - want to a make a PR against the new workflow?

logankilpatrick:
I update it in #4818af2

ericphanson:
Ah just saw that, alright! Now we wait for a package to be merged :)

logankilpatrick:
Yes, looks like #60696 is close

PR #60945, opened 2022-05-24
Yank Atom.jl@0.12.37

pfitzseb:
This version had incorrect compat bounds specified for a dependency, so it usually doesn't work.

-> https://github.com/JuliaRegistries/General/pull/60944


giordano:
There is a conflict

PR #61256, opened 2022-05-29
Yank BioSequences v3.0.1

TransGirlCodes:


giordano (changes requested):
No, you can ***not*** remove an already registered version: https://github.com/JuliaRegistries/General/blob/8ae8b1a96eb1df44b71b6dc911ed025679e36ac7/README.md#how-do-i-remove-a-package-or-version-from-the-registry. If you want to make it impossible for new users to install a version, you have to yank it, which means adding the line
```toml
yanked = true
```
after it. See for example #61264

PR #61385, opened 2022-05-31
upper compat bound for NNlib

CarloLucibello:
hotfix for https://github.com/FluxML/NNlib.jl/issues/412

See also discussion in https://github.com/FluxML/NNlibCUDA.jl/pull/50


CarloLucibello:
@giordano sorry for bothering but it would be good to merge this as soon as possible

fredrikekre:
Wouldn't it be better to yank the bad version?

PR #61893, opened 2022-06-07
new versions of my packages

aplavin:
... so that not to have multiple Registrator PRs with manual merge due to non-consecutive versions. Nothing changes for already registered versions.


giordano:
This fell through the cracks.  The fact is that these manual PRs aren't tested automatically (it isn't straightforward to detect what changed exactly in arbitrary cases) and dumping some hundreds of new versions would require extra work for maintainers to manual review all changes that no one is willing to undertake.

giordano:
I tried to locally install some of the newly added versions and they all worked, but there is no way I'm going to test all of them, so I'm going to hope all others are fine as well.

As a general remark, it's preferable to get a PR which skip version numbers, at least it's actually tested.  Asking for a manual merge  in `#pkg-registration` on Slack is ok in those cases.

PR #62986, opened 2022-06-24
[Plasmo] retroactively fix compat bounds

odow:
This package uses internal functionality of JuMP that will be removed in v1.2

Same as https://github.com/JuliaRegistries/General/pull/62984
x-ref https://github.com/plasmo-dev/Plasmo.jl/pull/64
x-ref https://github.com/jump-dev/JuMP.jl/pull/2955

cc @jalving


jalving:
looks good to me, thanks for taking care of this!

PR #64139, opened 2022-07-13
[BitemporalReactive] Correcting repo url

michaelfliegner:


giordano (changes requested):
The repositories are different and have different histories, how is this supposed to work?

michaelfliegner:
> The repositories are different and have different histories, how is this supposed to work?

The transfer to ASA commit had the wrong repo name
224216acd053a0a78ba10f1b3b4d9c6afcf4c8c6 

I want to correct that.

PR #64542, opened 2022-07-19
New version: LAPACK_jll v3.10.0+0

jlbuild:

* Registering JLL package LAPACK_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/LAPACK_jll.jl
* Version: v3.10.0+0
* Commit: b8276591031f86c07cf3ebc1379afd1e216b9efc
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/943878f01bc5fbb9775b4a5b679b5f0b84940508
* Created by: @ViralBShah


giordano:
Because the libblastrampoline you used to link lapack has an incompatible ABI than the one used in Julia v1.7, which is where CI is run here, which means that the package is effectively unusable in v1.7, despite what's written in the compat.  I'd recommend rebuilding lapack_jll, but specify julia v1.8 as compat, and also add `; compat="5.1.1"` to libblastrampoline `Dependency`

PR #64567, opened 2022-07-19
LLD_jll: Add compat branch for v12.0.0

jpsamaroo:
@vchuravy


L/LLD_jll/Compat.toml
```diff
@@ -1,3 +1,7 @@
+[12]
+LLVM_jll = "12.0.0"
```
vchuravy, line 2: I think you also want `12.0.0-13.0.1`?

jpsamaroo:
This appears to work correctly now.

L/LLD_jll/Compat.toml
```diff
@@ -1,3 +1,7 @@
+[12]
+LLVM_jll = "12.0.0-13.0.1"
+julia = "1.7"
```
vchuravy, line 3: LLVM 12 is Julia 1.6 so we can lower the bound?

PR #65311, opened 2022-07-29
Yank bad UserNSSandbox_jll version

staticfloat:
This version introduced a bug for some important users, let's yank it.

X-ref: https://github.com/staticfloat/Sandbox.jl/pull/97


DilumAluthge:
@staticfloat Do we still need to merge this yank?

staticfloat:
It's not critical, but it _is_ a bad version.

DilumAluthge (approved):

PR #65552, opened 2022-08-03
Yank latest Metatheory (1.4.0)

ChrisRackauckas:
@0x0f0f0f

PR #65562, opened 2022-08-03
Correctly yank Metatheory 1.4.0

DilumAluthge:
Follow-up to https://github.com/JuliaRegistries/General/pull/65552


DilumAluthge:
Because we don't (yetH have a test that covers this case.

https://github.com/JuliaRegistries/RegistryCI.jl/issues/454

PR #65921, opened 2022-08-09
Move my work stuff to a newly created organization

shamazmazum:
Hi! Can you please move links to my repos to a new organization?


DilumAluthge (approved):
1. I tested each of the URLs and made sure that each old URL redirects to the corresponding new URL.
2. The author of this pull request is also the user that was the repository owner when each of the repos was at its old URL.

PR #65995, opened 2022-08-10
remove Dynamical Billiards version 3.5.2 since it is impossible to install due to empty compat range

KristofferC:
The version spec `0.5.2 - 0.2` for MuladdMacro for DynamicalBilliards at version 3.5.2 is empty. Installing it is impossible. This fact also makes it a bit weird for tools that rewrite the registry so it is best to remove it.
```
julia> Pkg.Versions.VersionSpec("0.5.2 - 0.2")
VersionSpec("∅")

(jl_QrFp3d) pkg> add DynamicalBilliards@3.5.2
    Updating git-repo `https://github.com/JuliaRegistries/General`
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package MuladdMacro [46d2c3a1]:
 MuladdMacro [46d2c3a1] log:
 ├─possible versions are: 0.1.0-0.2.2 or uninstalled
 └─found to have no compatible versions left with DynamicalBilliards [4986ee89]
   └─DynamicalBilliards [4986ee89] log:
     ├─possible versions are: 2.0.0-4.0.0 or uninstalled
     └─restricted to versions 3.5.2 by an explicit requirement, leaving only versions 3.5.2
```

This patch was created by deleting the version from `Versions.toml` and then reloading + resaving the registry data for DynamicalBilliards


DilumAluthge:
For context, this is the `Project.toml` for DynamicBilliards 3.5.2:

```toml
name = "DynamicalBilliards"
uuid = "4986ee89-4ee5-5cef-b6b8-e49ba721d7a5"
repo = "https://github.com/JuliaDynamics/DynamicalBilliards.jl.git"
version = "3.5.2"

[deps]
Elliptic = "b305315f-e792-5b7a-8f41-49f472929428"
MuladdMacro = "46d2c3a1-f734-5fdb-9937-b9b9aeba4221"
Requires = "ae029012-a4dd-5104-9daa-d747884805df"
StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"

[compat]
julia = "≥ 1.0.0"
MuladdMacro = "≥ 0.5.2"
StaticArrays = "≥ 0.8.2"
Elliptic = "≥ 0.5.0"
```

Here is what I suspect happened:

1. DynamicBilliards 3.5.2 was registered with a `Project.toml` file that contains the compat entry `MuladdMacro = "≥ 0.5.2"`. In the registry, this is stored as `MuladdMacro = "0.5.2 - *"`.
2. Sometime later, we did a registry-wide RetroCapping. At the time that we did the RetroCapping, the latest registered release of MuladdMacro was 0.2.x. [^1]. Therefore, in the registry, RetroCap changes `MuladdMacro = "0.5.2 - *"` to `MuladdMacro = "0.5.2 - 0.2"`.

I think we have two options here:
1. Completely remove DynamicBilliards 3.5.2 from the registry.
2. Make both of the following changes:
    - In the registry, change the compat entry (for DynamicBilliards 3.5.2 only) from `MuladdMacro = "0.5.2 - 0.2"` to `MuladdMacro = "0.2 - 0.2"`, which is likely a more accurate representation of what the package authors intended.
    - Yank DynamicBilliards 3.5.2

[^1: I know that this is true, because today, the latest registered release of MuladdMacro is still 0.2.x: https://github.com/JuliaRegistries/General/blob/master/M/MuladdMacro/Versions.toml

KristofferC:
Since it is literally uninstallable I think it is fine to delete it like is done here. It won't break `instantiate`. Changing to `0.2-0.2` also works.

The yank option is not great because after normalization (https://github.com/JuliaRegistries/General/pull/65997) this will turn into an empty range which is printed as `[]` which I am not sure even parses back as a VersionRange.

PR #66039, opened 2022-08-10
Add DelimitedFiles package and compat for it to packages

KristofferC:
The plan is to move out DelimitedFiles from being an stdlib to being an external package. However, there is no compat on standard libraries. When people start using DelimitedFiles as a normal package, the lack of compat will be problematic.

As described in https://github.com/JuliaLang/julia/pull/45540#issuecomment-1154053683, this mirrors the compat for Julia to that of DelimitedFiles which should make sense for existing versions, new versions of packages pre 1.9 should have compat on DelimitedFiles that mirrors their Julia compat, packages only supporting 1.9+ are free to have any compat on DelimitedFiles.


KristofferC:
The PR here should be safe in the sense that the info here should be completely ignored on pre 1.9 Julias and it is also easy to revert since you would just remove all the lines with DelimitedFiles in Compat files.

DilumAluthge (approved):
To be safe, it would be nice to do a PkgEval run using this branch of the registry, just to make sure that we aren't missing anything.

KristofferC:
> To be safe, it would be nice to do a PkgEval run using this branch of the registry, just to make sure that we aren't missing anything.

https://github.com/JuliaLang/julia/pull/45540#issuecomment-1210612971 ran on a branch that included this PR and didn't show any problems.

PR #66044, opened 2022-08-10
Restore DynamicalBilliards 3.5.2, yank it, and modify one of its compat entries

DilumAluthge:
I think it would be nice if we could avoid deleting versions from the registry whenever possible. However, I agree with Kristoffer that we shouldn't have any compat entries in the registry that point to an empty range or invalid range. Therefore, this PR does the following:

1. Restore DynamicalBilliards 3.5.2 (which was deleted in https://github.com/JuliaRegistries/General/pull/65995)
2. For DynamicalBilliards 3.5.2 only, change the MuladdMacro compat entry from `MuladdMacro = "0.5.2 - 0.2"` to `MuladdMacro = "0.2"` (which is equivalent to `MuladdMacro = "0.2 - 0.2"`)
3. Yank DynamicalBilliards 3.5.2


KristofferC:
This has literally been uninstallable for many years so I don't see what difference this will do in practice.

DilumAluthge:
There is no difference in practice.

However, we try really hard to avoid ever deleting versions from the General registry. I think that's a good policy, and it would be nice if we could adhere to it as much as possible.

In this case, by merging this PR, we are able to fix the underlying issue (empty compat ranges) without deleting versions from the registry.

PR #66127, opened 2022-08-12
Post a feed of manual PRs to the `#pkg-registration-manual-prs` channel on Slack

DilumAluthge:


DilumAluthge:
@logankilpatrick @ericphanson Currently, it looks like these messages are getting posted to the `#new-packages-feed` channel.

What do I need to modify in order to get the messages to post to the `#pkg-registration-manual-prs` channel?

.github/workflows/feed_manual_prs.yml
```diff
@@ -0,0 +1,75 @@
...
+        data = Dict("attachments" => [attachment])
+
+        if !startswith(endpoint, "/")
+            endpoint = string("/", endpoint)
+        end
+        @info "Told Slack: $(data)"
+        response = sendattachmenttoslack(data, endpoint)
+        @info "Slack said: $response"
+      env:
+        SLACK_ENDPOINT: ${{ secrets.JULIALANGSLACKENDPOINT }}
```
ericphanson, line 72: These are slack web hooks (https://api.slack.com/messaging/webhooks) so they are tied to a single channel. So we need a Slack admin to make a new web hook for the new channel, and then to add it as a secret here. That should be it!

PR #66501, opened 2022-08-18
Changed organization for PlutoTeachingTools.jl

eford:


DilumAluthge (approved):
I have confirmed that the old URL redirects to the new URL.

PR #66582, opened 2022-08-19
Update InfiniteOpt link with Organization Change

pulsipher:
This updates the InfiniteOpt source link in accordance with it joining the InfiniteOpt Organization.


DilumAluthge (approved):
I confirmed that the old URL redirects to the new URL.

PR #65834, opened 2022-08-08
New version: ROCmDeviceLibs_jll v4.5.2+0

jlbuild:
Autogenerated JLL package registration

* Registering JLL package ROCmDeviceLibs_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/ROCmDeviceLibs_jll.jl
* Version: v4.5.2+0
* Commit: 2c95d958e65218a6e58f825847c67f4dc88f57eb
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/26b8d6bad3b5a0602297d95531fce25a5fe5f05d
* Created by: @pxl-th

PR #65644, opened 2022-08-04
New version: ROCmLLVM_jll v4.5.2+0

jlbuild:
Autogenerated JLL package registration

* Registering JLL package ROCmLLVM_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/ROCmLLVM_jll.jl
* Version: v4.5.2+0
* Commit: 86b04aed55151b8f3a75ea65238c951b101465cf
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/c14d170d7c993f48294b95e4e45f6f38cf5d59a4
* Created by: @pxl-th

PR #66835, opened 2022-08-23
Add rocm_cmake 4.5.2

pxl-th:
It was [added](https://github.com/JuliaBinaryWrappers/rocm_cmake_jll.jl), but not registered for some reason.


giordano:
Some reason being #66292. You've even been tagged there

pxl-th:
> Some reason being #66292. You've even been tagged there

That issue happened because CI used Julia 1.7, while for rocm_cmake of this version (4.5.2) we need Julia 1.8 (there is an explicit requirement).
The reason is that ROCm 4.5 uses LLVM 13 and we need a matching Julia's LLVM, which is in 1.8.

I've had similar issues with other ROCm packages: [ROCmDeviceLibs](https://github.com/JuliaRegistries/General/pull/65834) & [ROCmLLVM](https://github.com/JuliaRegistries/General/pull/65644).

PR #67264, opened 2022-08-29
Update URL for GeoSimulation.jl

juliohm:
The package has been deprecated and moved out of the organization.


fredrikekre:
Why not use the GitHub archive functionality and leave them in the org?

juliohm:
I will archive it as well after the URL is updated. The plan is to make the org clean with only actively maintained modules.

Issue #62795, opened 2022-06-21
Stop yanking packages


ranocha:
I would really like the General registry to stop yanking releases of ordinary Julia packages. For example, https://github.com/JuliaRegistries/General/pull/62675 re-introduced a bug in a package with Julia v1.8 we rely on in our packages (details of the packages do not necessarily matter here right now). Since this bug was fixed by the maintainers of our dependency a few days ago, I was quite puzzled. In the end, it cost me quite some valuable time to debug this and find out that the release that worked for us was yanked from the general registry.

In my opinion, yanking packages is completely against all rules of Julia for reproducible science we advocate. Making it impossible to install a version of a package I used for my research project yesterday in the future is just a no go for me. The proper fix for broken releases of packages is to release a new patch version of this broken package, either just reverting to the old working version or including a proper bug fix.

This may sound a bit strong, but please forgive me. I would just like to start a discussion and strongly encourage people to stop yanking releases.

fredrikekre:
> Making it impossible to install a version of a package I used for my research project yesterday in the future is just a no go for me.

You can still install yanked packages if you have them in the manifest (which you need to reproduce anyway).

ranocha:
Okay, so reproducibility is still fine (I just didn't know how exactly yanked package releases are treated). But what about the other aspect - why yank a package release instead of releasing a new version? In our case, I updated the compat bounds of our package to restrict our dependency to a working version (with the upcoming release of Julia v1.8). Since the dependency release is yanked, this makes this version of our package with updated compat bounds impossible to install. From my point of view, this is highly irritating behavior. Shouldn't we encourage people to just make a new patch release when they discover they introduced a bug in an earlier release?

fredrikekre:
I guess you have to ask the package maintainer about that. But sometimes yanking is the quickfix since you might have time to fix the broken release right away. In the vast majority of cases I think people don't yank and just release a patch version instead. (Otherwise, all package versions that is not the latest patch version should be yanked.) Note also that there is nothing stopping a yanked release from being un-yanked after a new patch version is released.

fredrikekre:
I suppose also that it is slightly unusual (?) to have compat on a patch release, but since all Julia packages seem to happily use [0Ver](https://0ver.org/) instead of utilizing all 3 available numbers you kind of have no choice...

sloede:
[...]

PR #67468, opened 2022-09-01
yank MLJDecisionTreeInterface 0.2.4

ablaom:
https://github.com/JuliaAI/MLJDecisionTreeInterface.jl/pull/31


ericphanson:
Some folks would prefer package authors push a revert commit and then tag a new package release, instead of yanking: https://github.com/JuliaRegistries/General/issues/62795. That has the advantage that one doesn't need to refer to the registry (only the package's Project.toml and git history) to understand the current state of the releases, and it doesn't require manual intervention by the registry maintainers.

But if you still want to yank it, we can do that.

DilumAluthge (approved):

PR #67573, opened 2022-09-02
Fix compat for hsa_rocr_jll

pxl-th:
This PR fixes compat to unblock https://github.com/JuliaRegistries/General/pull/67535


giordano:
This is touching lots of files

pxl-th:
> This is touching lots of files

Should be good now.

H/hsa_rocr_jll/Compat.toml
```diff
@@ -7,9 +7,10 @@ JLLWrappers = "1.2.0-1"
...
 
-["4.2-4"]
+["4.2"]
 julia = "1.7.0-1"
+Elfutils_jll = "0.182"
 
 ["4.5-4"]
-Elfutils_jll = "0.182"
 julia = "1.8.0-1"
+Elfutils_jll = "0.182"
```
giordano, line 16: Am I correct to say this is unnecessary? I'm struggling to see the difference.  Also, isn't this is the same version of `Elfutils_jll` as above?

pxl-th: Yes, reverted.

PR #67855, opened 2022-09-07
LLD_jll: Add Julia 1.8.1+ to LLD_jll 14 compat

jpsamaroo:
Unlike Julia 1.8.0, Julia 1.8.1 appears to also use LLVM 14, so we need to include `1.8.1-1` in the compat bounds for LLD_jll 14.

@vchuravy


jpsamaroo:
Nevermind, this only happens on Arch where they use a too-new LLVM (which I don't think we care to support)

PR #68116, opened 2022-09-12
Update URL for ShfitedArrays

piever:
[ShiftedArrays](https://github.com/JuliaArrays/ShiftedArrays.jl) was moved to JuliaArrays a while back. From what I understood, the JuliaRegistrator bot requires to manually change the URL here before tagging a new version.


DilumAluthge (approved):
I have confirmed that the old URL redirects to the new URL.

PR #68160, opened 2022-09-12
New version: ROCmDeviceLibs_jll v5.2.3+0

jlbuild:

* Registering JLL package ROCmDeviceLibs_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/ROCmDeviceLibs_jll.jl
* Version: v5.2.3+0
* Commit: 6e25cc7a80ca9ec047dcf83206cae5eaecba4c98
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/4ba58836cd8203e16abc754568790e6da2875a7a
* Created by: @pxl-th


pxl-th:
@vchuravy can you please take a look? I suspect it'll be with all ROCm 5.2.3 libraries...
Can it be avoided somehow?

pxl-th:
Or maybe @giordano can take a look...

vchuravy:
There are option, but none of them are nice ones. I assume the reason why this is limited to 1.9 is the LLVM version?

pxl-th:
> 

Yes, ROCm 5.2.3 uses LLVM 14

vchuravy:
Could it be loaded on a version of Julia that uses LLVM 13 or 15? Because we will encounter those scenarios eventually.

pxl-th:
If only loaded - yes, not sure about using it actually. Should I avoid setting `julia_compat=1.9` for now?

PR #68426, opened 2022-09-16
New package: rocSOLVER_jll v4.2.0+0

jlbuild:

* Registering JLL package rocSOLVER_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/rocSOLVER_jll.jl
* Version: v4.2.0+0
* Commit: be992388eee341e3e774f86142b711fa2fed5ea8
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/b82dd0493851f363b36cb2c57f610500148ba7fa
* Created by: @jpsamaroo


jpsamaroo:
This loads fine on 1.7; a different version will become available for 1.8 in the near future.

jpsamaroo:
@pxl-th does this seem right, or should we be removing the Julia compat instead?

pxl-th:
> @pxl-th does this seem right, or should we be removing the Julia compat instead?

I see that it used incorrect versions of dependencies, like ROCmOpenCLRuntime 5.2.3.
I'm thinking maybe we should really add back Julua 1.9 compats to ROCm 5.2.3?

jpsamaroo:
I think this should be fine to merge, since we've resolved the above issues with compat bounds on dependencies.

PR #68712, opened 2022-09-21
Fix ChebyshevTransforms link

Luapulu:
I moved the repo to JuliaApproximation, so the link needs to be updated.

### DIFF
--- a/C/ChebyshevTransforms/Package.toml
+++ b/C/ChebyshevTransforms/Package.toml
@@ -1,3 +1,3 @@
 name = "ChebyshevTransforms"
 uuid = "2752694e-a3a0-42bd-9f17-5f3242da1ce6"
-repo = "https://github.com/Luapulu/ChebyshevTransforms.jl.git"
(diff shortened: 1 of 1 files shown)


C/ChebyshevTransforms/Package.toml
```diff
@@ -1,3 +1,3 @@
 name = "ChebyshevTransforms"
 uuid = "2752694e-a3a0-42bd-9f17-5f3242da1ce6"
-repo = "https://github.com/Luapulu/ChebyshevTransforms.jl.git"
+repo = "https://github.com/JuliaApproximation/ChebyshevTransforms.jl"
```
fredrikekre, line 3:
```suggestion
repo = "https://github.com/JuliaApproximation/ChebyshevTransforms.jl.git"
```

fredrikekre (commented):
This doesn't look like a transfer, so the old URL doesn't automatically redirect. Would it be possible to fix this?

PR #69004, opened 2022-09-26
yank latest `Plots` (broken on windows)

t-bltg:
Until https://github.com/JuliaPlots/Plots.jl/issues/4375 is resolved.


t-bltg:
Many thanks @DilumAluthge.

PR #69061, opened 2022-09-27
Move SnowyOwl.jl to JuliaSingleCellOmics

yuehhua:


giordano:
Same comment as https://github.com/JuliaRegistries/General/pull/69058#issuecomment-1259973163

> I'm going to merge this because I verified the two repository have the same history, but in the future please follow https://github.com/JuliaRegistries/General/blob/2e106546f776d5a91d2fab35f51935ed9e45e03e/README.md#how-do-i-transfer-a-package-to-an-organization-or-another-user

PR #69581, opened 2022-10-05
Transfer ParameterSchedulers.jl to FluxML

darsnack:
Not sure if there is anything else to be done other than transferring the URL?


ericphanson:
Nope, that’s it! https://github.com/JuliaRegistries/General#how-do-i-transfer-a-package-to-an-organization-or-another-user

PR #69969, opened 2022-10-11
Fixed pathing to LaxZonedDateTimes

mattBrzezinski:
Forgot to make this change in https://github.com/JuliaRegistries/General/pull/69703


iamed2:
Issue https://github.com/JuliaRegistries/RegistryCI.jl/issues/310 from a previous occurrence. CI should probably prevent forgetting this.

PR #70698, opened 2022-10-20
Yank broken versions of LabelledArrays

chriselrod:
@ChrisRackauckas 
Broken versions fixed by: https://github.com/SciML/LabelledArrays.jl/pull/132


ChrisRackauckas:
Oh sorry about that. Yeah it should've bumped the lower bound.

chriselrod:
I ran into a couple of these `undefmatrix` issues because I had a few base packages built into my sysimage, so they weren't automatically upgrading.
On Julia 1.8, Pkg should make that okay by respecting the bounds when updating everything else -- but that only works if those bounds are correct.

Given I'm the only one that reported these problems, seems not many people are building light sysimages. They're probably either not doing it, or building a lot more into them.

PR #70721, opened 2022-10-20
Yank the 1.0.0+1 version of pprof_jll in favor of 1.0.1+0.

quinnj:


P/pprof_jll/Versions.toml
```diff
@@ -10,8 +10,5 @@ git-tree-sha1 = "6c056f2c6463c58a7083c1294710cf6ab7de697e"
 ["1.0.0+0"]
 git-tree-sha1 = "4a6859c2ffb30b8d6cb4ac73d2b6fcdd945c60c2"
 
-["1.0.0+1"]
-git-tree-sha1 = "d1a2f409c8573751f89f29065498f467ebc87785"
-
```
giordano, line 15: This isn't yanking.
```toml
["1.0.0+1"]
git-tree-sha1 = "d1a2f409c8573751f89f29065498f467ebc87785"
yanked = true
```
would be yanking.

quinnj: oops!

quinnj: ok, pushed a new version

giordano:
If you wanted to remove dependency on Graphviz for pprof_jll 1.0.0, you have also to fix https://github.com/JuliaRegistries/General/blob/cf0f29836861b9d8c6862873cdff4ac5494500c7/P/pprof_jll/Deps.toml#L9-L10

quinnj:
Ah, someone mentioned I _might_ have to do that; can you explain what needs to change there?

giordano:
`[1]` should becomes `[1.0.1]` or something like that (I don't know the exact syntax).

quinnj:
Ok, looking at other Deps.toml examples, I think it should be `["1.0.1-1"]`; I've pushed that now.

PR #70767, opened 2022-10-21
Yank Arpack `v0.5.0` and un-yank Arpack_jll `v3.8.0+0`

giordano:
Follow up to #59395.  In practice this doesn't change much because Arpack `v0.5.0` is not installable anyway, but it allows installing Arpack_jll `v3.8.0+0` if necessary.

CC: @ViralBShah @IanButterworth

### DIFF
--- a/A/Arpack/Versions.toml
+++ b/A/Arpack/Versions.toml
@@ -39,6 +39,7 @@ git-tree-sha1 = "2ff92b71ba1747c5fdd541f8fc87736d82f40ec9"
 
 ["0.5.0"]
 git-tree-sha1 = "288d58589d4249a63095f3f41ece91bf34c32c19"
+yanked = true
 
 ["0.5.1"]
 git-tree-sha1 = "9bdebb9148912e3c800dff80528b11064ddca424"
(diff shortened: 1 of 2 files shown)


ViralBShah:
Oh never mind - I did not read it right. These are just constraints on Arpack.jl versions 0.5 and 0.5.1.

The thing I am trying to do is test out new versions of Arpack_jll, and picking 3.7.0 which is not yanked, and bumping my dev version of Arpack.jl to v0.6.0 - but the resolver still won't let me:

```
(Arpack) pkg> up
    Updating registry at `~/.julia/registries/General.toml`
ERROR: Unsatisfiable requirements detected for package Arpack_jll [68821587]:
 Arpack_jll [68821587] log:
 ├─possible versions are: 3.5.0-3.5.1 or uninstalled
 └─restricted to versions 3.7.0-3 by Arpack [7d9fca2a] — no versions left
   └─Arpack [7d9fca2a] log:
     ├─possible versions are: 0.6.0 or uninstalled
     └─Arpack [7d9fca2a] is fixed to version 0.6.0
```

PR #71500, opened 2022-11-02
Restrict Arpack_jll v3.7+ to Julia v1.8+

staticfloat:
Because we gave libblastrampoline a versioned SONAME in Julia v1.8+, we broke compatibility with dependent libraries.  Arpack_jll from v3.7+ now uses the versioned SONAME which is incompatible with the older LBT libraries.


DilumAluthge:
CI is failing.

PR #71976, opened 2022-11-10
Pin GR to corresponding GR_jll

mkitti:
In anticipation of GR_jll v0.70 not being compatible with GR before v0.70, I would like to pin GR to the corresponding GR_jll within the minor version such that GR v0.69 is not compatible with GR_jll v0.70


mkitti:
Maybe this should be a single entry in case there are strange working combinations out there?

```
["0.53-0.69"]
GR_jll = ["0.53", "0.57", "0.58", "0.59", "0.60", "0.61", "0.62", "0.63", "0.64", "0.65", "0.66", "0.67", "0.68", "0.69"]
```

t-bltg:
I think this is a good idea, but it is hard to test if this will break anyone's workflow with older releases.

giordano:
Superseded by #71979

PR #71979, opened 2022-11-10
Pin GR to GR_jll, single entry

mkitti:
This is an alternative to #71976


G/GR/Compat.toml
```diff
@@ -52,3 +52,6 @@ Preferences = "1"
 
 ["0.69.1-0"]
 julia = "1.6.0-1"
+
+["0.53-0"]
+GR_jll = ["0.53", "0.57", "0.58", "0.59", "0.60", "0.61", "0.62", "0.63", "0.64", "0.65", "0.66", "0.67", "0.68", "0.69"]
```
t-bltg, line 57: Won't using range work here too ?
```toml
GR_jll = ["0.53", "0.57-0.69"]
```

mkitti: Done

PR #72022, opened 2022-11-10
Fix AutoMerge staging

ericphanson:
Right now, the project compat bounds it to v7, [breaking staging](https://github.com/JuliaRegistries/General/actions/runs/3440372735/jobs/5738763935). But we manually add the master branch anyway. So let's rm the project, as we already rm the manifests.

### DIFF
--- a/.github/workflows/automerge_staging.yml
+++ b/.github/workflows/automerge_staging.yml
@@ -29,10 +29,12 @@ jobs:
       - uses: julia-actions/setup-julia@0b9b1d2cd24245f151902702d8e73b3f6b910014 # v1.6.0
         with:
           version: ${{ matrix.version }}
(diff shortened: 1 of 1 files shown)


ericphanson:
Going to assume this is OK & merge but if there is post-hoc review I will address it in a follow up PR. (Only affects staging anyway).

DilumAluthge:
You can modify the compat entry in the Project file to allow both v7 and v8. Because RegistryCI v8 is not yet registered yet (IIRC), the "update manifests" action will not bump the regular (non-staging) manifests to include RegistryCI v8, and thus the regular (non-staging) jobs will continue to use RegistryCI v7.

ericphanson:
True, although I think it's nice if staging always uses master regardless of compat, since we generally want to try out the latest version

PR #72031, opened 2022-11-10
Rename ParametrisedConvexApproximators to ParameterizedConvexApproximators

JinraeKim:
as the name of the repo has been changed.


JinraeKim:
For someone seeing this:
I was following [this instruction for renaming registered packages](https://github.com/JuliaRegistries/General#how-do-i-rename-an-existing-registered-package) and I don't still know how it will end up being merged.

Is it ok to wait for a bit?

fredrikekre:
What do you want to achieve here? There is no way to rename a package, you will have to register it as a completely new one. Would be pretty unfortunate to have `ParametrisedConvexApproximators` and `ParameterizedConvexApproximators`...

JinraeKim:
> What do you want to achieve here? There is no way to rename a package, you will have to register it as a completely new one. Would be pretty unfortunate to have `ParametrisedConvexApproximators` and `ParameterizedConvexApproximators`...

@fredrikekre You're right. As I mentioned in [Julia discourse](https://discourse.julialang.org/t/how-to-rename-the-registered-package/90147/4?u=ihany), I know this is kinda unnecessary work but I was trying to match the same with the reference paper.
If you think this is not worthy it, I'll just close this PR. Please share your opinion.

PR #72087, opened 2022-11-12
README: explicitly mention that we don't allow empty packages

DilumAluthge:


Wimmerer:
Perhaps we could add a comment about early merging?  Or should we leave that undocumented to encourage people not to abuse it.

DilumAluthge:
Yeah, early merging isn't really something we have an official policy on; it's more of a courtesy that registry maintainers can choose to offer on a case-by-case basis, at their discretion.

And it's certainly not guaranteed, so I'd probably not want to list it in the README, because people might come to expect it.

ericphanson (approved):

fredrikekre:
Maybe this can go in this section? https://github.com/JuliaRegistries/General/blob/master/README.md#should-i-register-my-package

ericphanson:
Maybe we should also change that section title to “Should I register my package now?” (The new bit being “now”)

Because I think it can be interpreted as “well I plan to have those things eventually, so it fits”, instead of “I should have these things first”.

DilumAluthge:
I'm merging this just so that we can get something in, but @ericphanson and @fredrikekre want to make follow-up PRs to implement the improvements discussed above?

PR #72143, opened 2022-11-13
Bump to RegistryCI v8

ericphanson:


.ci/Project.toml
```diff
@@ -8,6 +8,6 @@ TimeZones = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 [compat]
 GitHub = "5"
 HTTP = "0.9, 1"
-RegistryCI = "7"
+RegistryCI = "7,8"
```
ericphanson, line 11: Actually, probably should only allow v8 (need to wait for registration though)
```suggestion
RegistryCI = "8"
```

.ci/Project.toml
```diff
@@ -8,6 +8,6 @@ TimeZones = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 [compat]
 GitHub = "5"
 HTTP = "0.9, 1"
-RegistryCI = "7"
+RegistryCI = "8"
```
DilumAluthge, line 11:
```suggestion
RegistryCI = "7, 8"
```

DilumAluthge (changes requested):
The order of operations should be:

1. Update the compat entry to include v8, but don't drop v7 from the compat entry.
2. Run the "update manifests" workflow, and merge the resulting PR.
3. Wait a while to make sure all is well. During this time, if something breaks, we can revert the "update manifests" PR.
4. After a while, make a PR to drop v7 from the compat entry.

ericphanson:
Ok, sounds good!

DilumAluthge (approved):

PR #72372, opened 2022-11-17
New version: libsingular_julia_jll v0.27.0+1

jlbuild:

* Registering JLL package libsingular_julia_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/libsingular_julia_jll.jl
* Version: v0.27.0+1
* Commit: 85d83ee91ae313bc8f986417b27ec05cde35664e
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/648eb2f519ce78b652dee235211a9b493f30781e
* Created by: @fingolfin


fingolfin:
@thofma @benlorenz so I missed that `libsingular_julia_jll` and `libpolymake_julia_jll` (see PR #72377) also restrict the `julia_compat` based on the "suported" Julia versions. Which is why this PR got stuck.

Do you remember why we did that (restrict `julia_compat`)? It makes such updates which just add a new Julia version a bit more annoying. We probably wanted to prevent other issues, but which? Might be a good idea to add a comment into the recipe.

In the meantime, how do we handle this? Just increase the versions of both JLLs?

giordano:
Is it ok now?

[noblock]

PR #72711, opened 2022-11-22
Revert "New package: GCTAccess v0.1.0"

mkitti:
Reverts JuliaRegistries/General#48281

Similar to https://github.com/JuliaRegistries/General/pull/48517

The entire content of this package is
```
function read(pa::String; na::String = "Name")::DataFrame

    return rename(
        select(
            DataFrame(File(pa; header = 3, delim = '\t')),
            Not("Description"),
        ),
        "Name" => na,
    )

end
```

cc: @KristofferC @StefanKarpinski @DilumAluthge

Issue #19033, opened 2020-08-05
What to do about "left-padded" packages (packages for which the Git repo has been deleted)


StefanKarpinski:
https://github.com/JuliaComputing/JuliaHub/issues/67

DilumAluthge:
We could remove those packages from the General registry.

They'll still be installable from the Pkg server.

DilumAluthge:
DecentralizedInternet is a false positive. The Julia package is in a subdirectory, which is allowed. https://github.com/Lonero-Team/Decentralized-Internet/tree/master/.julia-configs

fredrikekre:
> We could remove those packages from the General registry.
> They'll still be installable from the Pkg server.

But the PkgServer serves the registry so it has to be in there.

DilumAluthge:
In that list of packages that have been deleted on GitHub, we should subdivide the list into:
1. Packages that have no dependents in the General registry
2. Packages that have one or more dependents in the General registry

Then we can decide what to do with each of those two categories separately.

Gregliest:
Alright I did a first pass on analyzing the dependencies. First, the list that I generated and the list from the original issue are different. I did a spot check on the differences, and the extra packages that I checked from the original list are not currently in this registry. Examples include:
https://github.com/JuliaComputing/Blpapi.jl.git
https://github.com/JuliaComputing/MiletusPro.jl.git
https://github.com/aramirezreyes/SAMtools.jl.git
Also note that https://github.com/Lonero-Team/Decentralized-Internet.git appears to be working properly.

So, I'm assuming that the list I generated today is correct:
https://github.com/MUsmanZahid/TeXTable.jl
https://github.com/PetrKryslUCSD/MeshPorter.jl
https://github.com/mipals/SymSemiseparableMatrices.jl
https://github.com/owinter92/SimpleTweaks.jl
https://github.com/rjdverbeek-tud/Atmosphere.jl
https://github.com/anders-dc/Granular.jl
https://github.com/StanJulia/StanMCMCChain.jl
https://github.com/slmcbane/MirroredArrayViews.jl
https://github.com/mrtkp9993/Bioinformatics.jl
https://github.com/JonathanCourtois/MackeyGlass.jl
https://github.com/GuilhermeHaetinger/KelvinletsImage.jl
https://github.com/bcbi/CountdownNumbers.jl
https://github.com/MasFlam/Tui4jl
https://github.com/PetrKryslUCSD/MeshKeeper.jl
[...]

PR #72713, opened 2022-11-22
Revert "New package: GMTAccess v0.1.0"

mkitti:
Reverts JuliaRegistries/General#48280

Similar to https://github.com/JuliaRegistries/General/pull/48517
xref: https://github.com/JuliaRegistries/General/pull/72711

https://github.com/KwatMDPhD/GMTAccess.jl no longer exists.

The content consists of 

```julia
function read(pa::String)::Dict{String, Vector{String}}

    se_ge_ = Dict{String, Vector{String}}()

    for li in readlines(pa)

        sp_ = split(li, '\t')

        se = sp_[1]

        ge_ = [ge for ge in sp_[3:end] if ge != ""]

        if length(ge_) == 0

            println(se, " is empty")

        else

            se_ge_[se] = ge_

        end


    end

    return se_ge_

end

function read(pa_::Vector{String})::Dict{String, Vector{String}}

    se_ge_ = Dict{String, Vector{String}}()

    for pa in pa_

        merge!(se_ge_, read(pa))

    end

    return se_ge_

end
```

```julia
function write_txt(pa::String, se_ge_::Dict{String, Vector{String}})::Nothing

    if !endswith(pa, ".txt")

        error(pa, " is not .txt.")

    end

    open(pa, "w") do io

        for (se, ge_) in se_ge_

            write(io, string("-"^80, "\n"))

            write(io, string(se, " (", length(ge_), "):\n"))

            for ge in ge_

                write(io, string("  ", ge, "\n"))

            end

        end

    end

    return

end
```

cc: @KristofferC @StefanKarpinski @DilumAluthge

### DIFF
--- a/G/GMTAccess/Compat.toml
+++ b/G/GMTAccess/Compat.toml
@@ -1,4 +1,0 @@
-[0]
-BenchmarkTools = "1.2.0-1"
-Revise = "3.1.20-3"
-julia = "1.6.0-1"
--- a/G/GMTAccess/Deps.toml
+++ b/G/GMTAccess/Deps.toml
@@ -1,3 +1,0 @@
-[0]
-BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
-Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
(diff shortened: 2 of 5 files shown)


fredrikekre:
Why? Maybe it was a mistake to merge it, but it has been a year at this point.

mkitti:
Oh, 2021. I'll just leave this for a day for others to chime in and cross reference the left pad issue.

 xref: #19033

mkitti:
What is the threshold to catch packages like this?

PR #72976, opened 2022-11-28
Fix the Markdown rendering of some README links

DilumAluthge:
~~For some reason, the normal (inline) syntax isn't rendering correctly here. So let's switch these two links to footnotes.~~


fredrikekre:
The problem is the line break between `]` and `(`.

ericphanson:
thank you!

PR #73082, opened 2022-11-29
Change repo URL because of change of ownership (from antonuccig to gi…

giancarloantonucci:
…ancarloantonucci)


giancarloantonucci:
Can someone merge this?

giancarloantonucci:
That's me, but I see that I have no way to prove it... Or have I? Is there a way to show you that I've changed username? (in particular, see this: https://discourse.julialang.org/t/can-i-change-my-username-so-that-it-matches-my-github-username/70173)

ViralBShah:
I can verify it by going to the old URL and it redirects. However it redirects to https://github.com/giancarloantonucci/NSDETimeParallel.jl.

giancarloantonucci:
Changed it back (I want to put a big deprecation note before submit a new package with the new name). Can you merge this now?

PR #73120, opened 2022-11-29
yank

rayegun:
@vchuravy


P/PARMETIS_jll/Compat.toml
```diff
@@ -9,6 +9,5 @@ julia = "1.6.0-1"
 
 ["4.0.5-4"]
 JLLWrappers = "1.4.0-1"
-METIS_jll = "5.1.2-5"
```
fredrikekre, line 12: Why this change?

vchuravy: This was spuriously introduced and the reason for the yank. Previous build IDs didn't have this dependency.

We should have done a version bump instead

PR #73304, opened 2022-12-02
Yank ForwardDiff 0.10.33

mcabbott:
This turns out to cause more disruption than intended, so perhaps it should be 0.11 not a patch version. 

The offending PR is https://github.com/JuliaDiff/ForwardDiff.jl/pull/481 

Some reports of issues aren't issues, but the ones that are are:
https://github.com/JuliaDiff/ForwardDiff.jl/issues/606
https://github.com/JuliaDiff/ForwardDiff.jl/issues/607
https://github.com/JuliaDiff/ForwardDiff.jl/issues/609

### DIFF
--- a/F/ForwardDiff/Versions.toml
+++ b/F/ForwardDiff/Versions.toml
@@ -102,3 +102,4 @@ git-tree-sha1 = "187198a4ed8ccd7b5d99c41b69c679269ea2b2d4"
 
 ["0.10.33"]
 git-tree-sha1 = "10fa12fe96e4d76acfa738f4df2126589a67374f"
+yanked = true


oxinabox:
I think this case it would have been better to just revert and release 0.10.34,  without yanking 0.10.33

Like all patch releases, (though not true pre-1.0.0) the older version had bugs.
But we don't yank all penultimate patch versions. (Though I have suggested that Pkg should by default treat them that way, but that's another dicussion)

The particular thing that yanking does that just making a new release doesn't is it means you can do `]add --preserve=all Foo` for minimal changes to a known working system. (nor `update ---preserve=all Foo` in 1.9 I suspect).
And this broke stuff for Invenia's workflow trying to make minimal changes to prod.

mcabbott:
Well IDK, there appear to be many opinions. Want to make a branch for 0.10 versions & tag that as 0.10.34?

Reverting things seems to be a way to make sure they rot.

oxinabox:
revert release patch, then unrevert and release breaking IMO is best

> Well IDK, there appear to be many opinions. Want to make a branch for 0.10 versions & tag that as 0.10.34?

It wouldn't help. The problem is the yanking of 0.10.33 being more disruptive than just releasing a 0.10.34 that is identical to 0.10.32.

PR #73880, opened 2022-12-11
Bound all OrdinaryDiffEq <6.34 to DiffEqBase <6.109.0

DilumAluthge:
This PR was generated by the following script:

```julia
import Pkg
import RegistryTools
import TOML

const Compress = RegistryTools.Compress

ode_name = "OrdinaryDiffEq"
ode_versions_str = "<6.34"
ode_versions = Pkg.Types.semver_spec(ode_versions_str)
debase_name = "DiffEqBase"
debase_constraint_str = "<6.109.0"
debase_constraint = Pkg.Types.semver_spec(debase_constraint_str)

branch_name = "dpa/ode-debase-compat"
commit_message = "Bound all $(ode_name) $(ode_versions_str) to $(debase_name) $(debase_constraint_str)"

mktempdir() do dir
    cd(dir) do
        run(`git clone git@github.com:JuliaRegistries/General.git`)
        cd("General") do
            run(`git checkout master`)
            run(`git checkout -B $(branch_name)`)
            registry = TOML.parsefile("Registry.toml")
            ode_infos = [x for x in values(registry["packages"]) if x["name"] == ode_name]
            ode_info = only(ode_infos)
            ode_path = ode_info["path"]
            ode_compat_file = joinpath(ode_path, "Compat.toml")
            compat = Compress.load(ode_compat_file)
            for (k, v) in compat
                if k in ode_versions
                    old_compat_str = v[debase_name]
                    old_compat = Pkg.Types.VersionSpec(old_compat_str)
                    new_compat = intersect(old_compat, debase_constraint)
                    new_compat_str = string(new_compat)
                    v[debase_name] = new_compat_str
                end
            end
            Compress.save(ode_compat_file, compat)
            run(`git --no-pager diff HEAD`)
            run(`git add -A`)
            run(`git commit -m "$(commit_message)"`)
            run(`git push -f origin $(branch_name)`)
        end
    end
end
```

### DIFF
--- a/O/OrdinaryDiffEq/Compat.toml
+++ b/O/OrdinaryDiffEq/Compat.toml
@@ -122,7 +122,7 @@ Parameters = "0.11"
 Parameters = "0.10-0.11"
 
 ["5.16-5.17"]
-DiffEqBase = "6"
+DiffEqBase = "6-6.108"
 
 ["5.17-5.29"]
 Parameters = "0.10-0.12"
@@ -349,7 +349,7 @@ SciMLBase = "1.35.1-1"
 FastBroadcast = "0.1.9-0.2"
 
 ["6.15-6.25"]
-DiffEqBase = "6.81.3-6"
+DiffEqBase = "6.81.3-6.108"
 
 ["6.17-6"]
 ArrayInterfaceGPUArrays = "0.1-0.2"
@@ -379,7 +379,7 @@ SparseDiffTools = "1.26.2-1"
 SciMLBase = "1.53.0-1"
 
 ["6.26-6.33"]
-DiffEqBase = "6.99.0-6"
+DiffEqBase = "6.99.0-6.108"
 
 ["6.27-6"]
 Preferences = "1.3.0-1"


DilumAluthge:
cc: @ChrisRackauckas @chriselrod

ChrisRackauckas:
Well part of this is that I'm ripping off bandaids right now. There's a bunch of silly shit we did in 2016, and now that we have SciML global docs (https://docs.sciml.ai/Overview/stable/) things like "LinearSolve, NonlinearSolve, DifferentialEquations, GalacticOptim... wait that last one doesn't fit..." is just too obvious. So we've been on a rampage the last 6 months to do things like, rip out the nonlinear solver code that is in DiffEqBase and only half used, with it then calling halfway to OrdinaryDiffEq and NonlinearSolve, to actually just being internal to OrdinaryDiffEq and NonlinearSolve (which is what this was). All of the interfaces are getting documented and tested (see https://docs.sciml.ai/SciMLBase/stable/). But some of this requires acknowledging that one package used some internal that it should've have, ripping it out and making it use the now public API, and suffering the consequences. Otherwise we would've had like 6 breaking updates to a package with 300 downstreams this summer, even though almost all of the issues were just to one package.

But, I think we're just about done with that project. The last piece is https://docs.sciml.ai/SciMLOperators/stable/, but that will have a slower deprecation because DiffEqOperators are more user-facing.

chriselrod:
Was JuliaRegistries deleting my comment intentional?
It was a comment on why I kept running into these issues while seemingly few other people had, or at least few other people reported issues.

More typically, I'm on older fixed versions of everything (but then you have to be careful to use that same manifest when building the image, which is probably what I should have done), or on bleeding edges of everything.
But I was using an odd-ball assortment of recent and old versions, thereby testing combinations of packages together that few other people probably encountered.

Anyway, I agree that it is good to rip this bandaid off.
Consistent APIs, etc, are important. In an ideal world, things are consistent enough that people can guess what the API without having to look things up.
Consistent APIs are much easier to remember; remembering a couple patterns and rules is way easier than memorizing a litany of conventions.

PR #73935, opened 2022-12-12
URL change: DataToolkit, http → https

tecosaur:
I tried to `@JuliaRegistrator register`, and it complained about a URL change which is probably resolved by this.


tecosaur:
Thanks :slightly_smiling_face:

PR #73995, opened 2022-12-12
New version: Binutils_jll v2.39.0+0

jlbuild:

* Registering JLL package Binutils_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/Binutils_jll.jl
* Version: v2.39.0+0
* Commit: 4e95574b9d8d3c5a07a518ee6fbf53f7c6eb359e
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/db950e2fd0074d30caff75f69d78ff22e4fcc58c
* Created by: @mofeing


mofeing:
@giordano Is there sth wrong with my recipe?

giordano:
Apparently yes: `c++filt` isn't an acceptable name.  I should have noticed that you used `Symbol("c++filt")`, that should have been a red flag.  Can use something like `"cxxfilt"` instead.  But I also thought we had a check that the names have to be valid identifiers.

[noblock]

PR #74176, opened 2022-12-15
yank `1.0.2`

t-bltg:
https://github.com/JuliaPlots/Plots.jl/issues/4597
https://discourse.julialang.org/t/error-precompiling-plots-recipesbase-jl-607-needs-to-be-placed-at-the-top-level-or-use-eval/91694
https://github.com/timholy/SnoopCompile.jl/issues/317


ChrisRackauckas:
I'm going to just go ahead with this because there are a ton of reports about this right now. @timholy

timholy:
Glad you merged this quickly, sorry for the trouble.

ChrisRackauckas:
No worries, it happens. Happens to some of us more than others. I'm way more used to yanking things hahahaha

PR #74378, opened 2022-12-20
Revert "Restrict Arpack_jll v3.7+ to Julia v1.8+"

DilumAluthge:
Reverts JuliaRegistries/General#71500

Reverting per @rofinn's comment.

@staticfloat When we re-land, I think we have to make sure that for each version that has a compat entry for JLLWrappers, it also has it as a dep.

@fredrikekre Not sure why our CI didn't catch this?


odow:
Thanks for the quick fix!

DilumAluthge:
Now we just need to re-land a fixed version.

giordano:
> When we re-land, I think we have to make sure that for each version that has a compat entry for JLLWrappers, it also has it as a dep.

Sounds like RegistryCI is missing a check, this is something CI should catch.

PR #74621, opened 2022-12-25
Update url for Glycolysis.jl package

Denis-Titov:
I move Glycolysis.jl package from https://github.com/Denis-Titov/Glycolysis.jl to https://github.com/DenisTitovLab/Glycolysis.jl and want to update this in the registry. Hopefully this is the correct way to do this


Denis-Titov:
Just to add, I beleive I followed the guidelines here: https://github.com/JuliaRegistries/General/blob/06a950203a6c3857b1b0d741d581006bf6ebb88e/CONTRIBUTING.md

You can see that using old url: https://github.com/Denis-Titov/Glycolysis.jl lead to the page with new url: https://github.com/DenisTitovLab/Glycolysis.jl

Please let me know if I'm doing something wrong.

PR #58361, opened 2022-04-11
Transfer from kimmolinna to duckdb

jeremiahpslewis:
@kimmolinna Just trying to clean up loose ends here, think this is necessary for the 'handover' of the package to duckdb.

PR #75319, opened 2023-01-08
Update Package.toml

ngiann:
Changing URL to original repo

### DIFF
--- a/F/FastParzenWindows/Package.toml
+++ b/F/FastParzenWindows/Package.toml
@@ -1,3 +1,3 @@
 name = "FastParzenWindows"
 uuid = "d2256c36-f6cb-4f2b-bb45-9577641c7b75"
-repo = "https://github.com/ngiann/FastParzenWindows.jl"
+repo = "https://github.com/ngiann/FastParzenWindows.jl.git"
--- a/F/FastParzenWindows/Package.toml
+++ b/F/FastParzenWindows/Package.toml
@@ -1,3 +1,3 @@
 name = "FastParzenWindows"
 uuid = "d2256c36-f6cb-4f2b-bb45-9577641c7b75"
-repo = "https://github.com/HITS-AIN/FastParzenWindows.jl.git"
+repo = "https://github.com/ngiann/FastParzenWindows.jl"


ericphanson:
I’m concerned that old versions might not be installable. All the registrered tree hashes must be present in the new repo. Someone needs to confirm that is the case. Check out previous threads like https://github.com/JuliaRegistries/General/pull/58361

F/FastParzenWindows/Package.toml
```diff
@@ -1,3 +1,3 @@
 name = "FastParzenWindows"
 uuid = "d2256c36-f6cb-4f2b-bb45-9577641c7b75"
-repo = "https://github.com/HITS-AIN/FastParzenWindows.jl.git"
+repo = "https://github.com/ngiann/FastParzenWindows.jl"
```
ericphanson, line 3: Should still end with `.git`

ngiann: Thanks. Concerning older versions: I am very doubtful that this package is used by anyone else as it has not been advertised.

ngiann:
Concerning older versions: I am very doubtful that this package is used by anyone else as it has not been advertised. I can only offer my personal reassurances that it is not used. I can always insert a message in the currently registered repo that the URL has been updated. If this is not convincing, then let us drop this issue. Thanks for your input.

ericphanson:
It's OK if the repo has been updated, the problem is if older versions become uninstallable due to changing the URL here. If the code associated to a version of the package exists in the old repo but not the new repo, and the package manager looks in the new repo for that code and doesn't find it, suddenly the old working version will abruptly stop working (without the user doing anything). Even if there's very few users, that's not something the General registry wants to ever happen.

However, that doesn't mean you can't ever change repos. It just means the new repo has to contain all the old versions in the commit history somewhere. We can check this as follows, based on the code from @GunnarFarneback [here](https://github.com/JuliaRegistries/General/pull/35965#issuecomment-832721704). First we clone the new repo, cd into it, and start a julia session:
```bash
git clone https://github.com/ngiann/FastParzenWindows.jl.git
cd FastParzenWindows.jl
julia
```

Here, git cloning gives me this warning:
```bash
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'src/FastParzenWindows.jl'
  'src/fastparzenwindows.jl'

```
I think this is a real problem and means your package will be impossible to use on some systems - however, AutoMerge should catch this issue next time you try to register a version, so we don't need to deal with it now.

Then in our Julia session, we use the package RegistryInstances to check which tree hashes corresponding to versions of this package have been registered, and check if they are present in the new repo:
```julia
using RegistryInstances, UUIDs
pkg_name = "FastParzenWindows"

# Load General registry
general = only(filter!(r -> r.uuid == UUID("23338594-aafe-5451-b93e-139f81909106"), reachable_registries()))
pkg = general.pkgs[only(uuids_from_name(general, pkg_name))]
versions = registry_info(pkg).version_info
for version in sort(collect(keys(versions)))
    tree_sha = versions[version].git_tree_sha1
    print(version)
    try
        println(" found")
    catch
        println(" missing")
    end
end
```

This gives me just

```julia
0.1.2 found
```

Great! That means the only old version still has code present in the new repo, and changing the URL should be OK. I'll merge this then. However you'll want to take care of the file name clash before you register your next version of the package.

ngiann:
I can't thank you enough for taking the time and interest to provide me
with such a detailed answer! To be honest, I don't fully understand all the
technical details you mention, but I will reproduce the steps you outlined
on my local machine to better understand the situation. I will rename what
used to be the registered repo to something else. Many thanks again!

PR #77419, opened 2023-02-10
SHA v0.6+ don't require `Compat`

staticfloat:
This removes the erroneous dependency entry claiming that the SHA stdlib requires `Compat` (it dropped that requirement when it upgraded from `REQUIRES` format to `Project.toml` format: https://github.com/JuliaCrypto/SHA.jl/commit/38066fd19f31eb47469f9f3d2ae5f083449ea0c4).


staticfloat:
I'm not sure what the right syntax is here.  @DilumAluthge can you help me out?

DilumAluthge:
Yeah I can take a look. Generally it's easier to automate this process (using RegistryTools.Compress) than to edit them by hand.

So for any SHA.jl version >= 0.6.0, delete the dependency on Compat.jl?

staticfloat:
Ah, I figured out what the problem is; there are no versions higher than 0.5!  I'll add registrations for them.  :)

PR #78788, opened 2023-03-02
Move Mocking.jl to JuliaTesting

omus:
A simple repository ownership transfer, the git history is fully present.


DilumAluthge (approved):
I confirmed that the old URL redirects to the new URL.

PR #79083, opened 2023-03-07
Retroactively add PCRE_jll as dependency of OpenSlide_jll.

GunnarFarneback:
Due to specifics of the Yggdrasil build, the OpenSlide library ended up with PCRE as a direct library dependency, although it had only been brought in indirectly by the Glib library. For practical purposes this was masked by the Glib_jll dependency until Glib_jll 2.74, when Glib was built with PCRE2 instead of PCRE.

This manual PR retroactively adds PCRE_jll as a dependency of (the only existing version of) OpenSlide_jll so it works also with Glib_jll >= 2.74.


GunnarFarneback:
Sadly not effective when tested in a clean depot.

PR #79127, opened 2023-03-08
Fork MicroCollections.jl to the JuliaFolds2 organization

DilumAluthge:


ViralBShah:
Should we do a bunch of these in the same PR?

simsurace:
Maybe better to merge this one right away if there is insufficient manpower to fork the other ones as well?

PR #79133, opened 2023-03-08
Update Bigleaf URL to post-rename repo

bgctw:


bgctw:
After a discussion on Slack, I learned that renaming the case of a package creates more problem than it solvers. An example is having two packages with the same name on lower-case file systems in the local registry. 

Hence, I have to refrain from renaming the package. Please ignore this request, because I will rename the repository to its old name.

PR #80632, opened 2023-03-30
Add julia compat bounds to Ipopt_jll

Vaibhavdixit02:
@odow do you think this is reasonable?


odow:
Not really. Everything works on v1.8, except the Windows executable. And I don't know if we have shown that it works on Julia 1.9?

cc @amontoison

amontoison:
@Vaibhavdixit02 Why do you want to update the julia compat of Ipopt_jll?

Vaibhavdixit02:
>Everything works on v1.8, except the Windows executable.

Not the releases, right? The issue was not just in Windows, it doesn't work on mac and linux also https://github.com/jump-dev/AmplNLWriter.jl/actions/runs/4466436440/jobs/7844624260. Your PR https://github.com/jump-dev/AmplNLWriter.jl/pull/163 iiuc fixes it for everything other than Windows though?

>Why do you want to update the julia compat of Ipopt_jll?

Based on the comments https://github.com/jump-dev/AmplNLWriter.jl/pull/163#issuecomment-1480377235 and https://github.com/jump-dev/AmplNLWriter.jl/pull/163#issuecomment-1480380035 my understanding is that post 1.9 we will able to ship Ipopt_jll with only LBT and avoid the issue (https://github.com/jump-dev/AmplNLWriter.jl/pull/162)

odow:
Yip, the other platforms fail because we haven't merged https://github.com/jump-dev/AmplNLWriter.jl/pull/163.

> my understanding is that post 1.9 we will able to ship Ipopt_jll with only LBT and avoid the issue

I don't know if we've confirmed that that will fix the issue?

Vaibhavdixit02:
>I don't know if we've confirmed that that will fix the issue?

That's true, but we also know that it doesn't work on anything <1.9 😅

amontoison:
@Vaibhavdixit02
I'm sorry but I still don't understand the issue.
The `Ipopt_jll` precompiled works for platforms if we use it with Ipopt.jl.
If we use `Ipopt_jll` with `AmplNLWriter.jl` we need some modifications to support the artifacts precompiled with LBT but the release 1.0.2 fixed all problems by using the last `Ipopt_jll` precompiled with OpenBLAS32_jll.

Vaibhavdixit02:
I haven't really explained the issue I realise that now, sorry. The thing is https://github.com/SciML/ModelingToolkit.jl/actions/runs/4559568146/jobs/8071636678#step:6:41 because Ipopt_jll > 400 is available, the resolver doesn't pick AmplNLWriter 1.0.2 (or OptimizationMOI 0.1.11)

PR #81389, opened 2023-04-11
Yank OptimizationFlux, OptimizationMOI, OptimizationNLopt, OptimizationOptimJL, OptimizationSpeedMapping and OptimizationGCMAES

Vaibhavdixit02:


fredrikekre:
Context? It is not recommended to yank when you will just release a patch anyway (https://github.com/JuliaRegistries/General/pull/81391) in particular when you yank feature releases, which people might have compat on.

ChrisRackauckas:
All of these are missing a crucial lower bound so they are not good to have in the ecosystem. Other tags are coming with the fixed bound.

PR #82386, opened 2023-04-26
Yank StatsBase version 0.33.22

andreasnoack:
It's breaking. See e.g. https://github.com/JuliaStats/KernelDensity.jl/pull/114 and https://github.com/VaclavMacha/EvalMetrics.jl/blob/master/src/EvalMetrics.jl#L7. We should tag a 0.34.0 instead.

### DIFF
--- a/S/StatsBase/Versions.toml
+++ b/S/StatsBase/Versions.toml
@@ -102,3 +102,4 @@ git-tree-sha1 = "d1bf48bfcc554a3761a133fe3a9bb01488e06916"
 
 ["0.33.22"]
 git-tree-sha1 = "9e07a3a1e7373f389c87f8fdb932667cebbfbe23"
+yanked = true


DilumAluthge:
> It's breaking. See e.g. [JuliaStats/KernelDensity.jl#114](https://github.com/JuliaStats/KernelDensity.jl/pull/114) and https://github.com/VaclavMacha/EvalMetrics.jl/blob/master/src/EvalMetrics.jl#L7. We should tag a 0.34.0 instead.

Could you tag a 1.0.0 instead?

andreasnoack:
Let's hear the other stats folks as well. I'm not a believer in the stabilitizing effect of 1.0 but I'm also not against a 1.0.

svilupp:
FYI. The above error propagates to StatsPlots.jl, which has KernelDensity as its dependency, ie, I think a lot of people will be impacted

MilesCranmer:
SymbolicRegression.jl's tests seem to be breaking due to this as well

PR #82638, opened 2023-05-01
redirect repos

jd-lara:
This PR redirects the repos after the recent re-brand of the tools

### DIFF
--- a/I/InfrastructureSystems/Package.toml
+++ b/I/InfrastructureSystems/Package.toml
@@ -1,3 +1,3 @@
 name = "InfrastructureSystems"
 uuid = "2cd47ed4-ca9b-11e9-27f2-ab636a7671f1"
-repo = "https://github.com/NREL-SIIP/InfrastructureSystems.jl.git"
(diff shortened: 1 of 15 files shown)


StefanKarpinski:
Does GitHub not redirect changed user/org names? Just wondering, since as this stands we kind of have to trust that this new org is legit (although we know who you are, so I believe it, but we need to be careful).

jd-lara:
> Does GitHub not redirect changed user/org names? Just wondering, since as this stands we kind of have to trust that this new org is legit (although we know who you are, so I believe it, but we need to be careful).

yes, GitHub redirects but I opened this PR based on the recommendation in the README. Here is a link to the organizational link https://www.nrel.gov/analysis/sienna.html

jd-lara:
@StefanKarpinski it looks like the re-direct doesn't work anymore. 

image


I would appreciate merging to PR to be able to release the latests versions of all the packages please.

jd-lara:
bump to @StefanKarpinski about this PR, we need to tag new package versions urgently and we can't at the moment. See message above.

PR #84925, opened 2023-06-05
Retrospectively require Julia 1.6

timholy:
In response to https://github.com/JuliaDebug/Cthulhu.jl/issues/463,
this retrospectively requires Julia 1.6 or higher.

Note this *nominally* breaks earlier releases, although it seems quite unlikely that any of those ever could have worked. I will let someone who knows more about the policies of General make the decision about whether to merge this.


T/TypedSyntax/Compat.toml
```diff
@@ -1,5 +1,5 @@
 ["1-1.1"]
```
giordano, line 1: I think this should be
```suggestion
[1]
```
and you remove the last section?

timholy:
I'll merge this soon barring comments to the contrary.

PR #85087, opened 2023-06-07
LinearSolve.jl v2 broken on Julia 1.7/1.8

simonbyrne:
See https://github.com/SciML/LinearSolve.jl/issues/323

cc @ChrisRackauckas

### DIFF
--- a/L/LinearSolve/Compat.toml
+++ b/L/LinearSolve/Compat.toml
@@ -27,7 +27,7 @@ RecursiveFactorization = "0.2"
 ["0.1.2-1.9"]
 Krylov = "0.7.9-0.7"
 
-["0.1.2-2"]
+["0.1.2-1"]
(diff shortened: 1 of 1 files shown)


L/LinearSolve/Compat.toml
```diff
@@ -136,6 +136,7 @@ EnumX = "1"
 RecursiveFactorization = "0.2.8-0.2"
 
 [2]
+julia = "1.6.0-1.6, 1.9-1"
```
simonbyrne, line 139: How do I say v1.6 or v1.9+?

Vaibhavdixit02:
```suggestion
julia = "1.6.0-1.6,  >= 1.9"
```

Would this work?

simonbyrne:
A new version has been tagged, so this should now be resolved.

PR #85431, opened 2023-06-13
Add Julia <= 1.8 bound for old versions of SortingAlgorithms.jl

frankier:
See https://github.com/JuliaCollections/SortingAlgorithms.jl/pull/63

For some reason I sometimes get old versions of SortingAlgorithms pulled in on Julia 1.9 which does not precompile.

@LilithHafner / @oscardssmith would it be an idea to version bound the old SortingAlgorithm versions?


S/SortingAlgorithms/Compat.toml
```diff
@@ -1,7 +1,11 @@
 ["0-0.3.1"]
 DataStructures = "0.9.0-0.18"
-julia = ["0.7", "1"]
+julia = ["0.7", "1-1.8"]
 
-["0.3.2-1"]
+["0.3.2-0"]
```
LilithHafner, line 5:
```suggestion
["0.3.2-1.0"]
```
I think the current change fails to specify anything for versions 1.0.0 and 1.0.1.

LilithHafner (commented):
This is what happens when a package depends on Julia internals in order to be loaded with `using`. Retroactively bounding the allowed Julia version on old versions of SortingAlgorithms is the right thing to do.

Where is the syntax of this file documented? I can't give a reliable technical review without it.

LilithHafner:
Bump: Old versions of SortingAlgorithms don't load on new versions of Julia. This change should be made.

ViralBShah:
Should this be merged?

LilithHafner:
SortingAlgorithms 1.1.0 is the first version with support for Julia 1.9+, this PR is still relevant and should be merged (if the compat spec update is syntactically correct)

PR #85487, opened 2023-06-13
Eliminate date-versioned patchelf build

staticfloat:
Patchelf originally was built on a version that had nothing to do with any version number, so we applied a date version.  However, now `Pkg` things that version is the latest, so it always chooses that one when you try to install it.  To fix this, just drop that old version.


DilumAluthge:
Can you just yank it instead?

staticfloat:
I'm fine with either way; is yanking preferred?

DilumAluthge:
Yeah, I generally prefer yanking over removing versions entirely.

DilumAluthge (approved):

PR #85489, opened 2023-06-13
Update README.md to make renaming clearer

eirikbrandsaas:
Make it explicit that you also have to rename the package in the Project.toml file


giordano:
Thanks!

PR #86561, opened 2023-06-30
Change IGLWrap_jll package URL

peremato:


giordano:
@plut are you OK with taking over your package?

peremato:
I'm somehow blocked here. If the URL is not changed I cannot upgrade the version of IGLWrap_jll and I cannot make a new release of a package that requires it. The only solution I see is that I create a different package _jll for the same C++ library. But I do not think this is desired.

giordano:
Ok, we haven't heard from @plut for a few weeks and this is preventing us from developing this package on Yggdrasil.

I'm taking the executive decision to take over this JLL package so that we can continue working on it collaboratively on Yggdrasil.  Of course @plut will be welcome to contribute there.

peremato:
@giordano Thanks very much! Can you  please re-trigger the registration of https://github.com/JuliaPackaging/Yggdrasil/pull/6943, or is this done differently.

PR #86946, opened 2023-07-06
Relax Perl_jll - Readline_jll compat

benlorenz:
Readline does seem to follow semver, soname is `libreadline.so.8`.
I checked locally that Perl_jll built with Readline 8.1.1 works with 8.2.1 as well.

I will adjust the compat entry in Yggdrasil afterwards.

cc: @thofma @fingolfin

### DIFF
--- a/jll/P/Perl_jll/Compat.toml
+++ b/jll/P/Perl_jll/Compat.toml
@@ -4,5 +4,5 @@ julia = "1"
 
 ["5.34-5"]
 JLLWrappers = "1.2.0-1"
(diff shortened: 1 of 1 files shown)


thofma:
Thanks!

fingolfin:
Thank you. In related news, I just rebuilt GAP_jll against older readline to fix a runtime failure to load it on macOS (error happend when installed GAP_jll was new but Readline_jll old). See https://github.com/JuliaPackaging/Yggdrasil/pull/7012

PR #86950, opened 2023-07-06
Restrict StaticArrays.jl @v1.5- to Julia v1.10-

dkarrasch:
@vtjnash Is this how it's done? Related to https://github.com/JuliaLang/julia/pull/26307/


S/StaticArrays/Compat.toml
```diff
@@ -1,10 +1,19 @@
 ["0-0.10"]
-julia = ["0.7", "1"]
+julia = ["0.7", "1-1.10"]
```
giordano, line 2: This doesn't match the title of the PR, you're allowing Julia v1.10, or am I missing something?

dkarrasch: You're right. I thought we might include the `AbstractTriangular` commit, which removed the second type parameter in v1.10, but I guess we are not (it would need to get backported), so I changed it accordingly.

S/StaticArrays/Compat.toml
```diff
@@ -1,10 +1,19 @@
...
 
-["0.11-1.3"]
+["0.11"]
+julia = "1-1.10"
+
+["0.12"]
```
giordano, line 7: Based on https://github.com/JuliaRegistries/General/actions/runs/5533534662/jobs/10097202254?pr=86950#step:16:9:
```suggestion
["0.12-0"]
```

dkarrasch: Thanks! I was unable to read that output.

S/StaticArrays/Compat.toml
```diff
@@ -1,10 +1,19 @@
 ["0-0.10"]
-julia = ["0.7", "1"]
+julia = ["0.7", "1-1.10"]
 
-["0.11-1.3"]
+["0.11"]
+julia = "1-1.10"
+
+["0.12-0"]
 julia = "1"
```
giordano, line 8: Just to confirm, this version _is_ compatible with Julia v1.11+?

dkarrasch: Yes, we backported the relevant commit to v0.12.5 or so.

fhagemann:
The code erroring in julia-1.11 was removed from `1.5.16` to `1.5.17`, see [here](https://github.com/JuliaArrays/StaticArrays.jl/pull/1136/commits/fdf353022cde23b049ad59e4b7710995458276f4).
Is there a reason why the versions `1.5.0-1.5.16` (which still cause errors on julia-1.11) are not excluded as well?

image

dkarrasch:
In this case it seems to be an oversight.

PR #88097, opened 2023-07-22
yank DataFrames.jl 1.6.0

bkamins:
DataFrames.jl 1.6.0 contained incomplete Project.toml specification. Now that we have DataFrames.jl 1.6.1 released we want to signal Julia package manager that DataFrames.jl 1.6.0 should not be considered when performing package version resolution.

@KristofferC, @DilumAluthge - is this the proper way to achieve this effect? Thank you!

CC @nalimilan @davidanthoff @palday


DilumAluthge (approved):

bkamins:
Thank you!

PR #88183, opened 2023-07-24
Update LibPQ URL invenia -> iamed2

iamed2:
Repository was moved to follow maintainer (me) after the demise of Invenia.


aviks:
Would it make sense to move this to an org, say `juliadatabases`? Just a thought.

PR #81657, opened 2023-04-15
Extend to libcurl 8.0.1

joa-quim:

PR #88557, opened 2023-07-28
[JLLs] Allow LibCURL_jll v8

visr:
This addresses https://github.com/JuliaPackaging/Yggdrasil/issues/6857#issuecomment-1581518100, which should fix a bunch of packages that are broken on 1.10.

Continuation of https://github.com/JuliaRegistries/General/pull/81657 for the rest of the packages.


visr:
@giordano does this look good to you?

PR #90310, opened 2023-08-25
Fix StaticArrays and FillArrays compat bounds for Statistics

nalimilan:
On Julia 1.1 to 1.3, `Pkg.test` fails for these two packages as they have been marked to require Statistics 1.10 or below in accordance to their dependency on Julia 1.0 or below, but only Statistics 1.11 is registered. Registering an older Statistics version is apparently not enough to fix the problem, probably because of a resolver issue. Allow Statistics 1.11 as it works.

See https://github.com/JuliaRegistries/General/pull/89713#issuecomment-1693460400.

### DIFF
--- a/F/FillArrays/Compat.toml
+++ b/F/FillArrays/Compat.toml
@@ -3,7 +3,7 @@ Compat = "0.35-2"
 julia = "0.6-1.10"
 
 ["0.12.1-0.12"]
-Statistics = "1-1.10"
+Statistics = "1-1.11"
 
 ["0.13-1"]
 Statistics = "1.6.0-1"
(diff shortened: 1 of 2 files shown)


baumgold:
Hi, just checking on the status here?  Thanks.

stemann:
> Hi, just checking on the status here? Thanks.

Fix confirmed to re-enable testing e.g. on Julia 1.3: https://github.com/IHPSystems/ImageAnnotations.jl/commit/e90e51a73a1d7a02e052b3bbd64c8b34f277fcb2

Also, it got merged just in time for a successful weekly integration test run for all Julia 1 minor versions: https://gitlab.com/IHPSystems/julia/julia-gitlab-ci-templates/-/pipelines 🚀 😀

PR #91536, opened 2023-09-16
Limit older versions of ScopedValues to 1.10 and prior

vchuravy:
I hope I got the syntax right :)


DilumAluthge:
Probably what you want to do here is:

1. Use RegistryTools.Compress to uncompress.
2. Make the changes you want on the uncompressed dict.
3. Use RegistryTools.Compress to recompress

PR #91635, opened 2023-09-18
Fix GR_jll Qt6Base_jll compat

barche:


jll/G/GR_jll/Compat.toml
```diff
@@ -22,4 +22,4 @@ julia = "1.6.0-1"
 ["0.72.9-0"]
 FreeType2_jll = "2.10.4-2"
 Libtiff_jll = "4.5.1-4.5"
-Qt6Base_jll = "6.4.2-6"
+Qt6Base_jll = "~6.4.2-6"
```
giordano, line 25: If I understand what you want to do ("restrict compatibility to v6.4.2 and following releases in the 6.4.x series") then 
```suggestion
Qt6Base_jll = "6.4.2-6.4"
```
should do it.  Also, it'd be good as a follow up to make the same change in Yggdrasil, using `[skip build]` in the commit message and `[skip ci]` when merging the PR.

jll/G/GR_jll/Compat.toml
```diff
@@ -22,4 +22,4 @@ julia = "1.6.0-1"
 ["0.72.9-0"]
 FreeType2_jll = "2.10.4-2"
 Libtiff_jll = "4.5.1-4.5"
-Qt6Base_jll = "6.4.2-6"
+Qt6Base_jll = "6.4.2-6.5"
```
giordano, line 25: This include 6.5.x series, is this what you want to do?

barche: No,I misunderstood how to translate the ~, I'll make it 6.4.

PR #91870, opened 2023-09-21
[NLPModelsJuMP] Fix compat with JuMP

abelsiqueira:
NLPModelsJuMP uses internal functionality of JuMP that have been changed in 1.15.0

x-ref: https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/issues/163


N/NLPModelsJuMP/Compat.toml
```diff
@@ -26,13 +26,12 @@ NLPModels = "0.18-0.19"
...
 
 ["0.12-0.12.1"]
-JuMP = "1.2.0-1"
+JuMP = "1.2-1.14.1"
 
 ["0.12.1-0"]
 NLPModels = "0.18-0.20"
 
 ["0.12.2-0"]
-JuMP = "1.2-1.14.1"
```
giordano, line 35: This version doesn't have any compat at all with jump, or am I missing something?

abelsiqueira: No, you're right, I forgot to change the version range in line 28

odow:
@abelsiqueira can this be closed?

abelsiqueira:
Yes, thanks for pointing to it. Closing.

N/NLPModelsJuMP/Compat.toml
abelsiqueira: Just informing, @odow pointed out that we actually don't want to add constraints because we are still working, we just don't support the new features yet.

PR #91940, opened 2023-09-22
New version: Trilinos_jll v14.4.0+0

jlbuild:

* Registering JLL package Trilinos_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/Trilinos_jll.jl
* Version: v14.4.0+0
* Commit: d30cffd1e67d3fed0c8533da55cedccb69fe2cb1
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/838e0a8945c46a5bed59491e300faed4dcbf13a6
* Created by: @Keno


Keno:
Trilinos has a 1.10 julia compact. I'm not 100% sure why, but I think it has something to do with using some blas trampoline things that are new in 1.10. As a result, automerge can't install this. I don't know if we should have bumped the compat to 1.10 before that is released, but in any case, that wasn't new in what I did, so I'm gonna go ahead and merge this for forward progress :).

PR #92580, opened 2023-10-01
Add Pkg package and compat for it to packages

vchuravy:


vchuravy:
Ah this is missing the "Pkg package" part.

A/AbInitioSoftwareBase/Compat.toml
```diff
@@ -6,6 +6,7 @@ julia = "1"
 
 ["0-0.2.0"]
 Configurations = "0.3-0.15"
+Pkg = "1"
```
GunnarFarneback, line 10: I don't know exactly what changes this is supposed to make, but only updating compat for the oldest version seems unreasonable.

GunnarFarneback: Ok thanks. That was the connection I was missing.

KristofferC:
I think LazyArtifacts have to be made upgradable first.

vchuravy:
Yeah maybe it's time to redo #92622

PR #92622, opened 2023-10-02
Update compat for LazyArtifacts

vchuravy:
- Update LazyArtifacts compat
- register new version


vchuravy:
@DilumAluthge My preferred plan of action is to run the CompatStdlibUpdater for all stdlibs and then register the stdlibs one by one.
I think one of the challenges is what do we want to use as the commit for the 1.11.0 registration.

giordano:
I think this should have just used "1" everywhere.

DilumAluthge:
Tbh it probably should have used `<0.0.1, 1` everywhere.

giordano:
There's no version 0.0.0 in the registry, only 1.3 and 1.11

DilumAluthge:
When you do e.g. `Pkg.test()` in Julia 1.3, Pkg will set the version numbers of all stdlibs to 0.0.0.

giordano:
But LazyArtifacts isn't a stdlib in Julia v1.3, is it?

DilumAluthge:
The bug can still show up in later Julia versions. In e.g. Julia 1.8 and 1.9, if you do Pkg operations during your test suite, then you (and all of your recursive dependencies) need to have `<0.0.1` in the stdlib compat.

PR #93084, opened 2023-10-09
[MiniZinc_jll] fix HiGHS_jll compat

odow:
x-ref https://github.com/jump-dev/MiniZinc.jl/issues/48

This should have been 1.5.3. It is fixed in the Project.toml: https://github.com/JuliaBinaryWrappers/MiniZinc_jll.jl/blob/7cc3dd33d76582bc5ffc33b5dbb7ea0c2c7bed6d/Project.toml#L15


StefanKarpinski:
Just to check: it has to be only this specific version?

odow:
Yes, only v1.5.3. It should correspond to `=1.5.3` in the Project.toml

PR #93505, opened 2023-10-16
Fix JLD2 julia compat

IanButterworth:
Avoid
```
ERROR: LoadError: type DataType has no field size
Stacktrace:
 [1] getproperty(x::Type, f::Symbol)
   @ Base ./Base.jl:32
 [2] top-level scope
   @ ~/.julia/packages/JLD2/qncOK/src/data/number_types.jl:14
```


J/JLD2/Compat.toml
```diff
@@ -8,6 +8,9 @@ DataStructures = "0.0.0-0.17"
 ["0-0.4.17"]
 julia = ["0.7", "1"]
```
IanButterworth, line 9:
```suggestion
["0-0.4.27"]
julia = ["0.7", "1-1.8"]
```

J/JLD2/Compat.toml
```diff
@@ -8,6 +8,9 @@ DataStructures = "0.0.0-0.17"
...
 
+["0.4.28"]
+julia = ["1.9"]
```
IanButterworth, line 12:
```suggestion
julia = ["1.9-1"]
```

giordano:
For context, JLD2 used to access internal fields of Julia's datastructures, which was broken in version v1.9: https://github.com/JuliaIO/JLD2.jl/issues/428.  This was fixed by https://github.com/JuliaIO/JLD2.jl/pull/438, included in JLD2 v0.4.28+, but until now older versions are still allowed to be installed on Julia v1.9+, so this PR makes it impossible to install JLD2 v0.4.27- on Julia v1.9+, adding a compatibility cap to julia v1.8.

PR #93874, opened 2023-10-22
URL change request for JuliaWGPU repositories

arhik:
moving from `dvijaha` to `JuliaWGPU` organization


arhik:
All checks passed. Anything else I should be doing here ? Should I reach out to someone ?

StefanKarpinski:
Looks good. Verified that @arhik is the only public member of https://github.com/dvijaha org.

PR #94138, opened 2023-10-26
Allow RegistryCI v9

DilumAluthge:
I want to deploy https://github.com/JuliaRegistries/General/pull/94135 (which includes https://github.com/JuliaRegistries/RegistryCI.jl/pull/516). Since v9 is a breaking change, I need to first update the compat entry.

This PR does not actually deploy RegistryCI v9. That will happen in a follow-up PR to update the manifests.


DilumAluthge:
Closing in favor of https://github.com/JuliaRegistries/General/pull/94141

PR #94199, opened 2023-10-27
Add mention of Aqua and JET

gdalle:
Point users to Aqua and JET in the section about tests


PallHaraldsson:
I just learned of: https://github.com/crate-ci/typos

that has added supported, and is now used by Julia CI, so maybe point it out to users, if some user friendly way to use it:
https://github.com/JuliaLang/julia/pull/51704

PallHaraldsson:
Does it help to add: "It's helpful if your code is type-stable; to use `@code_warntype` for your code to ensure (for speed-critical code, most code isn't and it's not strictly needed for such code).

I think this advice may be redundant with either or both of the suggested packages. With the macro you at least don't need to install a package, but also while they automate, I think they look at all your code, and people may get overwhelmed. So while helpful, not all code needs to be perfect/type-stable, and you may know where best to check. Type-unstable isn't wrong, so maybe imperfect isn't the right word for it?

[I recently saw a user on discource wanting his code type-stable (non-allocating), it strictly wasn't needed, but he wanted his code-dependency to be type-stable, since his dependent was, to not make it no longer so. Maybe packages that are, should be marked, would that be possible? To be useful, dependencies would need to be recursively checked...]

gdalle:
I don't think expecting type-stability from registered packages is quite in the same category as good tests and documentation. But we can discuss this on another issue, cause apparently my one sentence was deemed mergeable ;)

PR #94242, opened 2023-10-27
yank AdvancedMH.jl 0.7.6

torfjelde:
Yank AdvancedMH.jl as it was mistakenly released and contains breaking changes.


devmotion:
FYI I don't think this problematic release can be fixed by reverting some changes and making another non-breaking release: the problematic release added AbstractMCMC 5 to the compat section but AbstractMCMC 5 support is broken and the package can't be updated to AbstractMCMC 5 without breaking changes.

PR #95271, opened 2023-11-13
Update URL in Package.toml of TensorGames

forrestlaine:
Repository moved due to changing account name.


lgoettgens:
I think you will need to ping a human to look at this. It's best to text on slack's `#pkg-registration` channel. [noblock]

PR #95368, opened 2023-11-15
Create EEEAntennaLab.jl

mtamer00:


lgoettgens:
@mtamer00 This is not how the general registry works. Please have a look at the README.md for an explanation on how to use the general registry and how to register new packages.

PR #95448, opened 2023-11-16
Set LazyArtifactsCompat to 1

vchuravy:
Ref: https://github.com/JuliaRegistries/General/pull/92622#issuecomment-1813255248


giordano:
I tried to compress the compat bounds with
```julia
julia> for (root, dirs, files) in walkdir(".") # "." was the top directory of the repository
           if "Compat.toml" in files
               compat = joinpath(root, "Compat.toml")
               RegistryTools.Compress.save(compat, RegistryTools.Compress.load(compat))
           end
       end
```
I have no idea whether it was right, I had to literally read the source code of https://github.com/JuliaRegistries/RegistryTools.jl/blob/77e2a02e62185ce865653bdae95203a3a40510f0/src/Compress.jl because `RegistryTools.jl` comes with zero documentation: https://github.com/JuliaRegistries/RegistryTools.jl/issues/94

PR #95815, opened 2023-11-22
Fix compat in ClimaCoreTempestRemap

Sbozzolo:
ClimaCore 0.11 introduced a braking change that broke ClimaCoreTempestRemap. The correct compat is `=0.11`.


juliasloan25 (approved):
thank you!

PR #98057, opened 2024-01-02
CI: run AutoMerge on Julia 1.10

DilumAluthge:


.github/workflows/automerge.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.9'
+          - '1.10'
```
giordano, line 22: Can we change the name of the CI jobs so that it doesn't include this version number?  Otherwise everytime we upgrade to a newer version of Julia all past CI jobs will be stuck

DilumAluthge: This is for the AutoMerge job, not the CI job. We don't have the AutoMerge job marked as a required status check (this is intentional; we want to let registry maintainers do manual merges even if AutoMerge is red).

Because the AutoMerge job isn't required, if we changed the name of the job, it won't block already open PRs.

PR #98171, opened 2024-01-04
Yank LoopVectorization versions with compat change

staticfloat:
These changes should have been done with a major version bump


staticfloat:
Chris Elrod OK'ed this: https://github.com/JuliaSIMD/LoopVectorization.jl/issues/520#issuecomment-1877626883

PR #98815, opened 2024-01-13
CI: Add workflow for author approved label

ericphanson:
This intends to facilitate AutoMerge reading this label and bypassing some checks if the author approves. The intended use is:
1. AutoMerge would have guidelines that can be bypassed if the author approves. Without approval, these fail and the automerge comment instructs the author (well: whoever registered the version/package) to comment with "approved".
2. If the author does so, that triggers this workflow, which adds the "author-approved" label to the PR
3. The AutoMerge workflow would be updated to retrigger when this label is applied. It would run through the same logic, but this time notice the label, and use it to bypass those checks.

Currently this workflow is useless as AutoMerge does not have such conditional checks, but I wanted to put a PR to try it out.

Note that https://github.com/JuliaRegistries/RegistryCI.jl/pull/536 shows how easy it is to use label information within automerge itself, which I think makes this a viable route.

For the sake of testing, I will add text to trigger the regex:

- Created by: @ericphanson

### DIFF
--- a/.github/workflows/author_approval.yml
+++ b/.github/workflows/author_approval.yml
@@ -6,11 +6,14 @@ on:
 jobs:
   label:
     permissions:
+        # We need `write` permissions on `pull-requests` in order to be able to
+        # add/remove labels from PRs. As far as we can tell, there is no narrower
+        # permission that we can use.
         pull-requests: write
     runs-on: ubuntu-latest
-    # Run on comments that are:
+    # Run on comments that are satisfy all of the following:
     # 1) on PRs, not issues,
-    # 2) not from bot users and,
+    # 2) not from bot users
     # 3) include the string "merge approved"
     # If so, we will do the work to check that the commenter is the package author,
     # and conditionally apply the author-approved label.
@@ -26,7 +29,7 @@ jobs:
           COMMENTER: ${{ github.event.comment.user.login }}
         shell: julia --compile=min --optimize=0 --color=yes {0}
         run: |
-          m = match(r"Created by: @([^\s]+)", ENV["PR_BODY"])
+          m = match(r"Created by: @([A-Za-z0-9]*+)(?:$|\n)", ENV["PR_BODY"])
           verified = !isnothing(m) && m[1] == ENV["COMMENTER"]
           println("Matched user: ", m === nothing ? nothing : m[1])
           println("Commenter: ", ENV["COMMENTER"])
@@ -38,6 +41,9 @@ jobs:
         if: ${{ steps.verify-author.outputs.verified == 'true' }}
         env:
           PR_NUM: ${{ github.event.issue.number }}
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # We cannot use `${{ secrets.GITHUB_TOKEN }}` here, because
+          # if we use `GITHUB_TOKEN` here, then the "label created" event
+          # will not trigger any further GitHub Actions.
+          GH_TOKEN: ${{ secrets.TAGBOT_TOKEN }}
           GH_REPO: ${{ github.repository }}
         run: gh pr edit "$PR_NUM" --add-label "package-author-approved"
--- a/.github/workflows/author_approval.yml
+++ b/.github/workflows/author_approval.yml
@@ -11,11 +11,11 @@ jobs:
     # Run on comments that are:
     # 1) on PRs, not issues,
     # 2) not from bot users and,
-    # 3) include the string "approved"
+    # 3) include the string "merge approved"
     # If so, we will do the work to check that the commenter is the package author,
     # and conditionally apply the author-approved label.
-    # note: `approved` here is case-sensitive 
-    if: ${{ github.event.issue.pull_request && github.event.issue.user.type != 'Bot' && contains(github.event.comment.body, 'approved') }}
+    # note: `merge approved` here is NOT case-sensitive, see https://docs.github.com/en/actions/learn-github-actions/expressions#contains
+    if: ${{ github.event.issue.pull_request && github.event.issue.user.type != 'Bot' && contains(github.event.comment.body, 'merge approved') }}
     steps:
       - name: Verify package author
         id: verify-author
--- a/.github/workflows/author_approval.yml
+++ b/.github/workflows/author_approval.yml
@@ -46,4 +46,4 @@ jobs:
           # will not trigger any further GitHub Actions.
           GH_TOKEN: ${{ secrets.TAGBOT_TOKEN }}
           GH_REPO: ${{ github.repository }}
-        run: gh pr edit "$PR_NUM" --add-label "Override AutoMerge: package author approved"
+        run: gh pr edit "${PR_NUM:?}" --add-label "Override AutoMerge: package author approved"
--- a/.github/workflows/author_approval.yml
+++ b/.github/workflows/author_approval.yml
@@ -14,11 +14,12 @@ jobs:
     # Run on comments that are satisfy all of the following:
     # 1) on PRs, not issues,
     # 2) not from bot users
-    # 3) include the string "merge approved"
+    # 3) include the string "[merge approved]"
+    # 4) include the string "[noblock]"
     # If so, we will do the work to check that the commenter is the package author,
     # and conditionally apply the author-approved label.
-    # note: `merge approved` here is NOT case-sensitive, see https://docs.github.com/en/actions/learn-github-actions/expressions#contains
-    if: ${{ github.event.issue.pull_request && github.event.issue.user.type != 'Bot' && contains(github.event.comment.body, 'merge approved') }}
+    # note: `[merge approved]` here is NOT case-sensitive, see https://docs.github.com/en/actions/learn-github-actions/expressions#contains
+    if: ${{ github.event.issue.pull_request && github.event.issue.user.type != 'Bot' && contains(github.event.comment.body, '[merge approved]') }}
     steps:
       - name: Verify package author
         id: verify-author
@@ -46,4 +47,4 @@ jobs:
           # will not trigger any further GitHub Actions.
           GH_TOKEN: ${{ secrets.TAGBOT_TOKEN }}
           GH_REPO: ${{ github.repository }}
-        run: gh pr edit "$PR_NUM" --add-label "package-author-approved"
+        run: gh pr edit "$PR_NUM" --add-label "Override AutoMerge: package author approved"
--- a/.github/workflows/author_approval.yml
+++ b/.github/workflows/author_approval.yml
@@ -15,7 +15,6 @@ jobs:
     # 1) on PRs, not issues,
     # 2) not from bot users
     # 3) include the string "[merge approved]"
-    # 4) include the string "[noblock]"
     # If so, we will do the work to check that the commenter is the package author,
     # and conditionally apply the author-approved label.
     # note: `[merge approved]` here is NOT case-sensitive, see https://docs.github.com/en/actions/learn-github-actions/expressions#contains
--- a/.github/workflows/author_approval.yml
+++ b/.github/workflows/author_approval.yml
@@ -1,0 +1,33 @@
+name: Author Approval Label
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    # Run on comments that are:
+    # 1) on PRs, not issues,
+    # 2) not from bot users and,
+    # 3) include the string "approved"
+    # If so, we will do the work to check that the commenter is the package author,
+    # and conditionally apply the author-approved label.
+    if: ${{ github.event.issue.pull_request && github.event.issue.user.type != 'Bot' && contains(github.event.comment.body, 'approved') }}
+    steps:
+      - name: Verify package author
+        id: verify-author
+        env:
+          PR_BODY: ${{ github.event.issue.body }}
+          COMMENTER: ${{ github.event.comment.user.login }}
+        shell: julia --color=yes {0}
+        run: |
+          m = match(r"Created by: @([^\s]+)", ENV["PR_BODY"])
+          verified = !isnothing(m) && m[1] == ENV["COMMENTER"]
+          open(ENV["GITHUB_OUTPUT"], "a") do io
+            println(io, "verified=$verified")
+          end
+      - name: Add label
+        if: ${{ steps.verify-author.outputs.verified == 'true' }}
+        env:
+          PR_NUM: ${{ github.event.issue.number }}
+        run: gh pr edit "$PR_NUM" --add-label "author-approved"
--- a/.github/workflows/author_approval.yml
+++ b/.github/workflows/author_approval.yml
@@ -30,4 +30,6 @@ jobs:
         if: ${{ steps.verify-author.outputs.verified == 'true' }}
         env:
           PR_NUM: ${{ github.event.issue.number }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
         run: gh pr edit "$PR_NUM" --add-label "author-approved"
--- a/.github/workflows/author_approval.yml
+++ b/.github/workflows/author_approval.yml
@@ -17,6 +17,8 @@ jobs:
       - name: Verify package author
         id: verify-author
         env:
+          # We use an env variable, not direct interpolation into the script, for security:
+          # https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable
           PR_BODY: ${{ github.event.issue.body }}
           COMMENTER: ${{ github.event.comment.user.login }}
         shell: julia --color=yes {0}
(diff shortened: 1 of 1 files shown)


ericphanson:
approved

ericphanson:
Ah, I can't test this here, since it's an [`issue_comment` triggered workflow](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment):

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

I suppose if we merge this, I can test it in a separate PR that could presumably update the workflow and it would get the changes from the branch instead of from main?

DilumAluthge:
> I suppose if we merge this, I can test it in a separate PR that could presumably update the workflow and it would get the changes from the branch instead of from main?

I think, for issue comments and PR comments, it always picks up the workflow file from the default branch :(

We could make a fork of General and do some testing there?

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,35 @@
...
+    # 2) not from bot users and,
+    # 3) include the string "approved"
+    # If so, we will do the work to check that the commenter is the package author,
+    # and conditionally apply the author-approved label.
+    if: ${{ github.event.issue.pull_request && github.event.issue.user.type != 'Bot' && contains(github.event.comment.body, 'approved') }}
+    steps:
+      - name: Verify package author
+        id: verify-author
+        env:
+          PR_BODY: ${{ github.event.issue.body }}
```
DilumAluthge, line 20: It might be good to add a comment here noting that this method (using intermediate environment variables) is a security recommendation from GitHub: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,35 @@
...
+          COMMENTER: ${{ github.event.comment.user.login }}
+        shell: julia --color=yes {0}
+        run: |
+          m = match(r"Created by: @([^\s]+)", ENV["PR_BODY"])
```
DilumAluthge, line 24: Do we need to enforce that at the end of the `@username`, we have either `\n` or `$`?

Otherwise, wouldn't an approval by user `@foo` be able to approve a PR created by user `@foobar`?

DilumAluthge:
I wonder if we could come up with a different name other than "author"? In the AutoMerge codebase, we always use "author" to refer to the actual GitHub PR author, so `@JuliaRegistrator`, `@jlbuild`, etc.

So it might be a little confusing to also use the same term "author" for the human that triggered Registrator.

Maybe "registration-trigger-user` or something like that?

.github/workflows/author_approval.yml
ericphanson: I think it's fine actually:
```julia
julia> r = r"Created by: @([^\s]+)"
r"Created by: @([^\s]+)"

julia> match(r, "Created by: @foobar")[1]
"foobar"

julia> match(r, "Created by: @foo")[1]
"foo"
```
i.e. we will match the whole username, whatever it is, and then we string-compare (not regex) against the commenter

ericphanson:
For the label? I think we should make sure it's understandable by the user, so they can know that their approval was picked up correctly. So I think it shouldn't be too obscure. What about `package-author-approved`?

.github/workflows/author_approval.yml
ericphanson:
```suggestion
          # We use an env variable, not direct interpolation, for security:
          # https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable
          PR_BODY: ${{ github.event.issue.body }}
```

ericphanson:
Ok, I've tested it [here](https://github.com/ericphanson/PkgA/pull/3), and updated this PR.

Mismatch:
Screenshot 2024-01-14 at 23 48 40
Match:
Screenshot 2024-01-14 at 23 48 40

Note the times on the Julia part vary from 1-2s up to 15-20s, I'm not really sure why. However that code only runs when a comment is posted from a non-bot user that includes "approved", so hopefully it doesn't need to run too much.

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,42 @@
+name: Author Approval Label
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  label:
+    permissions:
+        pull-requests: write
```
DilumAluthge, line 9: Is there any narrower permission that we can give, other than `pull-requests: write`?

ericphanson: I don't think there are narrower permissions: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

ericphanson: note however the token itself is only available to the final step of the workflow, which does not have any user-supplied inputs, so should be pretty secure

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: ubuntu-latest
+    # Run on comments that are:
+    # 1) on PRs, not issues,
+    # 2) not from bot users and,
+    # 3) include the string "approved"
+    # If so, we will do the work to check that the commenter is the package author,
+    # and conditionally apply the author-approved label.
+    if: ${{ github.event.issue.pull_request && github.event.issue.user.type != 'Bot' && contains(github.event.comment.body, 'approved') }}
```
DilumAluthge, line 17: The word `approved` will need to appear with exact case matching (case sensitive) in the document, right? Should we document this somewhere?

ericphanson: Yep, I think AutoMerge should mention it any time there’s a failure due to lack of approval (none exist yet). Perhaps once we have some we can add a faq entry as well. I don’t think it makes sense to add public docs with this PR as the mechanism isn’t used for anything yet.

DilumAluthge: For now can you stick a comment in the YAML (right above this line, maybe) just pointing out that it's case sensitive?

DilumAluthge:
Can we also exclude the `github-actions[bot]` user?

ericphanson:
I have excluded all bot users

.github/workflows/author_approval.yml
ericphanson:
```suggestion
    # note: `approved` here is case-sensitive 
    if: ${{ github.event.issue.pull_request && github.event.issue.user.type != 'Bot' && contains(github.event.comment.body, 'approved') }}
```

DilumAluthge:
> For the label? I think we should make sure it's understandable by the user, so they can know that their approval was picked up correctly. So I think it shouldn't be too obscure. What about `package-author-approved`?

Yeah let's go with that.

ericphanson:
I had already renamed the label to `package-author-approved`. Just now, I changed the approval string to "merge approved", and realized in fact it is not case-sensitive, as `contains` is not. I updated the comment with a reference.

I think this is good to go, and am writing a PR to RegistryCI to make use of this in the sequential version skipping check (probably to the delight of like 2 people who like skipping versions). We can also use it for other things down the line.

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Verify package author
+        id: verify-author
+        env:
+          # We use an env variable, not direct interpolation into the script, for security:
+          # https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable
+          PR_BODY: ${{ github.event.issue.body }}
+          COMMENTER: ${{ github.event.comment.user.login }}
+        shell: julia --compile=min --optimize=0 --color=yes {0}
+        run: |
+          m = match(r"Created by: @([^\s]+)", ENV["PR_BODY"])
```
DilumAluthge, line 29: Per the GitHub https://docs.github.com/en/enterprise-cloud@latest/admin/identity-and-access-management/iam-configuration-reference/username-considerations-for-external-authentication:

> Usernames for user accounts on GitHub.com can only contain alphanumeric characters and dashes (-).

So our regex only needs to check for alphanumeric characters (uppercase letters, lowercase letters, and numbers) and dashes.

Also, we can enforce that the "Created by" line either ends in a newline, or the end of the string.

```suggestion
          m = match(r"Created by: @([A-Za-z0-9]*+)(?:$|\n)", ENV["PR_BODY"])
```

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,43 @@
...
+          println("Commenter: ", ENV["COMMENTER"])
+          println("Verified: ", verified)
+          open(ENV["GITHUB_OUTPUT"], "a") do io
+            println(io, "verified=$verified")
+          end
+      - name: Add label
+        if: ${{ steps.verify-author.outputs.verified == 'true' }}
+        env:
+          PR_NUM: ${{ github.event.issue.number }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
DilumAluthge, line 41:
```suggestion
          # We cannot use `${{ secrets.GITHUB_TOKEN }}` here, because
          # if we use `GITHUB_TOKEN` here, then the "label created" event
          # will not trigger any further GitHub Actions.
          GH_TOKEN: ${{ secrets.TAGBOT_TOKEN }}
```

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  label:
+    permissions:
+        pull-requests: write
+    runs-on: ubuntu-latest
+    # Run on comments that are:
+    # 1) on PRs, not issues,
+    # 2) not from bot users and,
+    # 3) include the string "merge approved"
```
DilumAluthge, line 14:
```suggestion
    # 3) include the string "[merge approved]"
    # 4) include the string "[noblock]"
```

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,43 @@
...
+    # If so, we will do the work to check that the commenter is the package author,
+    # and conditionally apply the author-approved label.
+    # note: `merge approved` here is NOT case-sensitive, see https://docs.github.com/en/actions/learn-github-actions/expressions#contains
+    if: ${{ github.event.issue.pull_request && github.event.issue.user.type != 'Bot' && contains(github.event.comment.body, 'merge approved') }}
```
DilumAluthge, line 18:
```suggestion
    if: github.event.issue.pull_request && (github.event.issue.user.type != 'Bot') && (github.event.comment.body == '[merge approved]') && (github.event.comment.body == '[noblock]')
```

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,43 @@
+name: Author Approval Label
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  label:
+    permissions:
+        pull-requests: write
```
DilumAluthge, line 9:
```suggestion
        # We need `write` permissions on `pull-requests` in order to be able to
        # add/remove labels from PRs. As far as we can tell, there is no narrower
        # permission that we can use.
        pull-requests: write
```

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+    # Run on comments that are:
```
DilumAluthge, line 11:
```suggestion
    # Run on comments that are satisfy all of the following:
```

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,43 @@
...
+    # 1) on PRs, not issues,
+    # 2) not from bot users and,
```
DilumAluthge, line 13:
```suggestion
    # 2) not from bot users
```

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,43 @@
...
+    # 3) include the string "merge approved"
+    # If so, we will do the work to check that the commenter is the package author,
+    # and conditionally apply the author-approved label.
+    # note: `merge approved` here is NOT case-sensitive, see https://docs.github.com/en/actions/learn-github-actions/expressions#contains
```
DilumAluthge, line 17:
```suggestion
    # note: `[merge approved]` here is NOT case-sensitive, see https://docs.github.com/en/actions/learn-github-actions/expressions#contains
```

DilumAluthge (changes requested):

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,43 @@
...
+          open(ENV["GITHUB_OUTPUT"], "a") do io
+            println(io, "verified=$verified")
+          end
+      - name: Add label
+        if: ${{ steps.verify-author.outputs.verified == 'true' }}
+        env:
+          PR_NUM: ${{ github.event.issue.number }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: gh pr edit "$PR_NUM" --add-label "package-author-approved"
```
DilumAluthge, line 43: Can we rename the label from `package-author-approved` to `Override AutoMerge:package-author-approved`? That way, we'll have the convention that if a label is able to affect AutoMerge, then it starts with `Override AutoMerge:`?

DilumAluthge (changes requested):

.github/workflows/author_approval.yml
ericphanson: This looks much more complicated, and I’m sure users will make mistakes. I’d much rather what I wrote and to change the RegistryCI check to not block on this phrase (which is already done in my PR)

DilumAluthge: I don't love the special-casing in https://github.com/JuliaRegistries/RegistryCI.jl/pull/539.

It also seems simpler to me to tell users "always include `[noblock] in your comments", instead of saying "sometimes you need `[noblock]` in your comments, and other times you don't".

Can we just add text to the AutoMerge comment of the following form?

> If you are the author of this package, and you want to approve this PR and bypass the sequential-version-number check, please post a comment on this PR with the following text verbatim: `[merge approved] [noblock]`

.github/workflows/author_approval.yml
ericphanson: What about `Override AutoMerge: package author approved`? That matches better the other one, `Override AutoMerge: name similarity is okay`

.github/workflows/author_approval.yml
ericphanson: I think a longer comment means more chance to miss that text along with all the rest of it. It's already so long!

As a compromise, what about `[merge approved]` (with the square brackets), but without requiring `[noblock]`? Then the rule is "commands to AutoMerge occur in square brackets and comments with commands do not block automerge".

.github/workflows/author_approval.yml
DilumAluthge: SGTM.

.github/workflows/author_approval.yml
DilumAluthge: I think I can live with that.

.github/workflows/author_approval.yml
ericphanson:
```suggestion
        run: gh pr edit "$PR_NUM" --add-label "Override AutoMerge: package author approved"
```

.github/workflows/author_approval.yml
ericphanson:
```suggestion
    if: ${{ github.event.issue.pull_request && github.event.issue.user.type != 'Bot' && contains(github.event.comment.body, '[merge approved]') }}
```

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,50 @@
...
+        # We need `write` permissions on `pull-requests` in order to be able to
+        # add/remove labels from PRs. As far as we can tell, there is no narrower
+        # permission that we can use.
+        pull-requests: write
+    runs-on: ubuntu-latest
+    # Run on comments that are satisfy all of the following:
+    # 1) on PRs, not issues,
+    # 2) not from bot users
+    # 3) include the string "[merge approved]"
+    # 4) include the string "[noblock]"
```
ericphanson, line 18:
```suggestion
```

.github/workflows/author_approval.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Add label
+        if: ${{ steps.verify-author.outputs.verified == 'true' }}
+        env:
+          PR_NUM: ${{ github.event.issue.number }}
+          # We cannot use `${{ secrets.GITHUB_TOKEN }}` here, because
+          # if we use `GITHUB_TOKEN` here, then the "label created" event
+          # will not trigger any further GitHub Actions.
+          GH_TOKEN: ${{ secrets.TAGBOT_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: gh pr edit "$PR_NUM" --add-label "Override AutoMerge: package author approved"
```
DilumAluthge, line 49:
```suggestion
        run: gh pr edit "${PR_NUM:?}" --add-label "Override AutoMerge: package author approved"
```

DilumAluthge (changes requested):

DilumAluthge (approved):

PR #98962, opened 2024-01-16
Require Julia 1.9 for existing versions of StableHashTraits.jl

haberdashPI:
I have had to make a fix to make StableHashTraits to ensure stable hashes on Julia 1.10: (c.f. https://github.com/beacon-biosignals/StableHashTraits.jl/pull/43)

I'm wondering if it's possible to mark the existing versions as incompatible with 1.10, and (ideally, almost immediately thereafter) merge [#43](https://github.com/beacon-biosignals/StableHashTraits.jl/pull/43) and release a version that correctly supports 1.10.

Note that #43 is my short term solution to avoid making existing hash versions unusable on 1.10, and I have a longer term fix at [#46](https://github.com/beacon-biosignals/StableHashTraits.jl/pull/46)


GunnarFarneback:
As noted in a slack discussion, the compat syntax is wrong here and needs to be fixed.

staticfloat:
This has broken the ability to do things like install `Makie` on v1.10:

```
(@v1.10) pkg> add Makie@0.20
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package StableHashTraits [c5dd0088]:
 StableHashTraits [c5dd0088] log:
 ├─possible versions are: 0.1.1-1.1.4 or uninstalled
 ├─restricted by julia compatibility requirements to versions: uninstalled
 └─restricted by compatibility requirements with Makie [ee78f7c6] to versions: 1.1.0-1.1.4 — no versions left
   └─Makie [ee78f7c6] log:
     ├─possible versions are: 0.9.0-0.20.4 or uninstalled
     └─restricted to versions 0.20 by an explicit requirement, leaving only versions: 0.20.0-0.20.4
```

This should be reverted until a version that is installable on v1.10 is available.

giordano:
My understanding is that merging https://github.com/beacon-biosignals/StableHashTraits.jl/pull/43 would be the real solution, if the package is actually broken on v1.10 I'm not sure reverting this PR is helpful

staticfloat:
Yes, but it's bad to prevent users from instantiating Makie when it was working just fine (just with some cache misses, apparently).

giordano:
@haberdashPI the severity of the change wasn't well explained ("some cache misses" doesn't sound bad?), and most importantly the impact of the change (making Makie uninstallable) was not outlined at all.  It'd be much more helpful to present the context better next time.

staticfloat:
To be fair, it's possible that there are worse consequences for other packages, the only one that I know about is Makie.  But that is a popular enough package that I still think breaking it is unacceptable.

Lessons learned; if the current-stable release of Julia used to be able to install something but you need it to install something else, wait until that something else is available before pulling out support to install the previous thing.

PR #99111, opened 2024-01-18
update compat bounds for StableHashTraits

haberdashPI:
Now that https://github.com/beacon-biosignals/StableHashTraits.jl/pull/43 has merged I'd like to update the compat  bounds for StableHashTraits so that only the most recent version (1.1.5) and later can be installed on 1.10.

I've verified that this syntax seems to work as intended locally (by trying to add StableHashTraits at various versions in juliai 1.10 and julia 1.9).


S/StableHashTraits/Compat.toml
```diff
@@ -19,3 +21,4 @@ SHA = ["0.7", "1"]
 
 ["1.1.5-1"]
 PikaParser = "0.6"
+julia = "1.6.0-1.10"
```
giordano, line 24: Why you don't allow 1.11? This would break makie again for people using Julia nightly (like @staticfloat I believe)

staticfloat, line 24:
```suggestion
julia = "1.6.0-1"
```

haberdashPI: Ah okay, yes, I see. Makes sense.

PR #99577, opened 2024-01-25
Yank PETSc_jll 3.18.7+1

imciner2:
This build removed SuiteSparse support, which ended up breaking dependent packages because they had been built expecting SuiteSparse libraries available. This has been rebuilt as a new 3.18.8 version, so yank the broken one. See https://github.com/JuliaPackaging/Yggdrasil/pull/8004.

CC @boriskaus


boriskaus:
thanks!

PR #99725, opened 2024-01-28
[GTK4_jll] cap compat to gdk_pixbuf_jll v2.42.8

jwahlstrand:
Add a cap on gdk_pixbuf_jll compatibility to older versions of GTK4_jll. An upgrade of Libtiff_jll required similar caps on other packages: https://github.com/JuliaRegistries/General/pull/86099, and then an upgrade of gdk_pixbuf_jll: https://github.com/JuliaPackaging/Yggdrasil/pull/7996 uncovered a compatibility problem with GTK4_jll.

GTK4_jll didn't have Libtiff listed as a dependency before, which is why it was not capped in https://github.com/JuliaRegistries/General/pull/86099. This has now been fixed: https://github.com/JuliaPackaging/Yggdrasil/pull/8033 but we need to add this cap to previous versions so they will continue to work.

### DIFF
--- a/jll/G/GTK4_jll/Compat.toml
+++ b/jll/G/GTK4_jll/Compat.toml
@@ -10,6 +10,9 @@ Wayland_protocols_jll = "1.23.0-1"
 ["4-4.8"]
 Glib_jll = "2.68.3-2"
 
+["4-4.12.4"]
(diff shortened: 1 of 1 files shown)


jwahlstrand:
Sorry for the noise, this doesn't seem to be necessary -- I was confused.

PR #99772, opened 2024-01-29
ChunkSplitters.jl: transfer package to JuliaFolds2 organization

lmiq:


DilumAluthge (approved):
I verified that the old URL automatically redirects to the new URL.

PR #99796, opened 2024-01-29
CI: Use more permissive regex in `author_approval.yml`

ericphanson:
This is the one I tested in https://github.com/ericphanson/PkgA/pull/3 and works well. I don't think we need it to be super strict, because the workflow will not add the label unless the parsed name actually matches the commenter's name.

Note there are old github names like https://github.com/artur- which are valid but many regexes won't match (bc names like that aren't allowed in new accounts).

Created by: @ericphanson

### DIFF
--- a/.github/workflows/author_approval.yml
+++ b/.github/workflows/author_approval.yml
@@ -29,6 +29,10 @@ jobs:
           COMMENTER: ${{ github.event.comment.user.login }}
         shell: julia --compile=min --optimize=0 --color=yes {0}
         run: |
+          isempty(ENV["COMMENTER"]) && error("COMMENTER is empty")
+          if length(ENV["COMMENTER"]) != length(strip(ENV["COMMENTER"]))
+               error("COMMENTER has leading or lagging whitespace")
+          end
           m = match(r"Created by: @(\S+)", ENV["PR_BODY"])
           verified = !isnothing(m) && m[1] == ENV["COMMENTER"]
           println("Matched user: ", m === nothing ? nothing : m[1])
--- a/.github/workflows/author_approval.yml
+++ b/.github/workflows/author_approval.yml
@@ -29,7 +29,7 @@ jobs:
           COMMENTER: ${{ github.event.comment.user.login }}
         shell: julia --compile=min --optimize=0 --color=yes {0}
         run: |
-          m = match(r"Created by: @([^\s]+)", ENV["PR_BODY"])
+          m = match(r"Created by: @(\S+)", ENV["PR_BODY"])
           verified = !isnothing(m) && m[1] == ENV["COMMENTER"]
           println("Matched user: ", m === nothing ? nothing : m[1])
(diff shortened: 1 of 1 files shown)


ericphanson:
[merge approved]

ericphanson:
ah right, we can't test it on the PR itself, since it uses the version off master

DilumAluthge:
Can you add the checks for:
1. COMMENTER is not empty
2. COMMENTER has no leading or lagging whitespace.

.github/workflows/author_approval.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           COMMENTER: ${{ github.event.comment.user.login }}
         shell: julia --compile=min --optimize=0 --color=yes {0}
         run: |
-          m = match(r"Created by: @([A-Za-z0-9]*+)(?:$|\n)", ENV["PR_BODY"])
+          m = match(r"Created by: @([^\s]+)", ENV["PR_BODY"])
```
DilumAluthge, line 32: We don't need the `^`, do we?
```suggestion
          m = match(r"Created by: @([\s]+)", ENV["PR_BODY"])
```

DilumAluthge: Disregard previous comment.

DilumAluthge:
What's the difference between `[^\s]` and `[\S]`?

.github/workflows/author_approval.yml
DilumAluthge, line 32: `[^\s]` should be equivalent to `[\S]`, right?

```suggestion
          m = match(r"Created by: @([\S]+)", ENV["PR_BODY"])
```

ericphanson: yeah, true! And then we can drop the `[]`:
```suggestion
          m = match(r"Created by: @(\S+)", ENV["PR_BODY"])
```
e.g.
```julia
julia> match(r"Created by: @(\S+)", "Created by: @hello")
RegexMatch("Created by: @hello", 1="hello")
```

DilumAluthge (approved):

.github/workflows/author_approval.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           COMMENTER: ${{ github.event.comment.user.login }}
         shell: julia --compile=min --optimize=0 --color=yes {0}
         run: |
-          m = match(r"Created by: @([A-Za-z0-9]*+)(?:$|\n)", ENV["PR_BODY"])
+          m = match(r"Created by: @(\S+)", ENV["PR_BODY"])
```
ericphanson, line 32:
```suggestion
          isempty(ENV["COMMENTER"]) && error("COMMENTER is empty")
          if length(ENV["COMMENTER"]) != length(strip(ENV["COMMENTER"]))
                        error("COMMENTER has leading or lagging whitespace")
          end
          m = match(r"Created by: @(\S+)", ENV["PR_BODY"])
```

DilumAluthge (approved):

PR #99844, opened 2024-01-30
Rename consistency CI

Seelengrab:
Just having it called `CI` has the potential to be confusing for people not familiar with the CI on General. It can easily happen that people confuse the passing `CI` checks with the requirements for automerging.

I hope nothing breaks because of this though, that'd be a bit ironic.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: CI
+name: Registry Consistency
 on:
   pull_request:
     branches:


giordano:
I agree this creates less confusion.  Maybe rename also the individual job (not just whole workflow) name https://github.com/JuliaRegistries/General/blob/02d1ef6d652a83820c0152003f5b217db929bf92/.github/workflows/ci.yml#L15 to something like "Checks" to remove the last "CI" reference?

DilumAluthge:
> to something like "Checks" to remove the last "CI" reference?

FWIW this part would require changing the branch protection settings, and then we'd have to rebase all existing open PRs so that they pick up the new name.

giordano:
Uhm, ok, let's do that in a separate PR if necessary, this is already an improvement, hopefully.

DilumAluthge (approved):

goerz:
Uhoh, did this break things? https://github.com/JuliaRegistries/General/pull/99899

goerz:
Or is that just a particularly broken submission?

DilumAluthge:
We do not allow packages to have hypens (`-`) in the name.

That being said, we should improve the error message.

goerz:
Yeah, all good, there just needs to be stricter regex for the package name (see the [Slack thread](https://julialang.slack.com/archives/C6M4DQA5P/p1706653270061369)). Sorry for the noise.

PR #99871, opened 2024-01-30
CI: change `Registry Consistency / CI` to `Registry Consistency / check`

DilumAluthge:
Follow-up to #99844, based on Mosè's idea [here](https://github.com/JuliaRegistries/General/pull/99844#issuecomment-1917457626).

It's a little annoying to do (because we have to click the "Update branch" on all currently open PRs), but we only have to do that task once, and it might be worth it to get rid of the `CI` word anywhere in this job name.

### DIFF
--- a/.github/workflows/registry-consistency-ci.yml
+++ b/.github/workflows/registry-consistency-ci.yml
@@ -12,7 +12,7 @@ env:
 permissions:
   contents: read
 jobs:
-  CI:
+  check:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
     strategy:


.github/workflows/registry-consistency-ci.yml
```diff
@@ -12,7 +12,7 @@ env:
 permissions:
   contents: read
 jobs:
-  CI:
+  check:
```
DilumAluthge, line 15: We could pick something else (instead of `check`) if we want. We could do e.g. `consistency`.

ericphanson (approved):
check makes sense to me!

DilumAluthge:
@ericphanson @giordano Shall we move forward with this (and just click the "Update branch" button one time on all open PRs as soon as we merge this PR)?

DilumAluthge:
Let's give this a go.

DilumAluthge:
I've updated the branch protection settings.

PR #99974, opened 2024-01-31
Update URL for Variography.jl

juliohm:


vchuravy:
Moving a repository to an organization is normally something that we can quickly merge. Moving out of an organization is something that needs a bit more information.

What is the reason for the move and is there consensus among the JuliaEarth contributors that this is okay?

vchuravy:
Okay that's reasonable enough. You could in the future archive the repo within the org itself.

PR #99991, opened 2024-02-01
Adjust Onda's compatibility with TimeSpans

ararslan:
[Onda](https://github.com/beacon-biosignals/Onda.jl) currently defines ArrowTypes interoperability for [TimeSpans](https://github.com/beacon-biosignals/TimeSpans.jl) types, which is a holdover from before TimeSpans became its own package. Said interop will be moved to an ArrowTypes extension for TimeSpans and released as TimeSpans 1.1.0. Onda currently declares compatibility with TimeSpans 1, but in actuality it will be incompatible with TimeSpans 1.1 because the duplicated method definitions will break precompilation. To fix this, we can adjust the declared compatibility bound here in the registry from 1 to 1.0, which means that existing Onda versions won't be compatible with TimeSpans 1.1. A subsequent release of Onda will remove the interop and set its minimum TimeSpans version to 1.1.

cc @ericphanson


DilumAluthge:
Seems okay at a glance.

Have you tested out some `Pkg.add`s using this branch of the registry?

ararslan:
> Have you tested out some `Pkg.add`s using this branch of the registry?

Ah, I have not, that's a good idea. Thanks!

ararslan:
Okay, I _think_ this does what I want. I did the following to test:

- Created a separate depot
- Cloned General on this branch into the `registries` subdirectory
- `Pkg.add(; url="https://github.com/beacon-biosignals/TimeSpans.jl", rev="aa/weak-arrow")`
    - Package version is set to 1.1.0 on this branch
    - Includes ArrowTypes as a weak dependency (and direct to maintain Julia 1.6 support)
- `Pkg.add("Onda")`
    - Installs Onda 0.11.0; current is 0.15.5
    - `Pkg.status()` shows that newer versions are available
- `Pkg.update()`
    - No changes made
- Try to add recent Onda explicitly
   ```
   (@v1.10) pkg> add Onda@0.15.5
      Resolving package versions...
   ERROR: Unsatisfiable requirements detected for package TimeSpans [bb34ddd2]:
    TimeSpans [bb34ddd2] log:
    ├─possible versions are: 1.1.0 or uninstalled
    ├─TimeSpans [bb34ddd2] is fixed to version 1.1.0
    └─found to have no compatible versions left with Onda [e853f5be]
      └─Onda [e853f5be] log:
        ├─possible versions are: 0.7.2-0.15.5 or uninstalled
        └─restricted to versions 0.15.5 by an explicit requirement, leaving only versions: 0.15.5
   ```

Notably, Onda 0.11.0 was the last release without a dependency on TimeSpans. With TimeSpans at 1.1.0 and Onda's TimeSpans compatibility capped at 1.0, I believe that's the expected behavior.

Does that all look right to you, @DilumAluthge?

ararslan:
Okie doke, `Pkg.add(; url="https://github.com/beacon-biosignals/Onda.jl", rev="aa/rm-timespans-arrowtypes")`, which sets TimeSpans compatibility to 1.1 and its own version to 0.15.6, works as expected. So I think we're good!

PR #100066, opened 2024-02-01
CI: Add `types` to automerge.yml

ericphanson:
from https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request we need to specify our `types`

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -1,6 +1,7 @@
 name: AutoMerge
 on:
   pull_request:
+    types: ['opened', 'synchronize', 'reopened', 'labeled']
   schedule:
     # The cron job is just a fallback.
     # The "stopwatch" functionality that we have implemented should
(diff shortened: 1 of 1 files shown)


.github/workflows/automerge.yml
```diff
@@ -1,6 +1,7 @@
 name: AutoMerge
 on:
   pull_request:
+    types: [opened, reopened, labeled, synchronize]
```
DilumAluthge, line 4: Can you add a comment linking to the relevant GitHub docs?

DilumAluthge (approved):

.github/workflows/automerge.yml
ericphanson:
```suggestion
    # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
    types: [opened, reopened, labeled, synchronize]
```

DilumAluthge (approved):

PR #100172, opened 2024-02-02
populate tagbot token only when necessary

ericphanson:


DilumAluthge (commented):
Can you also print the value of `NEED_MERGE_TOKEN` to the log, to help with debugging?

DilumAluthge (approved):

.github/workflows/automerge.yml
```diff
@@ -34,6 +40,9 @@ jobs:
         arch:
           - x64
     steps:
+      # For debugging purposes, we print this value first
+      - name: Print env variable NEED_MERGE_TOKEN
+        run: echo "NEED_MERGE_TOKEN=$NEED_MERGE_TOKEN"
```
ericphanson, line 45: I think this should work? Not 100% sure though

ericphanson:
let's try this. We might have to revert if I've horribly messed it up and we stop automerging though

PR #101136, opened 2024-02-18
update package urls from gitlab to github

aplavin:
Registrator refuses to create these PRs automatically, hope everything is correct here...


carstenbauer:
The https://github.com/JuliaAPlavin "organisation" is a bit strange - organisations typically aren't personalized and have multiple owners - but we have many registered packages living under personal accounts, which isn't really different from this.

carstenbauer:
FWIW, I'm fine with merging this.

PR #101718, opened 2024-02-26
Yank KernelAbstractions 0.9.16

vchuravy:
Reverts https://github.com/JuliaGPU/KernelAbstractions.jl/pull/446 since that allowed for miscompilation on the CPU backend and yanks the corresponding release.


vchuravy:
> Just release a new patch version. If you need to roll back, just make a new patch release that is identical to the previous one.

https://github.com/JuliaRegistries/General/pull/101716

> merely buggy. 

I guess that always is up to debate. For me this release was not merely buggy, but should not have been made in the first place.

From my perspective this should have been the easiest yank for Pkg to handle, so I don't fully understand the furor this has caused.

PR #101764, opened 2024-02-27
Clapeyron.jl: restrict compat of Roots for earlier versions

longemen3000:
is this ok to do? i had some problems with a too lax of a compat with Roots.jl v2.0, and when they removed Setfield to use Accessors, earlier versions broke. (this was fixed on 0.5.10 of the library)


longemen3000:
fixes https://github.com/ClapeyronThermo/Clapeyron.jl/issues/257

PR #102617, opened 2024-03-10
yank DimensionalData 0.26.0

rafaqz:
Its not compatible with ealier julia versions


odow:
There's no need to yank patch versions. I think we over-use `yank`.

rafaqz:
No this is the case a yank is needed.

Without it DimensionalData will break JuMP.jl and other use cases on all versions below 1.10.

Because the 0.26.10 patch is to no longer support older versions, but 0.26.0 still does. A yank is the only way to stop upgrading to broken DD on old julia.

PR #102636, opened 2024-03-11
add yanking guidance to the readme faq

IanButterworth:
- I changed the header levels because in the readme the Q-A text formatting is poorly differentiated
- Added guidance on yanking (RFC)

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,12 +283,15 @@ Registry maintainers are discouraged from merging packages they directly *or* in
 ### When is yanking a release appropriate?
 
 Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release entry in `Versions.toml`. This will mean 
-that Pkg will no longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General. Yanking should 
-only be used when either:
-1) adding a standard patch release will not resolve the issue i.e. say `v2.10.0` was released using a feature not on julia `v1.6` but the compat entry for julia
+that Pkg will no longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General. 
+
+The primary reason yanking exists is to make a release impossible to install.
+This is intended to be used for circumstances when having this release installed poses a risk to security or safety.
+It is not intended for avoiding people installing releases with bugs, for that use a patch release -- potentially after doing nothing more than revert. All patch releases prior to the most recent have known bugs by definition.
+
+There is however, a special category of bugged releases that can not be resolved by having a patch release. These also need to be resolved by yanking. That special category is when the compat bounds have been set too wide.    i.e. say `v2.10.0` was released using a feature not on julia `v1.6` but the compat entry for julia
 was not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would not solve the issue as on julia v1.6 Pkg would still
 resolve the broken `v2.10.0`, and as a minor bump, reverting the code changes would not be valid in a patch bump.
-2) there are security issues with the release
 
 If yanking is urgent, open a PR and raise it on the `#pkg-registration` slack channel
 
--- a/README.md
+++ b/README.md
@@ -288,8 +288,9 @@ only be used when either:
 1) adding a standard patch release will not resolve the issue i.e. say `v2.10.0` was released using a feature not on julia `v1.6` but the compat entry for julia
 was not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would not solve the issue as on julia v1.6 Pkg would still
 resolve the broken `v2.10.0`, and as a minor bump, reverting the code changes would not be valid in a patch bump.
-3) there are security issues with the release
+2) there are security issues with the release
 
+If yanking is urgent, open a PR and raise it on the `#pkg-registration` slack channel
 
 ## Registry maintenance
 
--- a/README.md
+++ b/README.md
@@ -282,16 +282,18 @@ Registry maintainers are discouraged from merging packages they directly *or* in
 
 ### When is yanking a release appropriate?
 
-Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release entry in `Versions.toml`. This will mean 
-that Pkg will no longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General. 
+Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release entry in `Versions.toml`. This will mean
+that Pkg will no longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General.
 
-The primary reason yanking exists is to make a release impossible to install.
-This is intended to be used for circumstances when having this release installed poses a risk to security or safety.
-It is not intended for avoiding people installing releases with bugs, for that use a patch release -- potentially after doing nothing more than revert. All patch releases prior to the most recent have known bugs by definition.
+The primary reason yanking exists is to make a release impossible to install. This is intended to be used for circumstances when having this
+release installed poses a risk to security or safety. It is not intended for avoiding people installing releases with bugs, for that use a
+patch release -- potentially after doing nothing more than revert. All patch releases prior to the most recent have known bugs by definition.
 
-There is however, a special category of bugged releases that can not be resolved by having a patch release. These also need to be resolved by yanking. That special category is when the compat bounds have been set too wide.    i.e. say `v2.10.0` was released using a feature not on julia `v1.6` but the compat entry for julia
-was not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would not solve the issue as on julia v1.6 Pkg would still
-resolve the broken `v2.10.0`, and as a minor bump, reverting the code changes would not be valid in a patch bump.
+There is however, a special category of bugged releases that can not be resolved by having a patch release. These also need to be resolved by
+yanking. That special category is when the compat bounds have been set too wide. i.e. say `v2.10.0` was released using a feature not on julia
+`v1.6` but the compat entry for julia was not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would
+not solve the issue as on julia v1.6 Pkg would still resolve the broken `v2.10.0`, and as a minor bump, reverting the code changes would not
+be valid in a patch bump.
 
 If yanking is urgent, open a PR and raise it on the `#pkg-registration` slack channel
 
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Currently the waiting period is as follows:
 
 ## FAQ
 
-#### Do I need to register a package to install it?
+### Do I need to register a package to install it?
 
 No, you can simply do `using Pkg; Pkg.add(url="https://github.com/JuliaLang/Example.jl")`
 or `] add https://github.com/JuliaLang/Example.jl` in the Pkg REPL mode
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -275,12 +274,23 @@ This waiting period allows time for community comment and review, and to provide
...
+### When is yanking a release appropriate?
+
+Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release entry in `Versions.toml`. This will mean 
+that Pkg will no longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General. Yanking should 
+only be used when either:
+1) adding a standard patch release will not resolve the issue i.e. say `v2.10.0` was released using a feature not on julia `v1.6` but the compat entry for julia
+was not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would not solve the issue as on julia v1.6 Pkg would still
+resolve the broken `v2.10.0`, and as a minor bump, reverting the code changes would not be valid in a patch bump.
+3) there are security issues with the release
+
```
IanButterworth, line 292:
```suggestion
2) there are security issues with the release

If yanking is urgent, open a PR and raise it on the `#pkg-registration` slack channel
```

README.md
```diff
@@ -275,12 +274,24 @@ This waiting period allows time for community comment and review, and to provide
...
 
+### When is yanking a release appropriate?
+
+Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release entry in `Versions.toml`. This will mean 
+that Pkg will no longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General. Yanking should 
+only be used when either:
+1) adding a standard patch release will not resolve the issue i.e. say `v2.10.0` was released using a feature not on julia `v1.6` but the compat entry for julia
+was not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would not solve the issue as on julia v1.6 Pkg would still
+resolve the broken `v2.10.0`, and as a minor bump, reverting the code changes would not be valid in a patch bump.
+2) there are security issues with the release
```
oxinabox, line 291: I would list this first.
And I would say more along the lines of:

> The primary reason yanking exists is to make a release impossible to install.
This is intended to be used for circumstances when **having this release installed poses a risk to security or safety**.
It is not intended for avoiding people installing releases with bugs, for that use a patch release -- potentially after doing nothing more than revert . All patch releases prior to the most recent have known bugs by definition.

> There is however, a special category of bugged releases that can not be resolved by having a patch release. These also need to be resolved by yanking. That special category is when the compat bounds have been set too wide. [Insert your example here]

IanButterworth, line 291:
```suggestion
Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release entry in `Versions.toml`. This will mean 
that Pkg will no longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General. 

The primary reason yanking exists is to make a release impossible to install.
This is intended to be used for circumstances when having this release installed poses a risk to security or safety.
It is not intended for avoiding people installing releases with bugs, for that use a patch release -- potentially after doing nothing more than revert. All patch releases prior to the most recent have known bugs by definition.

There is however, a special category of bugged releases that can not be resolved by having a patch release. These also need to be resolved by yanking. That special category is when the compat bounds have been set too wide.    i.e. say `v2.10.0` was released using a feature not on julia `v1.6` but the compat entry for julia
was not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would not solve the issue as on julia v1.6 Pkg would still
resolve the broken `v2.10.0`, and as a minor bump, reverting the code changes would not be valid in a patch bump.
```

IanButterworth: sounds good. updated

README.md
```diff
@@ -275,12 +274,27 @@ This waiting period allows time for community comment and review, and to provide
...
 Early merges should still be discussed on the [#pkg-registration](https://julialang.slack.com/archives/C6M4DQA5P) slack channel before merging.
 Registry maintainers are discouraged from merging packages they directly *or* indirectly maintain as mentioned in [CONTRIBUTING.md](https://github.com/JuliaRegistries/General/blob/master/CONTRIBUTING.md#other-ways-to-help).
 
+### When is yanking a release appropriate?
+
+Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release entry in `Versions.toml`. This will mean 
+that Pkg will no longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General. 
+
+The primary reason yanking exists is to make a release impossible to install.
+This is intended to be used for circumstances when having this release installed poses a risk to security or safety.
```
vchuravy: How is security or safety defined?

README.md
```diff
@@ -275,12 +274,27 @@ This waiting period allows time for community comment and review, and to provide
...
+It is not intended for avoiding people installing releases with bugs, for that use a patch release -- potentially after doing nothing more than revert. All patch releases prior to the most recent have known bugs by definition.
```
vchuravy: While it is true that all patches prior have bugs, the goal of a patch release is to fix a bug. I agree that it would be weird to yank all prior variants of that release.

What should we do with a patch release that introduces a new bug? (And who determines criticalness?) The yank that has caused these recent discussions is was an atomic release (one that only changed one thing) and that introduced (from the kinds of usage the packages sees) a severe bug.

A new patch release reverting the change was made and the version was yanked. Modulo the judgement of the maintainer on severity that seems to be the exact purpose of yanking?

vchuravy (commented):
I do think we should not add more stringent guidelines around yanking. Explanation on the effect of it and what it does is great.

This feels like we are trying to fix a process failure (e.g. Pkg unable to handle yanking) with a guideline change.

README.md
```diff
@@ -275,12 +274,29 @@ This waiting period allows time for community comment and review, and to provide
...
 Registry maintainers are discouraged from merging packages they directly *or* indirectly maintain as mentioned in [CONTRIBUTING.md](https://github.com/JuliaRegistries/General/blob/master/CONTRIBUTING.md#other-ways-to-help).
 
+### When is yanking a release appropriate?
+
+Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release entry in `Versions.toml`. This will mean
+that Pkg will no longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General.
+
+The primary reason yanking exists is to make a release impossible to install. This is intended to be used for circumstances when having this
+release installed poses a risk to security or safety. It is not intended for avoiding people installing releases with bugs, for that use a
+patch release -- potentially after doing nothing more than revert. All patch releases prior to the most recent have known bugs by definition.
```
Seelengrab, line 290:
```suggestion
patch release -- potentially after doing nothing more than `git revert`  in the package repo. All patch releases prior to the most recent have known bugs by definition.
```

README.md
Seelengrab: > What should we do with a patch release that introduces a new bug?

Ask maintainers to make a new patch release.

> Modulo the judgement of the maintainer on severity that seems to be the exact purpose of yanking?

A severe package/usage bug is not generally a security vulnerability. A package merely not working correctly is different to that package causing security vulnerabilities down the road for users of that package.

README.md
```diff
@@ -275,12 +274,29 @@ This waiting period allows time for community comment and review, and to provide
...
+
+There is however, a special category of bugged releases that can not be resolved by having a patch release. These also need to be resolved by
+yanking. That special category is when the compat bounds have been set too wide. i.e. say `v2.10.0` was released using a feature not on julia
+`v1.6` but the compat entry for julia was not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would
+not solve the issue as on julia v1.6 Pkg would still resolve the broken `v2.10.0`, and as a minor bump, reverting the code changes would not
+be valid in a patch bump.
+
+If yanking is urgent, open a PR and raise it on the `#pkg-registration` slack channel
```
LilithHafner, line 298:
```suggestion
If yanking is urgent, open a PR and raise it on the `#pkg-registration` [slack channel](https://julialang.org/slack/)
```

README.md
```diff
@@ -275,12 +274,29 @@ This waiting period allows time for community comment and review, and to provide
...
+### Who can approve an early merge?
 
 Any registry maintainer may merge a package registration PR early.
 Early merges should still be discussed on the [#pkg-registration](https://julialang.slack.com/archives/C6M4DQA5P) slack channel before merging.
 Registry maintainers are discouraged from merging packages they directly *or* indirectly maintain as mentioned in [CONTRIBUTING.md](https://github.com/JuliaRegistries/General/blob/master/CONTRIBUTING.md#other-ways-to-help).
 
+### When is yanking a release appropriate?
+
+Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release entry in `Versions.toml`. This will mean
+that Pkg will no longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General.
```
LilithHafner, line 286:
```suggestion
Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release entry in `Versions.toml`
([example](https://github.com/JuliaRegistries/General/pull/94858/files)). This will mean that Pkg will no
longer consider them to exist and not resolve those versions, but that version cannot be re-registed in General.
```

README.md
```diff
@@ -275,12 +274,29 @@ This waiting period allows time for community comment and review, and to provide
...
+
+The primary reason yanking exists is to make a release impossible to install. This is intended to be used for circumstances when having this
```
LilithHafner, line 288:
```suggestion
The reason yanking exists is to make a release impossible to install. This is intended to be used for circumstances when having this
```
Are there any other reasons?

LilithHafner (commented):
I think that these are good guidelines for folks making PRs to yank versions.

For folks deciding whether to merge yank PRs, I think a good guideline is: if the person making the PR has the authority to act on behalf of the package, go ahead and merge it. The only cost to the general registry is the maintenance burden of verifying the authority of the person making the PR, and clicking merge. All other costs are born by the package's users and it's the responsibility of the package author(s) to consider those costs. If they put in the work to make the PR, we can put in the (minimal) work to merge.

README.md
oxinabox: Yes, the second reason is because the package manager generally will chose to use an older patch release if that is what is required for compatibility with something else.
Which is a problem if the bug in the older patch release is incorrect compat bounds.

(I have argued that for this reason Pkg shouldn't actually install older patch releases without a special flag, but people disagreed, so yanking is often required when your bug is with compat bounds)

ericphanson (commented):
oops, forgot to leave this comment

README.md
LilithHafner: I think of the compat reason as a subset of "make a release impossible to install."

LilithHafner: Though I guess the "primary" here refers to the following sentence, not the sentence it is a part of?

LilithHafner:
```suggestion
The reason yanking exists is to make a release impossible to install. This is primary intended to be used for circumstances when having this
```

oxinabox: we should add this, plus a footote aon "install" saying

```
To be precise it will still be possible to install if using the same manifest.
However any new manifest, such as those created by `Pkg.test` or `] up SomeUnrelatedPackage`, or `resolve` after changing julia version, will fail.
```

This will address @MasonProtter 's comment.

IanButterworth:
Can I ask that someone take this over please.

LilithHafner:
Moved to #104800

PR #102850, opened 2024-03-14
Yank DiskArrays 0.4.0

meggart:
The compat entries about the minimum required Julia version for Diskarrays 0.4.0 were not correct.


giordano:
> The compat entries about the minimum required Julia version for Diskarrays 0.4.0 were not correct.

Can't you fix the compat bounds, if that was the only issue?

meggart:
We already did for DiskArrays 0.4.1 and later, but this still leaves 0.4.0 registered and would mean the package manager would resolve to this version for users using julia 1.8 or lower, which leads to a broken installation. We want to make sure julia <=1.8 users stick to DiskArray 0.3.x and I don't see a way to achieve this.

giordano:
I'm still missing what prevents you from fixing the compat bounds in the registry.

meggart:
I see, you mean like this: #104235

PR #103146, opened 2024-03-19
[MITgcmTools] change URL

gaelforget:
Preparing to rename package. Following https://github.com/JuliaRegistries/General#how-do-i-rename-an-existing-registered-package , here I just update the URL


gaelforget:
Thanks @giordano

PR #104235, opened 2024-04-04
Fix DiskArrays compat

meggart:
Update compat instead of yanking release, replaces #102850


giordano:
https://github.com/JuliaRegistries/General/blob/00dfc2edb0c9ac583ef81d47954445adc1101e54/D/DiskArrays/Compat.toml#L7 has to change to something like
```toml
["0.3"]
```
otherwise you have conflicting specifications.

rafaqz:
What is happening here... the yank was such an obvious quick fix to just merge, and now we have chaos instead:

https://github.com/yeesian/ArchGDAL.jl/actions/runs/8969588301/job/24631262216?pr=423

giordano:
> is the idea that instead we just cut out 0.4.0 completely from any julia compatability?

No.

rafaqz:
But isn't that what your change does? wont there be no version if julia where 0.4.0 works?

If so, what version will it work on?

giordano:
> But isn't that what your change does?

No.

> wont there be no version of julia where 0.4.0 can be added?

Julia v1.9+

rafaqz:
Ahh right there always needs to be a version change in two places, not just the one, or there is a conflict.

The example link goes to the file without the current PR changes so skimming it looks like you meant "instead of", but I guess thats just github UI not letting you link the PR file.

meggart:
Sorry to all for losing attention on this one. I hope this will fix it now.

PR #104800, opened 2024-04-12
Add yanking guidance to the readme faq

LilithHafner:
Continuation of #102636, taken over at Ian Butterworth's request. New PR so that I can accept review suggestions. cc @oxinabox @ericphanson @vchuravy @Seelengrab

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,13 +311,13 @@ not raised in the release. In this case releasing a `v2.10.1` with the corrected
 solve the issue as on julia v1.6 Pkg would still resolve the broken `v2.10.0`, and as a minor bump, reverting
 the code changes would not be valid in a patch bump. In this case one may either submit a PR to retroactively
 adjust the compat bounds of previous versions (best user-facing results, but slow and error-prone to implement)
-or yank the offending release.
+or yank the offending release. See the [SciML collaborative practices for more guidance](https://github.com/SciML/ColPrac?tab=readme-ov-file#accidental-support-for-an-unsupported-dependency).
 
 If yanking is urgent, open a PR and raise it on the `#pkg-registration` [slack channel](https://julialang.org/slack/)
 
 For releases with actively exploited security vulnerabilities (i.e. malicious code) yanking is not suficient.
-In this case, the release should be completely deleated from the registry and a new patch release should be
-issued. This is a drastic measure that breaks reporoducivility garuntees and has never been warrented as of 
+In this case, the release should be completely deleted from the registry and a new patch release should be
+issued. This is a drastic measure that breaks reproducibility guarantees and has never been performed as of 
 April, 2024.
 
 ## Registry maintenance
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Currently the waiting period is as follows:
 
 ## FAQ
 
-#### Do I need to register a package to install it?
+### Do I need to register a package to install it?
 
 No, you can simply do `using Pkg; Pkg.add(url="https://github.com/JuliaLang/Example.jl")`
 or `] add https://github.com/JuliaLang/Example.jl` in the Pkg REPL mode
@@ -90,7 +90,7 @@ Registering allows the package to be added by `Pkg.add("Example")` or `] add Exa
 in the Pkg REPL mode. This is true if the package is installed in any registry
 you have installed, not just General; you can even create your own registry!
 
-#### Should I register my package now?
+### Should I register my package now?
 
 If your package is at a stage where it might be useful to others, or provide functionality other
 packages in General might want to rely on, go for it!
@@ -121,14 +121,14 @@ for example [LocalRegistry.jl](https://github.com/GunnarFarneback/LocalRegistry.
 if you decide to go this route.
 * "Empty" packages that do not yet have functionality are not ready to be registered.
 
-#### Can my package in this registry depend on unregistered packages?
+### Can my package in this registry depend on unregistered packages?
 
 No. In this registry, your package cannot depend on other packages that are
 unregistered. In addition, your package cannot depend on an unregistered
 version of an otherwise registered package. Both of these scenarios would cause
 this registry to be unreproducible.
 
-#### Can my package be registered if it requires a version of Julia that is not yet released?
+### Can my package be registered if it requires a version of Julia that is not yet released?
 
 Yes, if your package is ready for use, it can be registered before Julia itself has a compatible release.
 The [`compat` mechanism](https://pkgdocs.julialang.org/v1/compatibility/)
@@ -146,7 +146,7 @@ If you for some reason can't (or won't) adhere to the guidelines you will have
 to wait for a human to review/merge the pull request. You can contact a human
 in the `#pkg-registration` channel in the official Julia Slack to expedite this process.
 
-#### My package fails to load because it needs proprietary software/additional setup to work, what can I do?
+### My package fails to load because it needs proprietary software/additional setup to work, what can I do?
 
 Before merging a pull request, AutoMerge will check that your package can be installed and
 loaded.  It is OK for your package to not be fully functional, but making it at least load
@@ -162,17 +162,17 @@ packages with proprietary software that use the environment variable check inclu
 [`Gurobi.jl`](https://github.com/jump-dev/Gurobi.jl) and
 [`CPLEX.jl`](https://github.com/jump-dev/CPLEX.jl).
 
-#### My pull request has a merge conflict, what do I do?
+### My pull request has a merge conflict, what do I do?
 
 Retrigger Registrator.
 
-#### How do I retrigger Registrator in order to update my pull request?
+### How do I retrigger Registrator in order to update my pull request?
 
 Do what you did when you triggered Registrator the first time.
 
 For more details, please see the [Registrator.jl README](https://github.com/JuliaRegistries/Registrator.jl/blob/master/README.md).
 
-#### I commented `@JuliaRegistrator register` on a pull request in the General registry, but nothing happened.
+### I commented `@JuliaRegistrator register` on a pull request in the General registry, but nothing happened.
 
 If you want to retrigger Registrator by using the Registrator comment-bot,
 you need to post the `@JuliaRegistrator register` comment on a commit in
@@ -180,19 +180,19 @@ you need to post the `@JuliaRegistrator register` comment on a commit in
 any comments of the form `@JuliaRegistrator ...` in the `JuliaRegistries/General`
 repository.
 
-#### AutoMerge is blocked by one of my comments, how do I unblock it?
+### AutoMerge is blocked by one of my comments, how do I unblock it?
 
 Simply edit `[noblock]` into all your comments. AutoMerge periodically
 checks each PR, and if there are no blocking comments when it checks
 (i.e. all comments have `[noblock]` present), it will continue to merge
 (assuming of course that all of its other checks have passed).
 
-#### Are there any requirements for package names in the General registry?
+### Are there any requirements for package names in the General registry?
 
 There are no hard requirements, but it is *highly recommended* to follow
 the [package naming guidelines][naming-guidelines].
 
-#### What to do when asked to reconsider/update the package name?
+### What to do when asked to reconsider/update the package name?
 
 If someone comments on the name of your package when you first release it it is often
 because it does not follow the [naming guidelines][naming-guidelines]. If you think that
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -271,25 +270,46 @@ See [these](https://github.com/JuliaRegistries/General/pull/31549#issuecomment-7
...
 Registry maintainers are discouraged from merging packages they directly *or* indirectly maintain as mentioned in [CONTRIBUTING.md](https://github.com/JuliaRegistries/General/blob/master/CONTRIBUTING.md#other-ways-to-help).
 
+### When is yanking a release appropriate?
+
+Releases that have already merged can be marked as "yanked" by adding `yanked = true` under the release
+entry in `Versions.toml` ([example](https://github.com/JuliaRegistries/General/pull/94858/files)). This
+tells Pkg to no longer resolve those versions (e.g. when running `Pkg.add()` or `Pkg.resolve()`). However,
+to maintain perfect reproducibility of environments specified with a full manifest, yanked versions that
+are specifically listed in a manifest will continue to instantiate when running `Pkg.instantiate()` and
+yanked versions cannot be re-registered in General with different source code.
```
Seelengrab, line 299: There has been discussion recently on slack (cc @mbauman) with the whole xz kerfuffle about no longer instantiating yanked versions either, in order to protect users from a vulnerability they may not even know about. This is also consistent with the next paragraph, stating the primary purpose of yanking to be "to make a release impossible to install". 

We should decide on a direction this should move in before writing down a behavioral guarantee for instantiation.

README.md
```diff
@@ -271,25 +270,46 @@ See [these](https://github.com/JuliaRegistries/General/pull/31549#issuecomment-7
...
+
+The primary reason yanking exists is to make a release impossible to install. This is intended to be used for circumstances when having this
+release installed poses a risk to security or safety. It is not intended for avoiding people installing releases with bugs, for that use a
+patch release -- potentially after doing nothing more than `git revert`  in the package repo. All patch releases prior to the most recent have known bugs by definition.
+
+There is however, a special category of bugged releases that can not be resolved by having a patch release. These also need to be resolved by
+yanking. That special category is when the compat bounds have been set too wide. i.e. say `v2.10.0` was released using a feature not on julia
+`v1.6` but the compat entry for julia was not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would
+not solve the issue as on julia v1.6 Pkg would still resolve the broken `v2.10.0`, and as a minor bump, reverting the code changes would not
+be valid in a patch bump.
```
Seelengrab, line 309: Could this sort of too-wide compat be resolved by making a PR to the registry with the correct bounds instead of yanking? After all, the release is (a priori) good, just not for all originally specified versions.

Seelengrab:
Thanks for taking up the effort!

ericphanson (commented):
I disagree about the primary purpose of yanking being to make it **impossible** to install a version. Deleting the version from the registry does that. Yanking has the special property that existing manifests will continue to work. Therefore, I think _deleting the version_ should be used for actively exploited malicious activity. 

I think yanking should be used for lesser situations (e.g. in the XZ builds that were yanked, at the time of yanking it was suspected that the backdoor was not present in our builds, so yanking was a less-disruptive first step), or for compat-related issues as detailed here.

BTW: pretty much an aside, but Julia 1.0 does not respect yanks (support was added in v1.1), which is maybe good to know.

Seelengrab:
> Therefore, I think deleting the version should be used for actively exploited malicious activity.

Deleting also has the unintended sideeffect of making the registry (and by extension, registry maintainers/the community) forget that the version was already released previously and subsequently removed. Yes, that information is still technically available in the git history, but that's much less likely to be checked compared to an immediate "hey this was yanked previously".

On the other hand, what exactly is the upside of keeping known-insecure versions installable for existing manifests?

LilithHafner:
In my [very limited experience](https://github.com/JuliaRegistries/General/pull/85431), adjusting compat bounds on existing versions is pretty hard. But if that can be done easily and reliably then I think it should replace yanking for the second use case. Then Pkg could e.g. print a warning and prompt for confirmation when instantiating a yanked version (IMO there should always be _a_ way to install yanked versions, even if only an `UNSAFE_IGNORE_YANKS` variable)

ericphanson:
> > Therefore, I think deleting the version should be used for actively exploited malicious activity.
> 
> Deleting also has the unintended sideeffect of making the registry (and by extension, registry maintainers/the community) forget that the version was already released previously and subsequently removed. Yes, that information is still technically available in the git history, but that's much less likely to be checked compared to an immediate "hey this was yanked previously".

Given that has never actually happened, that doesn't seem that important? And it could be solved many other ways. We could keep a markdown doc in General with a list of such versions if we ever find one.

> On the other hand, what exactly is the upside of keeping known-insecure versions installable for existing manifests?

Huh? I am saying to not do that?

---

I think you want to use yanking for something other than what it is meant for. _Yanking is a Pkg.jl feature that exists in order to make a version installable-from-a-manifest, but not resolvable._ Like that is what is was made for. It has special support in Pkg to make this possible. The thing you want, to make something not installable, is 100% possible without special support in Pkg. Namely, to remove it from the registry. The reason yanking exists is because sometimes we want to make something not resolvable but without breaking reproducibility.

LilithHafner:
How does this sound? (up to definitions of "very bad" and "very very bad")

Compat bounds were too wide => PR General to retroactively adjust compat bounds
Very bad release, worse than any ordinary bug => Yank
Very very bad release e.g. actively exploited security vulnerability => Delete from registry

ericphanson:
That sounds reasonable to me. I think yanking is also ok for trying to address compat issues, given the registries compat format is undocumented, we have no tooling setup to adjust bounds that I know of, and we don’t have good CI checks to verify it is correct. IMO if we had some of those things we could prefer adjusting bounds over yanking, but given that we don’t, manually adjusting bounds is error prone and requires careful review, while yanking is straight forward.

LilithHafner:
Okay. Once #104849 is in better shape we can change the guidance to what I recommended above, and until then also recommend yanking for the too-wide compat bounds case.

Seelengrab:
> Huh? I am saying to not do that?

Both `]add` and `]instantiate` are an act of installing packages. Instantiating old manifests that have vulnerable versions is IMO just as bad as adding that version to new manifests. I'm proposing we treat these two cases the same, in order to protect the users of the registry from unknowingly installing vulnerable versions, both through instantiation as well as adding. No one is going to actively check their old manifests for security vulnerabilities; thus, this should be something that's handled directly by the registry.

> Very bad release, worse than any ordinary bug => Yank

I think this is not precise enough. What exactly is "worse than an ordinary bug"? In what situation would a package developer choose a yank over a straight-up deletion of a version, if the yank still permits the chance that someone gets a horribly broken version?

> Very very bad release e.g. actively exploited security vulnerability => Delete from registry

How are we going to keep record of previously deleted versions, to prevent double registrations? Are we actively going to check the git history of every package?

---

I know that we don't have tooling & documentation around the Registry.toml format and adjusting compat bounds. This is an opportunity to remedy that, since those are quite simply operations that occur during regular maintenance of the registry.

README.md
LilithHafner: For "impossible to install" we have deletion from the registry.

LilithHafner:
I put [my suggestion](https://github.com/JuliaRegistries/General/pull/104800#issuecomment-2053730798), with revisions due to @ericphanson's [observation](https://github.com/JuliaRegistries/General/pull/104800#issuecomment-2053733450) that editing compat bounds is harder than yanking, into prose in the latest commit.

README.md
Seelengrab: Deleting a version has a higher maintenance overhead for registry maintainers, since people then have to actively check the past git history for previously conflicting registrations. I don't think any previously registered version should ever be deleted from the registry, because it makes it much harder than necessary to keep track of changes to the registry.

README.md
```diff
@@ -271,25 +270,56 @@ See [these](https://github.com/JuliaRegistries/General/pull/31549#issuecomment-7
...
+the code changes would not be valid in a patch bump. In this case one may either submit a PR to retroactively
+adjust the compat bounds of previous versions (best user-facing results, but slow and error-prone to implement)
+or yank the offending release.
+
+If yanking is urgent, open a PR and raise it on the `#pkg-registration` [slack channel](https://julialang.org/slack/)
+
+For releases with actively exploited security vulnerabilities (i.e. malicious code) yanking is not suficient.
+In this case, the release should be completely deleated from the registry and a new patch release should be
+issued. This is a drastic measure that breaks reporoducivility garuntees and has never been warrented as of 
+April, 2024.
```
ericphanson, line 321:
```suggestion
issued. This is a drastic measure that breaks reproducibility guarantees and has never been performed as of 
April, 2024.
```

README.md
```diff
@@ -271,25 +270,56 @@ See [these](https://github.com/JuliaRegistries/General/pull/31549#issuecomment-7
...
+yank a very broken release, you should typically also release a patch release.
+
+There is however, a special category of bugged releases that can not be resolved by having a patch release.
+These also may to be resolved by yanking. That special category is when the compat bounds have been set too
+wide. i.e. say `v2.10.0` was released using a feature not on julia `v1.6` but the compat entry for julia was
+not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would not 
+solve the issue as on julia v1.6 Pkg would still resolve the broken `v2.10.0`, and as a minor bump, reverting
+the code changes would not be valid in a patch bump. In this case one may either submit a PR to retroactively
+adjust the compat bounds of previous versions (best user-facing results, but slow and error-prone to implement)
+or yank the offending release.
```
ericphanson, line 314: We could also link colprac here, which has some detailed guidance: https://github.com/SciML/ColPrac?tab=readme-ov-file#accidental-breaking-releases and https://github.com/SciML/ColPrac?tab=readme-ov-file#accidental-support-for-an-unsupported-dependency

README.md
```diff
@@ -271,25 +270,56 @@ See [these](https://github.com/JuliaRegistries/General/pull/31549#issuecomment-7
...
+
+If yanking is urgent, open a PR and raise it on the `#pkg-registration` [slack channel](https://julialang.org/slack/)
+
+For releases with actively exploited security vulnerabilities (i.e. malicious code) yanking is not suficient.
+In this case, the release should be completely deleated from the registry and a new patch release should be
```
ericphanson, line 319:
```suggestion
In this case, the release should be completely deleted from the registry and a new patch release should be
```

ericphanson (approved):

README.md
LilithHafner:
```suggestion
If a release is very broken (e.g. contains security vulnerabilities or bugs that are significantly worse 
than the package erroring on load), then it may be yanked. Releasing an ordinary patch release—potentially 
reusing the exact same source code as a previous release without the major bug—is a faster, easier to 
deploy, and less disruptive way to prevent most users from installing a specific version. If you seek to
yank a very broken release, you should typically also release a patch release.

There is however, a special category of bugged releases that can not be resolved by having a patch release.
These also may to be resolved by yanking. That special category is when the compat bounds have been set too
wide. i.e. say `v2.10.0` was released using a feature not on julia `v1.6` but the compat entry for julia was
not raised in the release. In this case releasing a `v2.10.1` with the corrected julia compat would not 
solve the issue as on julia v1.6 Pkg would still resolve the broken `v2.10.0`, and as a minor bump, reverting
the code changes would not be valid in a patch bump. In this case one may either submit a PR to retroactively
adjust the compat bounds of previous versions (best user-facing results, but slow and error-prone to implement)
or yank the offending release. See the [SciML collaborative practices for more guidance](https://github.com/SciML/ColPrac?tab=readme-ov-file#accidental-support-for-an-unsupported-dependency)
```

ericphanson:
> I think this is not precise enough. What exactly is "worse than an ordinary bug"?

I agree this is imprecise. But I think having something here is better than nothing, so let's merge this and if someone proposes more precise language in another PR we can consider that there.

> In what situation would a package developer/registry maintainer choose a yank over a straight-up deletion of a version, if the yank still permits the chance that someone gets a horribly broken version?

From the perspective of the registry, it's basically "fine" to get horribly broken code. That is the ordinary course of business when using code/the registry, and there is an ordinary, regularly-exercised fix: update to a non-broken version. We would never _delete_ a version due to the brokenness of the code. If it is doing the wrong thing, and we delete the bad version, how can you compare the old vs new to understand how much the bug affected your downstream application?

> How are we going to keep record of previously deleted versions, to prevent double registrations? Are we actively going to check the git history of every package?

I think we can cross that bridge when we get to it. Hopefully not for a long time.

> I know that we don't have tooling & documentation around the Registry.toml format and adjusting compat bounds. This is an opportunity to remedy that, since those are quite simply operations that occur during regular maintenance of the registry.

Is this more of an opportunity than any other time? The problem has been known for years. I'm not going to block merging this helpful advice which is fit for the current status quo in favor of waiting for something that may or may not happen.

LilithHafner:
@StefanKarpinski, would you like us to reconsider this decision and/or revert this PR?

PR #105052, opened 2024-04-16
Change `IteratorSampling` url for renaming to `StreamSampling`

ameligrana:
I'd like to change the name of the package from `IteratorSampling` to `StreamSampling` (repo: https://github.com/JuliaDynamics/StreamSampling.jl) because it better reflects its purpose. I have read at https://github.com/JuliaRegistries/General?tab=readme-ov-file#how-do-i-rename-an-existing-registered-package that this is the first step to do before registering it with the new name.


Tortar:
Can someone please take a look at this? Should take a second

PR #105696, opened 2024-04-26
[JLLs] Arrow_jll: avoid snappy_jll 1.2

visr:
The new snappy_jll broke Arrow_jll and therefore GDAL_jll; https://github.com/JuliaGeo/GDAL.jl/issues/171
This adds a compat entry in the registry to stop the breakage, giving us time to add compat for new builds in Yggdrasil.


kongdd:
Hope this PR could be accept quickly. It leads to many Geo packages broken.

PR #105736, opened 2024-04-27
Changing URL for HorizonsEphemeris

jdc-pub:
I have moved this package to the `EphemerisSources.jl` repository.


cadojo:
I have checked, and all previously published versions' commits are still available in the new GitHub repository, except the first two versions. Those versions are _also_ not available in `HorizonsEphemeris.jl`'s current repository; I believe this is a quirk that resulted from issues I ran into when using TagBot and the JuliaRegistrator bot before the package was released. These issues caused me to bump the version, without any change in content, to get the package released. If you'd like, I can track down the GitHub traffic that documents that issue, but I do not believe it is a blocker for merging. 

Please let me know if it would be helpful for me to provide more information!

PR #105738, opened 2024-04-27
Changing URL for SPICEBodies

jdc-pub:
I have moved this package to the `EphemerisSources.jl` repository.


cadojo:
I have checked, and all previously published versions' commits are still available in the new GitHub repository.

PR #106004, opened 2024-05-02
Retroactively require Gtk4 0.6.2 for ImageView 0.12.2

timholy:
xref https://github.com/JuliaImages/ImageView.jl/pull/304

CC @jwahlstrand


I/ImageView/Compat.toml
```diff
@@ -113,11 +113,12 @@ GtkObservables = "2"
 ["0.12.0"]
 Gtk4 = "0.5"
 
-["0.12.1-0"]
+["0.12.1"]
 Gtk4 = "0.6"
 
 ["0.12.2-0"]
 GtkObservables = "2.1.0-2"
+Gtk4 = "0.6.2"
```
giordano, line 121: I think this means only 0.6.2 is allowed, is that what you want?

timholy: Definitely not, thanks! Better now?

giordano (approved):

PR #106306, opened 2024-05-07
Adjust StyledStrings repo URL

tecosaur:
I think the JuliaHub register bot could do with updating, since in #106027 the repo URL was canonicalised in the PR text but not the `Package.toml`.

This means that a minor release that improves precompilation warnings when used with another package results in:

> Error while trying to register: Changing package repo URL not allowed, please submit a pull request with the URL change to the target registry and retry.

In the meantime, this PR fixes the situation with StyledStrings.


DilumAluthge (approved):

tecosaur:
Thanks!

PR #106353, opened 2024-05-07
fix compat for VideoIO on FFMPEG_jll

IanButterworth:
Currently the newer ffmpeg_jll version forces VideoIO downgrade because it's not compatible.

```
(Foo) pkg> up
    Updating registry at `~/.julia/registries/General.toml`
   Installed BinaryProvider ─ v0.5.10
   Installed FFMPEG ───────── v0.2.4
   Installed FFMPEG_jll ───── v6.1.1+0
   Installed VideoIO ──────── v1.0.8
  Downloaded artifact: FFMPEG
  No Changes to `~/Documents/GitHub/Foo/Project.toml`
    Updating `~/Documents/GitHub/Foo/Manifest.toml`
  [b99e7846] + BinaryProvider v0.5.10
⌃ [c87230d0] ↓ FFMPEG v0.4.1 ⇒ v0.2.4
⌃ [d6d074c3] ↓ VideoIO v1.1.0 ⇒ v1.0.8
  [b22a6f82] ↑ FFMPEG_jll v4.4.4+1 ⇒ v6.1.1+0
```

I don't think this is quite right yet as VideoIO only took on FFMPEG_jll as a dep in v1.0.9 but I thought I'd open for discussion


IanButterworth:
I don't believe it's valid to add dependencies in General to old versions, so I don't think this can be fixed here.

The only recourse is for the user to specify a lower bound of VideoIO v1.0.9

PR #106452, opened 2024-05-09
Ensure CURL_jll matches LibCURL_jll more precisely

staticfloat:
It is not allowable for CURL_jll to get a version of LibCURL_jll that varies too much from its own version.  Note that because LibCURL_jll is a stdlib this is not a problem for most users, but if you're using something like JLLPrefixes.jl to do your package instantiation, you can end up in a situation like this.

### DIFF
--- a/jll/C/CURL_jll/Compat.toml
+++ b/jll/C/CURL_jll/Compat.toml
@@ -11,7 +11,7 @@ Artifacts = "1"
 Libdl = "1"
 
 ["8-8.5"]
-LibCURL_jll = "8.5.0-8.5"
+LibCURL_jll = "8.5.0-8.6"
 
 ["8.6-8"]
 LibCURL_jll = "8.6.0-8.6"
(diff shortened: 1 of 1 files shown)


jll/C/CURL_jll/Compat.toml
```diff
@@ -11,7 +11,7 @@ Artifacts = "1"
 Libdl = "1"
 
 ["8-8.5"]
-LibCURL_jll = "8.5.0-8"
+LibCURL_jll = "8.5.0-8.5"
```
giordano, line 14: Can probably also allow 8.6 here
```suggestion
LibCURL_jll = "8.5.0-8.6"
```

staticfloat: I thought they have to match precisely?

giordano: I thought so too, but if no one noticed big problems before maybe we have some flexibility 😛 we can further tighten the compat later if necessary, but at least we avoid the completely incompatible combination with this PR.

PR #106848, opened 2024-05-15
[GeoMakie] Retroactively cap GeometryBasics versions to avoid method …

asinghvi17:
…overwriting

A hack from GeoMakie was upstreamed into GeometryBasics in v0.4.11 of GeometryBasics, this PR caps all versions before the upstreaming.


ViralBShah:
Would be nice to get another set of eyes on this. @asinghvi17 Please chime in to bump if we don't get it merged in a couple of days.

PR #107391, opened 2024-05-22
Yank some versions of OptimizationBase

gdalle:
This PR yanks three consecutive versions of OptimizationBase (v0.0.6, v0.0.7, v1.0.0) because they incorrectly promised compatibility with ADTypes v1.

Question: is this necessary, or would a patch release be enough? My fear is that the patch release v0.0.8 is incompatible with v0.0.7, and so some people would not automatically get it (due to the [double leading zero](https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0)). Am I reading that correctly?

See also:

- https://github.com/SciML/OptimizationBase.jl/issues/44
- https://github.com/SciML/ModelingToolkit.jl/pull/2726#issuecomment-2124473525


gdalle:
What do you conclude from this? That yanking is indeed necessary here?

devmotion:
IMO based on the README yanking would be fine, but (if maintainers - and user? - prefer it) something like https://github.com/JuliaRegistries/General/pull/107409 could be an alternative.

Vaibhavdixit02:
Tbh I had just overlooked the part that you can retrospectively change the compat bounds. I don't quite get what errors can be expected mentioned there?

gdalle:
I'm just wary (and I guess David too) because the syntax of compat bounds is slightly harder to get right than just `yanked=true`

ChrisRackauckas:
Yank is fine given there are versions before and after. Thanks for going through with this, sorry it took awhile to get to reviewing.

PR #107409, opened 2024-05-22
Fix compat of OptimizationBase

devmotion:
**I have never adjusted compat entries in the registry (I think) and have no idea whether this PR is actually correct**

Possible alternative to #107391, as indicated in the [README](https://github.com/JuliaRegistries/General?tab=readme-ov-file#when-is-yanking-a-release-appropriate):

> There is however, a special category of bugged releases that can not be resolved by having a patch release. These also may to be resolved by yanking. That special category is when the compat bounds have been set too wide. i.e. say v2.10.0 was released using a feature not on julia v1.6 but the compat entry for julia was not raised in the release. In this case releasing a v2.10.1 with the corrected julia compat would not solve the issue as on julia v1.6 Pkg would still resolve the broken v2.10.0, and as a minor bump, reverting the code changes would not be valid in a patch bump. In this case one may either submit a PR to retroactively adjust the compat bounds of previous versions (best user-facing results, but slow and error-prone to implement) or yank the offending release.


O/OptimizationBase/Compat.toml
```diff
@@ -1,4 +1,4 @@
-["0-0.0.5"]
+[0]
```
giordano, line 1: Why not
```suggestion
["0-1.0.0"]
```
and remove the duplicate section below?  I'm surprised RegistryCI doesn't compress it like that already.

devmotion: Because I was not sure since also currently on master the ADTypes compat for 1.0.0 is not combined with the other versions.

devmotion:
It seems #107391 was merged as well - should we revert the other PR @giordano?

PR #107721, opened 2024-05-27
[JLL] Yank broken GDAL and breaking PROJ

visr:
The latest GDAL_jll release is broken on Windows, and since we don't know the cause I don't think we can fix the ecosystem that builds on GDAL and PROJ anytime soon without yanking.

The latest PROJ_jll is breaking on Windows and was mistakenly not released as breaking. This causes GDAL_jll to fail as well since the DLL name changed.

cc @eschnett @rafaqz @joa-quim @evetion 

Ref
https://github.com/JuliaPackaging/Yggdrasil/pull/8746#issuecomment-2130321111
https://github.com/JuliaGeo/GDAL.jl/issues/173

### DIFF
--- a/jll/G/GDAL_jll/Versions.toml
+++ b/jll/G/GDAL_jll/Versions.toml
@@ -109,3 +109,4 @@ git-tree-sha1 = "1740251f4f1ce850d4eaad052d5e419229d4b2af"
 
 ["301.901.0+0"]
 git-tree-sha1 = "420559d37b7f2c3e30f3127d16d2fbd521e4126e"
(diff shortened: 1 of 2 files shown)


joa-quim:
Don't know if related but the error I get on Windows is of not being able to load the libgeotiff library (_symbol not found_). It happens that a new version of libgeotiff was released a few days ago, so maybe there can be a relation. A possibility of solving the GDAL issue for now would be to use its internal geotiff library, but it wouldn't solve the PROJ break (that I was not aware).

PR #109007, opened 2024-06-14
yank pandoc_crossref_jll 0.3.17+1

palday:
The `+1` build added a new compat entry which applies to all builds and this breaks previously resolvable code. This PR yanks the problematic build version and removes the invalid compat.

cc @omus @giordano @kleinschmidt


omus:
ref: https://github.com/JuliaPackaging/Yggdrasil/pull/8892

omus (approved):

PR #109326, opened 2024-06-19
Update repo URLs for OpenFHE and SecureArithmetic

sloede:
They have been moved to their new home in the HPSC Lab org (https://github.com/hpsc-lab):
* https://github.com/hpsc-lab/OpenFHE.jl
* https://github.com/hpsc-lab/SecureArithmetic.jl


sloede:
cc @ArseniyKholod

PR #109381, opened 2024-06-19
update Package.toml for QuartetNetworkGoodnessFit: new url

cecileane:
after transfer to github organization JuliaPhylo


cecileane:
This package still needs a url update after transfer to the JuliaPhylo organization.

PR #109382, opened 2024-06-19
update Package.toml of PhyloCoalSimulations with new url

cecileane:
after transfer to github organization JuliaPhylo


cecileane:
Hello, this is 1 of 2 packages that still need a url update after transfer to the JuliaPhylo organization.
Thank you so much for a great package manager!

PR #109505, opened 2024-06-21
Update Tapir.jl URL

willtebbutt:


wsmoses:
Does it make sense to fix the package name too at the same time?

willtebbutt:
I'd rather we don't conflate this technical issue around @yebai having renamed the organisation in which we host this package with the package name itself.

I wouldn't mind, but this PR is in the way of our release cycle, and I'd like to keep things moving (we operate continuous delivery and a tag-each-PR-to-src policy, so having this PR open is reasonably disruptive).

PR #111061, opened 2024-07-14
Yank Static v1.1.0

chriselrod:
This is to prevent the resolver from being able to select Static 1.1.0 as a new version of static with an old precompile tools, so we don't get any test failures like https://github.com/sumiya11/Groebner.jl/actions/runs/9929029946/job/27426062429

@ChrisRackauckas

### DIFF
--- a/S/Static/Versions.toml
+++ b/S/Static/Versions.toml
@@ -153,6 +153,7 @@ git-tree-sha1 = "85fec8e64b6b8168d7a172d7f503c6770c7cfc08"
 
 ["1.1.0"]
 git-tree-sha1 = "0bbff21027dd8a107551847528127b62a35f7594"
+yanked = true
 
 ["1.1.1"]
 git-tree-sha1 = "87d51a3ee9a4b0d2fe054bdd3fc2436258db2603"


IanButterworth:
Was this not fixable with a patch release?

My CI was passing with Static v1.1.0 in the manifest.

Yanking this caused my `Pkg.test` sandbox resolver to fail because it couldn't find the yanked version. So caused test failure (which this was done to avoid).

IanButterworth:
i.e. patch with a revert then reland and minor release?

chriselrod:
Static 1.1.1 was released. 

We needed the yank to stop the resolver from pairing Static 1.1.0 with PrecompileTools 1.0, an invalid combination. 
As long as Static 1.1.0 is selectable by the resolver while PrecompileTools 1.1 is not, I fear it can select the combination of Static 1.1.0 + PrecompileTools 1.0.

Static 1.1.1's compat requires PrecompileTools 1.1, hence 1.1.1 won't be picked. 
The resolver may then prefer Static 1.1.0 over 1.0, because the former is newer. The former, however, doesn't precompile successfully. Hence, the need to eliminate it from the candidates. 

I think a one time fix of your environment is better than allowing people to resolve or update their environments into broken configurations, and then need to manually debug and pin their way out of it.
You should be able up just re-reaolve or update to fix it.

IanButterworth:
I don't mean to make a big deal about me having to do something, I'm just trying to weigh up whether there's a way to fix these things without yanking (which causes the issue I raised).

what happened:
```
Static 1.0.0 compat=nothing
Static 1.1.0 compat=PrecompileTools 1 # bad because the compat is wrong
Static 1.1.1 compat=PrecompileTools 1.1
```
what could have been done:
```
Static 1.0.0 compat=nothing
Static 1.1.0 compat=PrecompileTools 1
Static 1.1.1 compat=nothing # revert the code using PrecompileTools
Static 1.1.2 compat=PrecompileTools 1.1 # add the code back and fix the compat
```

Perhaps that last should be Static 1.2.0? Unclear to me.

Though, with all this I had missed that the CI linked to in the OP was for a downgrade CI action... which perhaps does make the yank unavoidable, which is a shame.. Or perhaps downgrade CI should always select the highest patch for the lowest minor?

ChrisRackauckas:
> Or perhaps downgrade CI should always select the highest patch for the lowest minor?

It should, that is currently a request to @StefanKarpinski as part of the new resolver.

PR #111286, opened 2024-07-18
Bugfix/abstract xsd types UUID clash

Tom-Lemmens:
When moving AbstractXsdTypes from the ASML innersource registry to the general registry a clash in UUID was discovered.
This PR aims to fix this UUID clash by changing the UUID for the AbstractXsdTypes package on the general registry.


A/AbstractXsdTypes/Package.toml
```diff
@@ -1,4 +1,4 @@
 name = "AbstractXsdTypes"
-uuid = "1d4aff4b-e088-40cb-b713-ffe4caf1580b"
+uuid = "894546dd-dc3f-42e8-9b69-a7785ccf72be"
```
giordano, line 2: You also need to change the uuid in the top-level Registry.toml file (and move it to the right place)

A/AbstractXsdTypes/Versions.toml
```diff
@@ -1,2 +1,2 @@
 ["0.1.0"]
-git-tree-sha1 = "9b0214aa1bdb299d8a65dc6f25678322f35b2795"
+git-tree-sha1 = "7a644fc2d9a859a953b5fa106f663312cdeb6deb"
```
giordano, line 2: I'm not sure you need to change the commit of the registered version.

ericphanson: probably a good idea since the uuid is commited in the project.toml

Tom-Lemmens: That was indeed my reasoning for updating the commit.

giordano: Uhm, but this should be the git-tree-hash, not the commit has itself.

A/AbstractXsdTypes/Versions.toml
```diff
@@ -1,2 +1,2 @@
 ["0.1.0"]
-git-tree-sha1 = "9b0214aa1bdb299d8a65dc6f25678322f35b2795"
+git-tree-sha1 = "210965258757a9257b98c1bd55e3f1e72b852ee2"
```
giordano, line 2: I believe this should be the git tree hash of the subdir.  @ericphanson can you confirm?
```julia-repl
julia> using Pkg

julia> bytes2hex(Pkg.GitTools.tree_hash("AbstractXsdTypes.jl"))
"c5aea50b0e207f546f48b2470ce0786018a8173f"
```

A/AbstractXsdTypes/Versions.toml
Tom-Lemmens: My bad, copied the wrong value.

A/AbstractXsdTypes/Versions.toml
ericphanson: yeah, it should be the result of that function called on the subdir itself

Tom-Lemmens: Didn't take into account the subdir subtlety.
Should be correct now.
```
julia> bytes2hex(Pkg.GitTools.tree_hash("AbstractXsdTypes.jl"))
"e99adf5c628d7c21e88f3bfbfeb39a4e0c1f061b"
```

giordano: Uhm, on which commit you computed that?
```console
% git log -1 --oneline 
7a644fc (HEAD -> main, origin/main, origin/HEAD) change UUID AbstractXsdTypes
% julia -E 'using Pkg; bytes2hex(Pkg.GitTools.tree_hash("AbstractXsdTypes.jl"))'
"c5aea50b0e207f546f48b2470ce0786018a8173f"
```

Tom-Lemmens: The latest on main:
7a644fc (HEAD -> main, origin/main, origin/HEAD) change UUID AbstractXsdTypes

GunnarFarneback: A tip is to try a registration with LocalRegistry (`commit=false` is a good idea, and you may need to explicitly point to a clone of General) and the chances that you get a good tree hash are good. Anyway, LocalRegistry finds the tree hash with
```
/tmp/XmlStructTools.jl$ git rev-parse HEAD:AbstractXsdTypes.jl
c5aea50b0e207f546f48b2470ce0786018a8173f
```
which agrees with Mose's result above.

Tom-Lemmens: Now I do get the same result:
```
$ git rev-parse HEAD:AbstractXsdTypes.jl
c5aea50b0e207f546f48b2470ce0786018a8173f
```

giordano (approved):
We had a lengthy discussion about this change, and we decided to accept it because it's for a newly registered package, unlikely to be used by anyone else at this present time. But this shouldn't be taken as an endorsement for similar changes in the future, they'll need to be evaluated on a case-by-case basis.  Not changing the General registry in a disruptive way always remains the preferred option.

PR #111362, opened 2024-07-19
[alsa_plugins_jll] Fix compat bound with FFMPEG_jll

giordano:
This should unlock #111347.


Kyjor:
Thank you !! @giordano

PR #111403, opened 2024-07-20
[Exodus_jll] Update Compat.toml

cmhamel:
Adding compat entry to Fmt_jll after the release of Fmt_jll v11

### DIFF
--- a/jll/E/Exodus_jll/Compat.toml
+++ b/jll/E/Exodus_jll/Compat.toml
@@ -1,5 +1,4 @@
 [0]
-Fmt_jll = "9.0.0-1"
 HDF5_jll = "1.12"
 JLLWrappers = "1.2.0-1"
 julia = "1.6.0-1"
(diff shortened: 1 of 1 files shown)


jll/E/Exodus_jll/Compat.toml
```diff
@@ -1,4 +1,5 @@
 [0]
+Fmt_jll = "9.0.0-1"
```
giordano, line 2: This syntax doesn't make sense (it means from version 9.0.0 to 1), you need something like
```toml
Fmt_jll = "9"
```
But also, if the compat is the same everywhere, you can also have a single section
```toml
["0-8"]
Fmt_jll = "9"
```

cmhamel: I went with your second recommendation for a single section. Thanks for the help. This was the first time I had to do anything with the registry directly.

PR #111705, opened 2024-07-24
Transfer NonNegLeastSquares to JuliaLinearAlgebra

timholy:
https://github.com/JuliaLinearAlgebra/NonNegLeastSquares.jl/pull/37#issuecomment-2227438428


DilumAluthge (approved):
I have verified that the old URL automatically redirects to the new URL.

PR #112224, opened 2024-08-01
removing Fable.jl

leios:
Hello,

I would like to remove this package from the General registries. It's not being used by anyone and has a simple enough name that it might conflict with other packages. It is also meant as a standalone piece of software, so removing it can not break anyone's code downstream.

I don't know if this is possible or not, but I thought I would try.

Thanks!


fredrikekre:
Packages are not removed.

> it might conflict with other packages

Let's discuss a possible name transfer when this isn't just a hypothetical issue.

PR #114503, opened 2024-09-04
[BinaryProvider] Make it not installable on Julia v1.11+

giordano:
BinaryProvider is an old and unmaintained package which reaches into Julia's internals, but it isn't compatible with recent Julia versions on many platforms, e.g. Apple Silicon (this architecture is more recent than latest version of BinaryProvider) and Windows.  Sometimes the resolver favours particularly old versions of packages which used to require BinaryProvider, causing errors and frustration for users who don't know what to do.

I'm capping to Julia v1.10 (included) to avoid too much churn for environments instantiated with stable versions of Julia, but I'm happy to discuss this.

I verified that installing `FFMPEG_jll` first and `Plots` afterwards with this PR on Julia v1.11 installs a version of `FFMPEG` which doesn't use `BinaryProvider` (since it isn't installable) and only downgrades `FFMPEG_jll` to a lower version.


fingolfin:
Yes please

staticfloat (approved):

PR #114739, opened 2024-09-07
[DynamicSumTypes] Change link for rename of package

ameligrana:
Hi! I'm in the process to rename DynamicSumTypes.jl to LightSumTypes.jl because I was notified that it was a confusing name and I absolutely agree, so I update the link as required in the official procedure for renaming packages


Tortar:
I think this needs a manual merging from someone, the new package is at https://github.com/JuliaDynamics/LightSumTypes.jl

Issue #104849, opened 2024-04-13
Streamline retroactive compat bounds adjustment


LilithHafner:
- [ ] Document the compatibility specification format
- [ ] Add CI checks to verify that specifications are properly formatted (and only run checks on edited files)
- [ ] Add instructions for retroactive compat bounds adjustment to README and link exemplary merged PRs
- [ ] Add some highly automated @JuliaRegistrator set_compat_bounds(...) invocation

(the 4th item is likely excessive)

felixcremer:
The compat format is described here: https://pkgdocs.julialang.org/v1/registries/#Registry-Compat.toml

DilumAluthge:
> * Document the compatibility specification format

https://pkgdocs.julialang.org/v1/registries/#Registry-Compat.toml

DilumAluthge:
> * Add CI checks to verify that specifications are properly formatted

I think that the existing RegistryCI "registry consistency checks" (that we already run on every PR to the General registry) already perform some checks. Are there specific examples of things that our existing checks miss?

DilumAluthge:
> * Add some highly automated `@JuliaRegistrator set_compat_bounds(...)` invocation

DilumAluthge:
> * Add some highly automated `@JuliaRegistrator set_compat_bounds(...)` invocation

Let's move this to a separate issue on the [Registrator.jl repo](https://github.com/JuliaRegistries/Registrator.jl): https://github.com/JuliaRegistries/Registrator.jl/issues/436

DilumAluthge:
> * Add CI checks to verify that specifications are properly formatted (and only run checks on edited files)

https://github.com/JuliaRegistries/RegistryCI.jl/pull/575

goerz:
Could someone elaborate on what “ retroactive compat bounds adjustment” means, exactly? How would it not break reproducibility? (Cf https://github.com/JuliaRegistries/General/pull/119607#issuecomment-2491369573)

LilithHafner:
> Could someone elaborate on what “ retroactive compat bounds adjustment” means, exactly?

[...]

PR #115345, opened 2024-09-17
Copulas.jl - retroactive compat bound tightening on TaylorSeries.jl

lrnv:
Hey, 

Following guidance from @BioTurboNick on slack and while #104849 is still pending, I try in this PR to retroactively tighten the lower bound on `TaylorSeries.jl` on recent versions of `Copulas.jl` : at the time, I did use a feature that was only available in `TaylorSeries > 0.15` but kept the old compat bound... 

These days, it results in issues such as https://github.com/lrnv/Copulas.jl/issues/207 and https://github.com/lrnv/Copulas.jl/issues/217. If these issues are usually solved by `]up` or even `]up TaylorSeries`, it is still anoying.

**Warning:** Although I have respected [these docs of the compat file](https://pkgdocs.julialang.org/v1/registries/#Registry-Compat.toml), I am **not sure** the fix I applied will be working correctly, and would appreciate some human taking the time to check it up before merging :)


lrnv:
As explained in the PR, I would like this to be manually merged (its a correction of a mistakeon compat bounds). [noblock]

BioTurboNick:
I suggest pinging a maintainer specifically, possibly this is lost in the swamp of PRs.

BioTurboNick:
> @BioTurboNick Do you have commit access to this repo?

Nope, just being helpful :-)

lrnv:
@ViralBShah Okay this time should be good if CI does pass.

PR #117169, opened 2024-10-12
CI: Stop running consistency tests on 1.4 and 1.5 (but keep running them on 1.3)

DilumAluthge:
The idea here is to try to conserve CI resources.

We intentionally keep the 1.3 job; hopefully, that'll help us continue to catch problems that might affect older Julia versions.

### DIFF
--- a/.github/workflows/registry-consistency-ci-cron.yml
+++ b/.github/workflows/registry-consistency-ci-cron.yml
@@ -1,6 +1,6 @@
 name: Registry Consistency (Cron)
 on:
-    schedule:
+  schedule:
     - cron:  '0 0 * * *'
   workflow_dispatch:
 env:
--- a/.github/workflows/registry-consistency-ci-cron.yml
+++ b/.github/workflows/registry-consistency-ci-cron.yml
@@ -1,11 +1,7 @@
-name: Registry Consistency
+name: Registry Consistency (Cron)
 on:
-  pull_request:
-    branches:
-      - master
-  push:
-    branches:
-      - master
+    schedule:
+    - cron:  '0 0 * * *'
   workflow_dispatch:
 env:
     JULIA_PKG_USE_CLI_GIT: true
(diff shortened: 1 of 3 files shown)


ViralBShah (dismissed):

ericphanson:
should we run them weekly or something? 🤔

DilumAluthge:
> should we run them weekly or something? 🤔

Yeah I think that would be useful. And it would have a relatively small impact on CI resource consumption.

DilumAluthge:
Let me figure out a way to do that...

DilumAluthge:
Honestly even once per day would be fine.

DilumAluthge:
Ugh, I can't use `matrix` in a job-level `if:`.

Okay, so it might be easier to make the cron job a completely separate job.

DilumAluthge:
Alright, I made the cron job a completely separate workflow file, because it wouldn't let me use `matrix` in a job-level `if` condition.

If there is a way to simplify or de-duplicate this, then we can do so in a follow-up PR.

DilumAluthge:
I will remove 1.4 and 1.5 from the list of required status checks in the branch protection settings for this repo.

DilumAluthge:
Let's give this a try.

giordano:
> If there is a way to simplify or de-duplicate this, then we can do so in a follow-up PR.

Make it a reusable worfklow.

ViralBShah:
If it can be run once a day, then can we remove 1.7 and 1.8 too?

DilumAluthge:
Let's drop 1.7.

Julia 1.9.0 was released on May 9, 2023, so not much longer than a year ago. And presumably people were using Julia 1.8.x up until then. So maybe let's keep 1.9 for now, and we can drop it once 1.12 comes out?

ViralBShah:
Yeah dropping 1.7 only for now makes sense, and we can remove 1.8 when 1.12 is out.

ViralBShah:
The other useful thing would be to run pre-releases (at least RCs) when a pre-release is in effect. We don't have a good mechanism though for such conditional things (unless we can download the xml file and see if there is a pre-release that is not yet followed by a stable release, and hence test against that).

PR #117174, opened 2024-10-12
CI: Drop 1.7 from the every-PR CI (but keep it in the once-daily cron job CI)

DilumAluthge:


ViralBShah (approved):

ViralBShah:
What is the `check` thing that seems to run on Julia 1.7? Can we bump it to 1.11?

DilumAluthge:
That's just the job we've dropped in this PR. I'll remove it from the required status checks in the branch protection settings.

ViralBShah:
Should we add 1.11 as Required in the branch protection settings too (like the others)? I thought I had added it, but it isn't showing as Required.

DilumAluthge:
We'll need to wait a few days. Because currently there are open PRs (e.g. new package PRs waiting for their 3-day waiting period) that were opened before we started running 1.11. Once we make 1.11 required, those PRs won't be able to be merged unless someone hits the "Update branch" button. This means that we would have to hit the "Update branch" button on all currently open PRs, which will run a lot of CI. It's easier to wait a few days for those old PRs to hit their 3-day waiting period and get automerged. Then we can make 1.11 required.

PR #117175, opened 2024-10-13
Revert "New package: FortranNamelistParser v0.1.0"

ericphanson:
Reverts JuliaRegistries/General#116786 cc @anchal-physics

context: https://discourse.julialang.org/t/remove-package-from-general-registry/71668/5

rationale: 
* in general we don't remove packages from General, but we have on occasion when very soon after registering and we think no one else is using the package yet, and the author made some mistake that is very difficult or labor intensive to fix otherwise
* I don’t see any open-source usages of it on github outside of ProjectTorreyPines org which has that registry they mention and @anchal-physics, the package owner, and [what looks like a mirror of general](https://github.com/codedownio/General) (https://github.com/search?q=FortranNamelistParser&type=code&p=1), but of course there’s code other than open-source code on github. 
* I would vote for reverting today; I don’t think it needs to be precedent but rather, this is handled on a case by case basis and in this case it seems like it won’t affect anyone outside this group, for whom reverting will help a lot. I think maintaining reproducibility and not breaking code is essential but it seems like that won’t happen here, and combining registries can be messy and someone made a mistake that we can help fix.
* waiting longer just increases the odds someone else will use it, so we should revert it quickly or not at all


Keno:
Acceptable to me for reasons provided.

anchal-physics:
Thanks a lot for quick resolution to this issue.

goerz:
Followup: https://github.com/JuliaRegistries/General/pull/117266

PR #116935, opened 2024-10-09
Restrict StaticArrays.jl @v1.5.16- to Julia v1.10-

fhagemann:
Following up on https://github.com/JuliaRegistries/General/pull/86950#issuecomment-2403361516:
In #86950, the `julia` compat was restricted to `≤ 1.10` for `StaticArrays@1-1.5`. 
However, the code which errors was only removed in `1.5.17`, see [here](https://github.com/JuliaArrays/StaticArrays.jl/pull/1136/commits/fdf353022cde23b049ad59e4b7710995458276f4).
This is my attempt on adjusting the Compat.toml accordingly.

PR #117626, opened 2024-10-19
Revert "Restrict StaticArrays.jl @v1.5.16- to Julia v1.10-"

giordano:
Reverts JuliaRegistries/General#116935


fhagemann:
Can we tackle this again after having fixed what caused the error in Pkg.jl?
https://github.com/JuliaLang/Pkg.jl/pull/4077

PR #117703, opened 2024-10-20
Move Finch.jl to a github organization

willow-ahrens:
move Finch.jl to a Github organization (update github url)


DilumAluthge (approved):
1. I have verified that the old URL automatically redirected the browser to the new URL.
2. The author of this registry PR is also the repository owner of the old URL.

PR #118431, opened 2024-10-31
Transfer packages to WIAS-PDELib organization

j-fu:
I (@j-fu) and Christian Merdon (@chmerdon) want to transfer a number of packages authored by us to a new organization WIAS-PDELib. We are both owners of the new organization. In addition, we want to transfer WIAS-BERLIN/PDELib.jl to that org. (I am one of the owners of WIAS-BERLIN org).


j-fu:
@pjaap

chmerdon:
I agree with the suggested changes.

PR #118541, opened 2024-11-02
Register `Compiler` subfolder of JuliaLang/julia

Keno:
This is a companion PR to JuliaLang/julia#56409. The two PRs will need to be merged somewhat contemporaneously to allow downstream consumers to transition their use of `Core.Compiler` to the `Compiler` pseudo-stdlib without interruption.

As described in JuliaLang/julia#56409, the idea here is rather than using the tranditional Julia stdlib mechanism, which does not track exact package contents or versions, the Compiler instead becomes an ordinary Julia package that just happens to live in JuliaLang/julia (for now). Base has a private copy of this package, but users ordinarily aren't supposed to interact with it.

Since it is an ordinary package, it will need to be registered in General, rather than relying on the entry it would have otherwise had in the julia distribution's to-level folder. This PR is intended to facilitate that registration.

Note that I am opening this PR without any actual tagged versions of `Compiler`. My intention is to tag `0.0.1` upon merging the corresponding julia PR, but the full versioning strategy for `Compiler` has yet to be decided upon.

As mentioned, the purpose of this PR is to get pre-agreement on the registration itself to make sure we can merge/tag simultaneously without doing something overly special on the registry side. I do not know if this PR can be merged as is without any versions tagged or if some part of the tooling will be confused by that. If it can, that's probably easiest. If not, I will amend this PR with the appropriate version once the julia PR is merged and expect it to merge manually shortly thereafter.


ericphanson:
looks like this is missing a compat bound for `julia`? BTW I think you can just use Registrator for this next time which would catch that and create the PR automatically.

Keno:
I somewhat deliberately left the compat bound off, so packages could depend on it without becoming uninstallable on older julia versions (where they would use Core.Compiler instead).

PR #118853, opened 2024-11-06
README: Fix CI badge for the registry consistency tests

ViralBShah:
@DilumAluthge I couldn't quite tell what the updated link here should be. Is it ok to delete, or if update it, what should it be updated to?


DilumAluthge:
Let me generate the new URL.

DilumAluthge:
I've put in the updated links.

PR #119150, opened 2024-11-11
Update IndividualDisplacements URL to post-rename repo

gaelforget:


gaelforget:
See https://github.com/JuliaRegistries/General/pull/119126

gaelforget:
I do not understand why `Merging is blocked`.

gaelforget:
I opened this PR because https://github.com/JuliaRegistries/General#readme says 

`if you rename the repository so it has a new URL, make a PR to edit the URL stored in the registry for the old package name to point to the new URL ([example](https://github.com/JuliaRegistries/General/pull/40190/files)). This allows the old versions of the package under the previous name to continue to work.`

and it seems desirable that previous name continues to work.

Did I do something wrong?

GunnarFarneback:
> I do not understand why Merging is blocked.

That's the same for all PRs to General unless you happen to have merge authorization.

> Did I do something wrong?

This PR looks fine to me but since it is a manual PR, automerge doesn't handle it and it needs to be merged by someone who does have merge authorization.

gaelforget:
> > I do not understand why Merging is blocked.
> 
> That's the same for all PRs to General unless you happen to have merge authorization.
> 
> > Did I do something wrong?
> 
> This PR looks fine to me but since it is a manual PR, automerge doesn't handle it and it needs to be merged by someone who does have merge authorization.

Makes sense now. Thanks for the clarification

gaelforget:
Hi. Could this get merged?

PR #119376, opened 2024-11-14
New package: StateSelection

Keno:
This registers https://github.com/JuliaComputing/StateSelection.jl. This package implements structural transformation and analysis of equation systems. This code has bit of history. It started out as a translation of similar code in Modia.jl in a private package. It was thereafter open sourced and incorporated in MTK, then split out again into this package, which is the version used in Cedar's DAECompiler. The package is currently registered in the Cedar public registry, but not General. This PR moves the registration to General to enable other packages in general to depend on it. Hopefully this will become the canonical version of this code that everybody just uses.

Two versions are being registered, 0.2.0, and 0.2.1, which are the two versions that are public in the Cedar registry. There is also a 0.1.0 version, but this version was never publicly registered, so it is omitted from this registration as well.

### DIFF
--- a/S/StateSelection/Package.toml
+++ b/S/StateSelection/Package.toml
@@ -1,0 +1,3 @@
+name = "StateSelection"
+uuid = "64909d44-ed92-46a8-bbd9-f047dfbdc84b"
+repo = "git@github.com:JuliaComputing/StateSelection.jl.git"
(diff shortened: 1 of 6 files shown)


S/StateSelection/Package.toml
```diff
@@ -0,0 +1,3 @@
+name = "StateSelection"
+uuid = "64909d44-ed92-46a8-bbd9-f047dfbdc84b"
+repo = "git@github.com:JuliaComputing/StateSelection.jl.git"
```
giordano, line 3: I believe for most repos we use the https protocol here.

Issue #111019, opened 2024-07-13
Auto block registration of packages pre-1.0


oxinabox:
From [SemVer 2.0](https://semver.org/)

> ### How do I know when to release 1.0.0?

> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backward compatibility, you should probably already be 1.0.0.

I am going to argue that this is *normally* actually the case when you register in general.
If you don't intend for people to use it you shouldn't be registering it.
Perhaps by exception people are registering with intent to iterate rapidly in a breaking way.
But it is exceptional and we can always use the functionality to bypass that rule.

It's problematic we people register pre-1.0
Because 

1. People struggle to draw the line when to hug 1.0, but on registration is a good time 
1. Releasing 1.0 once "stability is achieved" is breaking for no reason.
1. They can't use full range of semver 3 digits, which makes backporting sometimes impossible.

I propose we only enforce this on new package registration. And allow old packages to be grandfathered.

We have talked about this on and off for years

tecosaur:
I don't think it would be a good hard rule to have, but I think it would be a good guideline/soft check.

I have one or two packages that are `0.x` myself at the moment, because they're just about useful enough to be registered but still very much having their API/design worked out.

In most cases though, I think it's very useful to have the three channels of information in the version number:
- Breaking changes
- New features
- Bugfixes

So perhaps a way this could be done is as an automerge check, allowing for manual intervention if the registrant insists that they've thought about it and really do want 0.x?

oxinabox:
>  So perhaps a way this could be done is as an automerge check, allowing for manual intervention if the registrant insists that they've thought about it and really do want 0.x?

I agree.
We make it an automerge check, and be generous with exceptions.
And since I propose only doing that check it at registration time, it will not be too much barrier or inconvenience.

oxinabox:
I would argue its actually he opposite.
There are so so many packages with very stable API that haven't changed in years that are pre-1.0
there is no cost for initially releasing at 1.0 and then releasing 2.0 later.

tecosaur:
To me *the* motivating issue here is that we have a bunch of `0.x` packages that release versions with new features, don't break API, but are considered a semver breaking release because they don't have access to that crucial third channel of semver information.

This results in needless comparability bound churn across the ecosystem. *This* is the problem that I see this proposal helping with.

[...]

PR #119607, opened 2024-11-17
yank StructArrays@0.7.0

aplavin:
The version was tagged as breaking but actually isn't. To fix that, it has been released as 0.6.20 afterwards (same exact content).
Meanwhile, 0.7.0 should probably be yanked to avoid further confusion – this PR.

### DIFF
--- a/S/StructArrays/Versions.toml
+++ b/S/StructArrays/Versions.toml
@@ -115,3 +115,4 @@ git-tree-sha1 = "f4dc295e983502292c4c3f951dbb4e985e35b3be"
 
 ["0.7.0"]
 git-tree-sha1 = "5a3a31c41e15a1e042d60f2f4942adccba05d3c9"
+yanked = true


goerz:
Just for the record, I strongly object to the laissez-faire attitude towards yanking that seems to be developing here. It should only be done in the most extreme emergencies. As I said on Slack:

> I’d just point out that SemVer does not require a major release to have breaking changes (only the other way around). On the other hand, yanking versions is extremely disruptive. I don’t think we should ever yank versions, except maybe in a malware situation

DilumAluthge:
Suppose that someone releases a version Foo.jl version 1.1.0,  but they actually made a breaking change (and thus they should have actually released Foo 2.0.0). In that case, if a downstream user (that has `Foo = "1"` in their `[compat]` section) installs Foo 1.1.0, there is a real chance that their code might break. Therefore, in the registry we often choose to yank Foo 1.1.0.

This is the opposite situation. Someone released a version Bar.jl version 0.7.0, but they didn't actually have any breaking changes, and thus they could have (and should have) released a Bar 0.6.x. In this case, if a downstream user installs Bar 0.7.0, their code won't actually break (because Bar 0.7.0 didn't have any breaking changes). In this case, because there is no risk of a downstream user having their code break, we do not need to yank Bar 0.7.0 in the registry.

goerz:
> Not sure I see how is yanking suitable for malware situations... Then, best would be to just remove the version so that nobody installs it anymore.

I'm not sure if I fully appreciated the specific way "yanking" is implemented, as opposed to deleting/reverting a package or release. I do agree that actual hypothetical malware should be purged from the registry in a way that makes it impossible to install.

> Suppose that someone releases a version […] 1.1.0, but they actually made a breaking change […], in the registry we often choose to yank

That's not really necessary, and arguably violates SemVer. The [SemVer specification explicitly discusses this particular situation](https://semver.org/#what-do-i-do-if-i-accidentally-release-a-backward-incompatible-change-as-a-minor-version):

> As soon as you realize that you’ve broken the Semantic Versioning spec, fix the problem and release a new minor version that corrects the problem and restores backward compatibility. Even under this circumstance, it is unacceptable to modify versioned releases. If it’s appropriate, document the offending version and inform your users of the problem so that they are aware of the offending version.

Maybe the "yanking" feature in the registry was intended as a way to implement "inform your users"? If so, that would be a good thing, but I've seen situations recently where things got screwed up because of a yanked package. If not, I still think we should almost never yank anything. There might be some situations related to wrong compat bounds where mistakes can't be repaired _within_ SemVer, but I'm not even sure about that.

goerz:
Just to conclude this: the intended behavior of yanking is described in the [General Registry README](https://github.com/JuliaRegistries/General?tab=readme-ov-file#when-is-yanking-a-release-appropriate). The way it is described there, it should generally be okay to yank "broken" releases under the conditions described there.

Note that this PR is not one of those conditions.

Generally speaking, an accidental major release that actually wasn't breaking is certainly unfortunate / confusing (in the sense of "unexpected"), but it's not a violation of SemVer. There's nothing that can be done that wouldn't violate SemVer except to just keep going from the major release.

I would also point out that StructArrays is pre-1.0, so a change from v0.6 to v0.7 is actually *not breaking* (despite `Pkg` treating it as such). [The SemVer spec is extremely clear about this](https://semver.org/#spec-item-4): Before `v1.0`, "anything can happen". Pre-1.0 packages simply do not have a "stable" API, and thus you can't "break" it. Of course, it's fine to use `v0.x` releases as "big change" (a.k.a, probably breaks code depending on this unstable API), and `v0.x.y` as "small change (a.k.a. probably does not break dependent code). But this is a social convention of top of SemVer, and a bit on the fuzzy side.

It might be worth looking at whether `StructArray` should tag a 1.0 release: [If you’re worrying a lot about backward compatibility, you should probably already be 1.0.0.](https://semver.org/#how-do-i-know-when-to-release-100). Just for good measure, I would point out that going from `v0.x` to `v1.0` is not a breaking change! That is, it's fine to tag a `v1.0` simply to indicate "the API is now stable according to SemVer". In fact, I strongly advocate for a `v1.0` release not to have any major changes compared to the last `v0.x` release, up to changes in documentation and maybe removing "deprecated" code (even though "deprecation" doesn't make sense pre-1.0, since, again, there is no such thing as a stable API pre-.1.0)

At this time, it's unclear whether `Pkg` might have a bug where instantiating an environment that has a yanked version might fail. There seems to be some anecdotal evidence, but @DilumAluthge hasn't been able to confirm it with a reproducible example. So let's assume yanking works as intended. In that case, it seems fine to use yanking as "informing your users" about very broken releases that shouldn't be installed.

This might be a moot point, and I also might be misunderstanding something, but the [*reasoning* for when yanking is necessary](https://github.com/JuliaRegistries/General?tab=readme-ov-file#when-is-yanking-a-release-appropriate) seems off to me. Specifically,

>  reverting the code changes would not be valid in a patch bump.

seems just wrong: it is never invalid to fix a bug in a patch bump. The only way to fix the bug in that example is to revert the feature, but that doesn't make it breaking/not-a-bugfix. The correct way to handle the "compat bounds too wide" example is to make _two_ bugfix releases: one that reverts the previous release (but keeps the compat bound), and then one that re-applies the previous release but also fixes the compat bound. So, it's not *necessary* to yank anything, although I certainly don't object to yanking the broken release assuming yanking behaves as described.

goerz:
> Not really sure what you mean here?

I'd really want to make a distinction (maybe an overly subtle / hairsplitting one) between what `Pkg` considers "compatible"/"incompatible" vs "non-breaking"/"breaking". I'm on board with `Pkg` treating "0.x" version as "incompatible", but that doesn't mean they're "breaking", even if we sometimes might use that word, in a handwaving manner.

>  there are lots of pre-1.0 packages with stable API

There shouldn't be! SemVer specifically says that a package does not have a stable API if it is pre-1.0. This isn't just something where we can just say "Julia uses a minor variation of SemVer". Pre-1.0 versions are a core part of the SemVer philosophy, with a specific meaning. Disregarding that is completely antithetical to SemVer.

> Such a bump can make sense when one feels the need for an extra number in the version, but otherwise it doesn't matter if the versions go 0.1, 0.2, 0.3, ..., 0.23 or 1, 2, 3, ..., 23.

I couldn't disagree more! There is a very fundamental difference between the `0.x` (unstable) development life cycle and the post-1.0 (stable) development life cycle of a project: In the former, backwards-compatibility is not a concern, in the latter, it is. If there is no difference between `0.x` minor version and major versions, then why even have `0.x` versions? Indeed, disallowing pre-1.0 registrations is precisely what people have proposed based on this misunderstanding: https://github.com/JuliaRegistries/General/issues/111019. It just completely disregards a core part of SemVer, as I elaborate on in [my comment there](https://github.com/JuliaRegistries/General/issues/111019#issuecomment-2400321877). I feel very strongly that pushing people to tag releases that are not stable as `v1.0` (or, conversely, not tagging a package that in fact has a stable API as `v1.0`) hurts the quality of the ecosystem.

It's fine for Julia to have some minor variations on top of SemVer, like making an additional distinction between `v0.x` and `v0.x.y` releases as incompatible/compatible, or allowing some additional ordering or semantics for build numbers (which I'm very much in favor of). But we should really not throw something so fundamental as stable vs non-stable releases overboard. Maybe from the perspective of `Pkg`, it's splitting hairs. From a larger perspective of how we should push people to tag their packages, or how worried they should be about some mild "SemVer-violation" when their package is still pre-1.0 (the situation in the PR), I think it's pretty fundamental.

goerz:
> If the yanking has become as strict as you say

I'm not sure it has "become" strict… I don't think it's changed (but I'm not sure about the history). I'm only reacting to how it is [currently described in the README](https://github.com/JuliaRegistries/General?tab=readme-ov-file#when-is-yanking-a-release-appropriate).

> Wasn't the motivation for yanking being able to avoid Pkg resolving new envs with an erroneously tagged release?

Not as far as I can tell (how it's currently documented). It's to (optionally) mark a release as "broken" after a patch release to fix the broken release, or to deal with the "special category of bugged releases" described in the README.

> Wrong compat bounds? Change them in the registry.

I don't think that's possible. That would break re-instantiating existing environments, and  "reproducibility" is basically _the_ core tenet of General. It would be fine in a LocalRegistry that didn't have such strict rules. Releases are basically immutable.

Quite honestly, I don't think yanking is really ever *necessary*. Even in the "special category" of compat bound issues. I think these can be fixed entirely within SemVer by making _two_ immediate bugfix releases (I might be wrong). But yanking still seems like a good way to attach a "broken" tag to a particular release. So it's probably helpful to have, as a feature.


> If v0.7.0 was just yanked here, no additional changes would be required

Does the CompatHelper bot take into account yanking (i.e., would yanking stop CompatHelper prompting peopel to update their compat bounds to `v0.7.0`)? If so, and if it's not already too late (CompatHelper has already notified everyone), then I suppose yanking would save people from having to update the compat bounds later.

But you'd still have to make `v0.8.0` your next breaking release, and then people will just be even more confused at that point. I don't see any way to avoid people needing to deal with the existence of the `v0.7.0` release *at some point*, and it seems to me the least confusing way is do it now: document it in your changelog ("Upgrade guidelines: v0.7 contained no breaking changes, just set your compat spec to `0.6,0.7`"), and move on.

goerz:
> This is not just routinely done

You seem to be right, e.g., https://github.com/JuliaRegistries/General/pull/115069

> but even planned to make more straightforward

I don't get how that doesn't break reproducibility. But that's a discussion I'll take over there…

>> But you'd still have to make v0.8.0 your next breaking release
>
> Don't see any issues with that whatsoever.

Then I guess `v0.7.0` really should be yanked. You'll have to deal with the fallout of having apparently skipped a version later on. Not what I would do. I'm pretty sure it will cause confusion down the road, but that's between you and the dependents of `StructArrays`

PR #119714, opened 2024-11-18
Migrate our manifest naming convention from `Manifest.1.x.toml` to `Manifest-v1.x.toml` (to match upstream Julia)

DilumAluthge:
Upstream Julia now (as of https://github.com/JuliaLang/julia/pull/43845) uses the naming convention `Manifest-v1.x.toml`. So let's adopt that same naming convention here.

### DIFF
--- a/.ci/Manifest-v1.10.toml
+++ b/.ci/Manifest-v1.10.toml
@@ -1,0 +1,439 @@
+# This file is machine-generated - editing it directly is not advised
+
+julia_version = "1.10.6"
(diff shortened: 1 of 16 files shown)


DilumAluthge:
cc: @IanButterworth

DilumAluthge:
In this PR, I didn't delete the old manifest files (`Manifest.1.x.toml`). I'll do that in a follow-up PR, once I confirm that our "manifest updater" CI job works with the new naming convention.

PR #120133, opened 2024-11-25
[Singular_jll] yank some breaking releases

lgoettgens:
The `404.0.707+0` release was faulty, the following ones tried to fix that, but unfortunately, they were still breaking downstream (cf https://github.com/oscar-system/Oscar.jl/issues/4341), although marked as a patch release.

https://github.com/oscar-system/Oscar.jl/issues/4341 will take care of releasing the previous working version as another new patch release, but nevertheless it would be good to get these versions out of the resolvers options.

ping @fingolfin @benlorenz @thofma


fingolfin (approved):

fingolfin:
In the meantime we also merged https://github.com/JuliaPackaging/Yggdrasil/pull/9864 which should also solve the problem, so perhaps we don't need this after all (though I was the one requesting this -- apologies to @lgoettgens :-/ ).

The background there is that from a technical POV yanking these "feels" right as these versions are really broken in various ways (the last one the least, it "only" causes doctest failures). So anyone using them will run into problems; but of course once the new JLL version lands nobody should ever get these unless they explicitly specify the version (which makes it their own fault), or if they had the misfortune of updating in the last couple days (but then running `Pkg.update()` will rescue them).

lgoettgens:
@fingolfin do we still wanna have this? If yes, let's ping the slack channel. If not, I'll just close this

fingolfin:
Let's close it

PR #120760, opened 2024-12-05
Add GitHub action that deploys GitHub Pages with redirects to package repositories

jkrumbiegel:
As discussed here https://discourse.julialang.org/t/automatic-package-links/106275 and here https://discourse.julialang.org/t/automatic-package-link-pages-have-poor-layout/123087, we currently seem to be lacking a good way to link to Julia packages when we only know the package name. There's the JuliaHub UI but there's a lot of controversy around linking to JuliaHub by default. The whole docs generation thing was a bit of a mess.

So my proposal is to just add a GitHub Pages to the General Registry repo which contains one HTML page per package. That page contains a redirect header that redirects immediately to the repo that the registry stores for that package. I've tested this workflow on my fork https://github.com/jkrumbiegel/General/ and the deployed page allows to link to package repos with links like

(Edit: changed the format to be more clear about redirection)

https://jkrumbiegel.com/General/packages/redirect_to_repo/Makie
https://jkrumbiegel.com/General/packages/redirect_to_repo/DifferentialEquations
https://jkrumbiegel.com/General/packages/redirect_to_repo/InPartS (this one will not auto-redirect due to the uncommon host)

and so on. Discourse could then change its linkifier logic to use such links. Currently the same format for this repo would be

https://juliaregistries.github.io/General/packages/redirect_to_repo/Makie
https://juliaregistries.github.io/General/packages/redirect_to_repo/DifferentialEquations
https://juliaregistries.github.io/General/packages/redirect_to_repo/InPartS

but if desired a nicer domain could be chosen so this becomes something like

https://juliaregistry.com/Makie
https://juliaregistry.com/DifferentialEquations
https://juliaregistry.com/InPartS

Currently the workflow only has manual trigger logic, but a cron trigger should be added at some reasonable interval. One deploy currently takes about 40 seconds.

cc @MasonProtter


DilumAluthge:
> Currently the same format for this repo would be
> 
> https://juliaregistries.github.io/General/Makie

Can we make the URL communicate more clearly that a redirect is happening? Maybe something like this instead?

```
https://juliaregistries.github.io/General/redirect_to_repo/Makie
```

.github/workflows/redirects_page.yml
```diff
@@ -0,0 +1,37 @@
+name: Generate and Deploy Redirects
+
+on:
+  workflow_dispatch:
```
DilumAluthge, line 4: I think it would be reasonable to have a cron job that runs once every 6 hours (in addition to the `workflow_dispatch`).

generate_redirects.jl
```diff
@@ -0,0 +1,42 @@
+using TOML
```
DilumAluthge, line 1:
```suggestion
import TOML
```

generate_redirects.jl
```diff
@@ -0,0 +1,42 @@
...
+        <!DOCTYPE html>
+        <html lang="en">
+        <head>
+            <meta charset="UTF-8">
+            <meta name="viewport" content="width=device-width, initial-scale=1.0">
+            <title>Redirecting...</title>
+            <meta http-equiv="refresh" content="0; url=$repo">
+        </head>
+        <body>
+            <p>If you are not redirected automatically, follow this <a href="$repo">link</a>.</p>
```
DilumAluthge, line 25: Make sure we have `rel=nofollow` here.

.github/workflows/redirects_page.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  build-and-deploy:
```
DilumAluthge, line 7: Can you give this job a timeout?

DilumAluthge:
Julius and I discussed this - briefly, we'll do an automatic redirect for known-hosts such as `https://github.com/*` and `https://gitlab.com/*`. For other websites, instead of automatically redirecting, we'll display the URL to the user, and the user will be asked whether or not they want to continue to the URL.

jkrumbiegel:
@DilumAluthge I've implemented all your review comments. I've added github.com, gitlab.com and codeberg.org as known hosts, the full list was very short actually:

```julia
Dict{String, Int64} with 6 entries:
  "codeberg.org"   => 4
  "git.sr.ht"      => 2
  "github.com"     => 11580
  "gitlab.com"     => 61
  "gitlab.gwdg.de" => 4
  "gitlab.emse.fr" => 1
```

Even though codeberg.org and gitlab.com have few entries they seemed more "common" than the other ones to me so I included them.

giordano:
Why this needs to be in this repository since it only needs to read and it's scheduled?  Can't this live in a different repo?

jkrumbiegel:
It could live anywhere, that's true. I thought that being associated to the official registry would fit if its used by default in discourse. But we can move it somewhere else if we'd rather leave this repo untouched.

giordano:
The thing is that these files would end up being downloaded in everyone's registries, for no reason. There are of course many other github actions workflows in this repo, but I believe they all need to operate on this repo, not just read it.

A new home for this workflow could be a repo called something like "GeneralRedirects" which would also address the comment above about including "redirects" somewhere in the URL.

DilumAluthge:
That's a good point. There's no need for us to include these HTML files in the registry that everyone downloads.

DilumAluthge:
I've created a separate repo: https://github.com/JuliaRegistries/GeneralRedirects

ericphanson:
One disadvantage of a separate repo is:

> In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days. For information on re-enabling a disabled workflow, see "[Disabling and enabling a workflow](https://docs.github.com/en/enterprise-server@3.12/actions/using-workflows/disabling-and-enabling-a-workflow#enabling-a-workflow)."

So someone will have to go and reenable that workflow every 2 months. I think if it were here, the activity in General would keep it from being disabled.

DilumAluthge:
Okay, so, crazy idea: the workflow lives in this repo (and thus never needs to be re-enabled), but we push the docs to the separate repo (GeneralRedirects) using e.g. an SSH deploy key.

So we never need to re-enable the workflow, but the actual HTML files never get saved in this repo, because we push them to the GeneralRedirects repo.

ericphanson:
could the workflow live here, but it pushes to https://github.com/JuliaRegistries/GeneralRedirects instead of this repo's github pages?

ericphanson:
jinx 😄. That seems like a good approach to me.

DilumAluthge:
@jkrumbiegel I think the implementation looks something like this. In the workflow in this repo, we use an SSH deploy key to push to the master branch of GeneralRedirects.

And then, in the GeneralRedirects repo, that's where we have a workflow that does the "upload page artifact" and "deploy pages" stuff. That workflow will only have triggers for `push` and `workflow_dispatch` (no cron trigger), and thus it will never get automatically disabled by GitHub.

I'll generate an SSH deploy key, and give it write access to GeneralRedirects, and then I'll add the corresponding private key as a GitHub Secret in this repo

ericphanson:
I'm not sure if it's easier or harder, but an alternate approach could be to push directly to the github pages branch of the other repo, and skip the artifact stuff. That is how Documenter works IIUC.

MasonProtter:
Couldn't all the HTML stuff just be generated and stored in a different branch in this repo too?

DilumAluthge:
That probably is easier.

ericphanson:
> Couldn't all the HTML stuff just be generated and stored in a different branch in this repo too?

That's what this PR does, but when the registry is cloned, that branch would be included too IIUC. So it bloats the repo and slows down non-pkg-server use.

DilumAluthge:
> Couldn't all the HTML stuff just be generated and stored in a different branch in this repo too?

Hmmm. It'll still be included in Git clones of the registry (although admittedly most people should be using the Pkg Server registry, which presumably doesn't include anything that's not on the master branch).

Although, the Pkg servers have to periodically fetch the Git version of the registry, and presumably those fetches would become slower as the repo size becomes bigger?

MasonProtter:
Shouldn't those `git fetch`es just fetch the master branch anyways though? It'd be pretty wasteful to fetch all the other branches.

jkrumbiegel:
Just to be sure, you're not saying the actual html files are going to bloat this repo right? Because this is not a `gh-pages` branch deploy but via the artifact workflow that deploys directly to pages.

DilumAluthge:
Hmmm, I think I don't understand how this workflow works. Where are these generated HTML files stored?

ericphanson:
> Just to be sure, you're not saying the actual html files are going to bloat this repo right? Because this is not a `gh-pages` branch deploy but via the artifact workflow that deploys directly to pages.

Oh, I thought it added the html files to the repo itself. If it is just the jl script and the workflow I think it’s fine to add here.

jkrumbiegel:
I think github pages has some separate servers that content needs to be uploaded to. The legacy workflow was that you'd push to gh-pages and that would trigger the internal deploy mechanism. Now, you can skip to the deploy step via the artifact system so nothing gets modified in the repo.

But of course the jl file is an addition to the repo that could be argued is unnecessary.

DilumAluthge:
Ah, yeah I was also under the impression that the HTML files would exist in the repo itself.

One YAML file (the GitHub Actions workflow file) and one Julia file isn't too bad.

MasonProtter:
@ericphanson  so given the conversation in https://github.com/JuliaDocs/Documenter.jl/issues/2614, do you think that'd be a compelling reason to not use artifacts here, or no?

ericphanson:
no, this workflow does not need persistent, so there is no issue. The problem with Documenter using it is that Documenter only builds the docs for v1 when you tag v1. Then when you tag v2, it builds the docs for v2 and adds them to the branch. So we accrete docs in the branch. When switching to artifacts we need to maintain that state in artifact-land which isn't possible since they expire (i.e. we need to download the v1 artifact when generating the v2 artifact so we can merge them and deploy them together).

Here we regenerate all HTML from scratch each deploy, so no problem. We don't need to download old artifacts.

MasonProtter:
Nice, thanks for the explanation. So what do we still need to do to make this PR happen?

ericphanson:
@giordano do you object to this PR still? Given:

- there are no checked in HTML files produced by this workflow, only the jl script and workflow file
- if we moved those to another repo, they would need to be manually re-enabled every 60 days

jkrumbiegel:
@DilumAluthge will do

giordano (dismissed):

generate_redirects.jl
```diff
@@ -0,0 +1,92 @@
...
+    end
+
+    open(package_path(name * ".html"), "w") do io
+        write(io, """
+        <!DOCTYPE html>
+        <html lang="en">
+        <head>
+            <meta charset="UTF-8">
+            <meta name="viewport" content="width=device-width, initial-scale=1.0">
+            <title>Redirecting to $name...</title>
```
ericphanson, line 46: Should this title be conditional? Since we don’t always auto redirect

generate_redirects.jl
```diff
@@ -0,0 +1,92 @@
...
+                    display: flex;
+                    justify-content: center;
+                    align-items: center;
+                    min-height: 100vh;
+                    margin: 0;
+                    font-family: Arial, sans-serif;
+                    background-color: #f9f9f9;
+                }
+                .centered-div {
+                    border: 2px solid #1E88E5;
```
ericphanson, line 59: Maybe there’s some Julia branding / colors to use?

ericphanson (dismissed):
Lgtm; I had some small comments about title/colors but those could always be refined in follow-ups and community feedback, so I’m also ok with merging as-is

.github/workflows/redirects_page.yml
```diff
@@ -0,0 +1,39 @@
+name: Generate and Deploy Redirects
+
+on:
+  workflow_dispatch:
+  schedule:
```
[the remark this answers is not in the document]
DilumAluthge: `Package.toml`, presumably?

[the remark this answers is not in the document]
giordano: The files that contain the URL, whatever they're called.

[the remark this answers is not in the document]
DilumAluthge: Seems reasonable to me. It might still be good to have a once-daily cron job (because the jobs might fail sometimes, due to the usual GitHub flakiness). But yes, building on pushes to master that modify a `Package.toml` file seems like it would be a good approach.

generate_redirects.jl
jkrumbiegel: ChatGPT thought this was Julia blue but it actually wasn't 😅 I've made it Julia purple now which is the most recognizable I think

.github/workflows/redirects_page.yml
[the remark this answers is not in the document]
jkrumbiegel: I've replaced the cron job with a path-filtered trigger on push to master. I don't think an additional cron job is necessary, the action is unlikely to fail and the worst that can happen is that an old repo link stays active a little longer or a new package isn't linked for a short time.

generate_redirects.jl
giordano: You don't need to ask ChatGPT, colours are defined at https://github.com/JuliaLang/julia-logo-graphics

jkrumbiegel: hehe yeah I know, that was mostly laziness for getting the centered div css :)

jkrumbiegel:
Ok I think I've addressed all review comments. The thing now has a Julia-purple border instead of the random blue :)

grafik

ericphanson:
Great, let's go ahead and merge this and we can make tweaks if needed in followups. We should keep an eye on how often it runs, I think basically every PR will trigger this so we might want to switch it back to a cron to avoid wasting compute. But I think we can see how it goes first.

generate_redirects.jl
fredrikekre, line 1: This file should probably not be at the top level but perhaps in the `.ci` folder?

jkrumbiegel:
I checked the last couple merges on master and none of those had Package.toml changes. It's mostly versions and compat. So it won't trigger every time at least. And it's pretty quick.

jkrumbiegel:
Oh and some admin needs to actually enable the github pages build via actions from the repo settings, otherwise the workflow will fail.

ericphanson:
ah, yeah: https://github.com/JuliaRegistries/General/actions/runs/12254152092

@DilumAluthge can you enable github pages when you get a chance?

and @jkrumbiegel can you make a PR to move the redirect script to the `.ci` folder? https://github.com/JuliaRegistries/General/pull/120760#discussion_r1877813675

fredrikekre:
I enabled it but the rerun failed because now there are two artifacts? One for the original run and one for the rerun...

ericphanson:
awesome! these seem to work now:

https://juliaregistries.github.io/General/packages/redirect_to_repo/Makie
https://juliaregistries.github.io/General/packages/redirect_to_repo/DifferentialEquations
https://juliaregistries.github.io/General/packages/redirect_to_repo/InPartS

fredrikekre:
Perhaps it would be neat with some intermediate pages so that for example https://juliaregistries/github.io doesn't 404?

MasonProtter:
It'd also be good if there was a fallback page for e.g. https://juliaregistries.github.io/General/packages/redirect_to_repo/xyz that tells users something like 

> xyz.jl is not a registered package in the julia General Registry. Would you like to search [Github](https://github.com/search?q=xyz.jl&type=repositories), [Gitlab](https://about.gitlab.com/search/?searchText=xyz.jl), [Google](https://www.google.com/search?q=xyz.jl), or [DuckDuckGo](https://duckduckgo.com/?t=h_&q=xyz.jl)?

generate_redirects.jl
ericphanson: - #121111

tecosaur:
Just a quick question, are dedicated redirect pages needed instead of directly serving 302 redirect responses?

jkrumbiegel:
I don't think you can do that without setting up your own server. This was explicitly a low tech solution where that's not necessary :)

giordano:
> I think basically every PR will trigger this so we might want to switch it back to a cron to avoid wasting compute.

Over last 10k commits to this repository, only about 7% (711 to be more precise) touched a `Project.toml` file, that doesn't look like a lot to me.  This is more frequent than running a schedule job every 6 hours (last 10k commits go back to 2024-07-07, which was 156 days ago, 3 jobs a day would be 468 in total), but more accurate.

<details>
<summary>Script used to count the number of commits</summary>

```bash
#!/bin/bash

REGISTRY_DIR="${HOME}/.julia/registries/General"
COMMITS=$(git -C "${REGISTRY_DIR}" log --oneline --format=%H -10000)

i=0
for commit in ${COMMITS}; do
    if [[ $(git -C "${REGISTRY_DIR}" diff-tree --diff-filter=d --name-only -r "${commit}^" "${commit}") == *Package.toml* ]]; then
        ((i+=1))
    fi
done
echo "${i} commits touched a Package.toml file"
```

This took about a minute to run on my laptop, hold tight.
</details>

DilumAluthge:
This is nice. Thank you @jkrumbiegel!

PR #120786, opened 2024-12-06
New packages: PRASCore, PRASFiles, PRASCapacityCredits, PRAS v0.7.0

GordStephen:
This is a manual pull request, prepared with LocalRegistry.jl, to simultaneously register v0.7.0 of four packages from NREL's Probabilistic Resource Adequacy Suite (PRAS). The four packages are PRASCore.jl, PRASFiles.jl, PRASCapacityCredits.jl, and PRAS.jl, a convenience metapackage that loads all the others. All four packages reside in https://github.com/NREL/PRAS

For convenience I'll try to proactively talk through some of the automerge checks:

- About the package name - yes, we realize this probably isn't going to be looked on too favorably, particularly PRAS.jl which fails multiple automerge rules. PRAS as a brand exists outside of the Julia ecosystem (e.g. https://www.nrel.gov/analysis/pras.html) and we're hoping we can get an exception on this. We received feedback from @goerz and @gdalle in https://github.com/JuliaRegistries/General/pull/120172 that this may be possible, and are hoping folks here agree?

- About the repository name - while today PRAS is exclusively a Julia project, we're actually in the middle of a rewrite and our monorepo will be hosting Rust, Julia, and Python code in the near future. So we don't think having a repo name that ends in .jl would be appropriate.

- About the version number - while 0.7.0 is the first version of PRAS submitted to General, it is not the first version of PRAS. v0.6 and earlier were registered in NREL's public registry (https://github.com/NREL/JuliaRegistry).

- All four packages are MIT licensed and a copy of the license is included in each of the respective subdirectories of the monorepo.

Happy to answer any further questions or take feedback if there are things here we should be doing differently!


goerz:
Given the maturity of the project, I don’t see any problem with the name `PRAS` and would be happy to see this merged.

You will probably have to go on Slack and post a message in the `#pgk-registration` channel to find someone with merge permissions.

GordStephen:
Per guidance on Slack, we're submitting these individually via Registrator with the understanding that this PR will count as the three days waiting period.

PR #121111, opened 2024-12-10
Move `generate_redirects.jl` to `.ci`

jkrumbiegel:


.ci/generate_redirects.jl
```diff
@@ -90,10 +90,12 @@ function create_redirect_page(; name, path)
...
-    mkpath(package_path())
-    packages_info = get_packages_info()
-    for (; name, path) in packages_info
-        create_redirect_page(; name, path)
+    cd(joinpath(@__DIR__, "..")) do
+        mkpath(package_path())
+        packages_info = get_packages_info()
+        for (; name, path) in packages_info
+            create_redirect_page(; name, path)
+        end
```
ericphanson, line 98: I think no change is needed here, since the cwd hasn't changed. That being said this is maybe a bit more robust since it doesn't rely on the cwd

ericphanson (approved):

jkrumbiegel:
for some reason I have the commits from the other branch in here, too, but squash merge will take care of that

PR #121137, opened 2024-12-10
Redirect pages: Add 404 pages

jkrumbiegel:
Adds a special error mode if the path matches the required format but the package is wrong:

grafik

And a generic 404 page otherwise

grafik

### DIFF
--- a/.ci/generate_redirects.jl
+++ b/.ci/generate_redirects.jl
@@ -27,6 +27,37 @@ function package_path(args...)
     joinpath("webroot", "packages", "redirect_to_repo", args...)
 end
 
+function style_block(backgroundcolor, color)
+    """
+    <style>
+        body {
+            display: flex;
+            justify-content: center;
+            align-items: center;
+            min-height: 100vh;
+            margin: 0;
+            font-family: Arial, sans-serif;
+            background-color: #f9f9f9;
+        }
+        .centered-div {
+            border: 3px solid $color;
+            background-color: $backgroundcolor;
+            border-radius: 10px;
+            padding: 20px;
+            margin: 20px;
+            text-align: center;
+            color: #333;
+            max-width: 30em;
+            word-wrap: break-word;
+        }
+        .centered-div a {
+            color: $color;
+            font-weight: bold;
+        }
+    </style>
+    """
+end
+
 function create_redirect_page(; name, path)
     repo = get_repo(path)
     host = get_host(repo)
@@ -52,32 +83,7 @@ function create_redirect_page(; name, path)
             <meta name="viewport" content="width=device-width, initial-scale=1.0">
             <title>$title</title>
             $meta_redirection
-            <style>
-                body {
-                    display: flex;
-                    justify-content: center;
-                    align-items: center;
-                    min-height: 100vh;
-                    margin: 0;
-                    font-family: Arial, sans-serif;
-                    background-color: #f9f9f9;
-                }
-                .centered-div {
-                    border: 3px solid #9558B2;
-                    background-color: #f8e9ff;
-                    border-radius: 10px;
-                    padding: 20px;
-                    margin: 20px;
-                    text-align: center;
-                    color: #333;
-                    max-width: 30em;
-                    word-wrap: break-word;
-                }
-                .centered-div a {
-                    color: #9558B2;
-                    font-weight: bold;
-                }
-            </style>
+            $(style_block("#f8e9ff", "#9558B2"))
         </head>
         <body>
             <div class="centered-div">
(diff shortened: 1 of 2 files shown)


DilumAluthge:
I don't think all those links to search engines are necessary. I think the "no package exists" message is fine, and then maybe just a link to the JuliaLang.org homepage.

DilumAluthge:
And for the generic 404 page, maybe a link to the JuliaLang.org homepage would be nice.

jkrumbiegel:
I just implemented what @MasonProtter suggested as two other people also liked that, I have no strong preference regarding the exact content of the 404 pages

.ci/generate_redirects.jl
```diff
@@ -89,13 +95,60 @@ function create_redirect_page(; name, path)
...
+            const match = urlPath.match(pattern);
+
+            if (match) {
+                const packageName = match[1]; // Extract the package name
+                const messageElement = document.querySelector(".centered-div p");
+                // Update the paragraph text
+                messageElement.innerHTML = `\
+                    There is no package \${packageName} registered in the Julia General Registry.\
+                    <br><br>\
+                    Would you like to try searching <a href="https://github.com/search?q=\${packageName}.jl&type=repositories">Github</a>, \
```
DilumAluthge, line 131:
```suggestion
                    Would you like to try searching <a href="https://github.com/search?q=\${packageName}.jl&type=repositories">GitHub</a>, \
```

.ci/generate_redirects.jl
```diff
@@ -89,13 +95,60 @@ function create_redirect_page(; name, path)
...
+                    <a href="https://about.gitlab.com/search/?searchText=\${packageName}.jl">Gitlab</a>, \
```
DilumAluthge, line 132:
```suggestion
                    <a href="https://about.gitlab.com/search/?searchText=\${packageName}.jl">GitLab</a>, \
```

.ci/generate_redirects.jl
```diff
@@ -89,13 +95,60 @@ function create_redirect_page(; name, path)
...
+
+            // Check if the URL matches the pattern
+            const match = urlPath.match(pattern);
+
+            if (match) {
+                const packageName = match[1]; // Extract the package name
+                const messageElement = document.querySelector(".centered-div p");
+                // Update the paragraph text
+                messageElement.innerHTML = `\
+                    There is no package \${packageName} registered in the Julia General Registry.\
```
DilumAluthge, line 129: Would it be better to say `Example.jl` instead of just `Example`?

```suggestion
                    There is no package \${packageName}.jl registered in the Julia General Registry.\
```

DilumAluthge:
I thought about it a bit more. I think I'm fine with keeping the search engine links for now. We can re-evaluate and remove later if we change our minds.

DilumAluthge:
What does the user experience look like if they have JavaScript disabled or blocked?

giordano:
This PR reminded me that probably https://github.com/JuliaRegistries/General/blob/75b1eaca50e544716180365df13bc8778315337f/.github/workflows/redirects_page.yml#L8 should include also `.github/workflows/redirects_page.yml` and `.ci/generate_redirects.jl`

MasonProtter:
Personally, I think the search engine links are very helpful, because not all julia packages are registered. 

This gives an easy way for me to write Foo.jl in Discourse / Zulip and then when that gets auto-linkified, users can search for it if they desire.

.ci/generate_redirects.jl
```diff
@@ -89,13 +95,60 @@ function create_redirect_page(; name, path)
...
+                    <br><br>\
+                    Would you like to try searching <a href="https://github.com/search?q=\${packageName}.jl&type=repositories">Github</a>, \
+                    <a href="https://about.gitlab.com/search/?searchText=\${packageName}.jl">Gitlab</a>, \
+                    <a href="https://www.google.com/search?q=\${packageName}.jl">Google</a>, \
+                    or <a href="https://duckduckgo.com/?t=h_&q=\${packageName}.jl">DuckDuckGo</a> for it?`;
```
DilumAluthge, line 134: What does the `t=h_` do?

DilumAluthge: Looks like it's used by partners/affiliates of DDG? https://duckduckgo.com/duckduckgo-help-pages/privacy/t/

Seems like we should remove the `t=` parameter?

DilumAluthge:
```suggestion
                    or <a href="https://duckduckgo.com/?q=\${packageName}.jl">DuckDuckGo</a> for it?`;
```

MasonProtter: ah sorry about that. The URL came from me copy-pasting from Firefox.

MasonProtter:
So all that needs to done is change the `Github` -> `GitHub` and `Gitlab` -> `GitLab`?

DilumAluthge:
I think it might also be nice to change "no package Example" to "no package Example.jl" in the user-facing message.

Because the user will likely want to do a web search for the package, and doing a web search for "Example.jl" will likely be more useful than doing a web search for "Example".

jkrumbiegel:
Are the jlls also called .jl? By the way, the search links already include the jl

DilumAluthge:
Yeah, JLLs are named e.g. `Foo_jll.jl`. I think including the `.jl` in the visible error message would be good, in case a user just wants to copy-paste their search, or manually type it (if they prefer a search engine other than one of the ones that we link).

DilumAluthge (dismissed):

DilumAluthge:
@ericphanson @giordano Any other thoughts here?

giordano:
https://github.com/JuliaRegistries/General/pull/121137#issuecomment-2532993111

DilumAluthge (approved):

giordano (approved):

PR #65999, opened 2022-08-10
make whitespace consistent in the registry

KristofferC:
Pulled out of https://github.com/JuliaRegistries/General/pull/65997

PR #121256, opened 2024-12-12
Add explicit compat bounds for the Distributed stdlib

JamesWrigley:
Unfortunately this also includes a bunch of extraneous whitespace changes. In total there are 512 affected packages.

I believe all the previous steps have already been done: https://github.com/JuliaLang/julia/issues/50697
Requires this PR to reproduce: https://github.com/KristofferC/StdlibRegistryCompatUpdater/pull/2

CC @vchuravy


KristofferC:
Would it be possible to have a PR that first normalizes all the whitespace in the registry (which would be easy to review) and then have this on top of it so that it wouldn't mix actual changes with whitespace only?

JamesWrigley:
I can make the PR but you'll have to merge it pretty quickly to avoid merge conflicts :stuck_out_tongue:

fredrikekre:
I thought there were tests that verify that files roundtrip after parsing and rewriting. Perhaps we are just testing that the parsed output is equal to the parsed-written-reparsed output?

Is this a change in TOML.jl?

KristofferC:
Hm, the changes here seems to undo https://github.com/JuliaRegistries/General/pull/65999.

JamesWrigley:
Is there like a preference for formatting that I'm missing or something?

JamesWrigley:
Running the stdlib updater on 1.10 seems to have removed the whitespace changes, but now there's 500 modified files instead of 512 :thinking:

JamesWrigley:
These are the packages that were skipped when running in 1.10:
```julia-repl
julia> setdiff(dot_11, dot_10)
12-element Vector{SubString{String}}:
 "BLAKJac"
 "ClimaCalibrate"
 "CloudClusters"
 "EwaldSummations"
 "GeoStatsProcesses"
 "LibraryAugmentedSymbolicRegression"
 "OpenGeoSysUncertaintyQuantification"
 "Oscar"
 "ParametricProcesses"
 "PulseInputDDM"
 "TableTransforms"
 "TopologicalNumbers"
```

They all already have compat bounds and the difference between the diffs looks to be purely whitespace changes, e.g.:
```diff
modified   P/ParametricProcesses/Compat.toml
@@ -1,3 +1,3 @@
 [0]
-Distributed = "1.9.2-1"
-julia = "1.5.0-1"
+Distributed = "1.9.2 - 1"
+julia = "1.5.0 - 1"
```

So I think the new diff from 1.10 can be trusted.

JamesWrigley:
Rebased to fix merge conflicts. I believe this is ready for review.

PR #121577, opened 2024-12-17
Enable release note requirement for breaking releases

ericphanson:
turns on the check from https://github.com/JuliaRegistries/RegistryCI.jl/pull/586


DilumAluthge (approved):

IanButterworth:
I think this also needs a release including https://github.com/JuliaRegistries/RegistryCI.jl/pull/587 and a Manifest bump here?

ericphanson:
that one is merged and updated here, so I think we're good to go

PR #121870, opened 2024-12-22
Restrict StaticArrays.jl @v1.5.16- to Julia v1.10-

fhagemann:
Repeating #116935 :

> Following up on https://github.com/JuliaRegistries/General/pull/86950#issuecomment-2403361516:
In https://github.com/JuliaRegistries/General/pull/86950, the julia compat was restricted to ≤ 1.10 for StaticArrays@1-1.5.
However, the code which errors was only removed in 1.5.17, see [here](https://github.com/JuliaArrays/StaticArrays.jl/pull/1136/commits/fdf353022cde23b049ad59e4b7710995458276f4).

This change was already merged in #116935, but reverted in #117626 because of tests failing in Pkg.jl (which were pulling `StaticArrays@1.5.0`).

I proposed a change in the Pkg tests  in https://github.com/JuliaLang/Pkg.jl/pull/4077, where now `StaticArrays@1.5.20` is pulled instead of `StaticArrays@1.5.0`.

Now, remerging #116935 should be fine.


fhagemann:
Maybe someone could have a look at this. This PR was originally approved, reverted because it was breaking the `Pkg` tests. But the `Pkg` tests have been fixed in the mean time.

fhagemann:
@carstenbauer @giordano :innocent:

PR #123382, opened 2025-01-21
Restrict CURL compat

staticfloat:
CURL doesn't work well with different versions of LibCURL, so let's restrict those more.

I have been testing this with the following script:

```
using BinaryBuilderSources, Base.BinaryPlatforms
using BinaryBuilderSources: PackageSpec
using Pkg.Types: VersionSpec

platform = HostPlatform()
jlls = [
    JLLSource(PackageSpec(;name="CURL_jll", version=v"8.11.0+0"), platform),
    JLLSource(PackageSpec(;name="LibCURL_jll", version=v"8.9.0+0"), platform),
]

mktempdir() do prefix
    prepare(jlls)
    deploy(jlls, prefix)
    cd(joinpath(prefix, "bin")) do
        run(`curl -L https://cflo.at`)
    end
end
```

### DIFF
--- a/jll/C/CURL_jll/Compat.toml
+++ b/jll/C/CURL_jll/Compat.toml
@@ -21,10 +21,10 @@ Libdl = ["0.0.0", "1"]
 OpenSSL_jll = "3.0.15-3"
 
 ["8.11.0"]
-LibCURL_jll = "8.11.0-8"
+LibCURL_jll = "8.11.0-8.11"
 
 ["8.11.1-8"]
-LibCURL_jll = "8.11.1-8"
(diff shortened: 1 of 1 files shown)


staticfloat:
@giordano I think we need to tighten [this up](https://github.com/JuliaPackaging/Yggdrasil/blob/master/L/LibCURL/common.jl#L144), but I honestly forget what the correct PkgSpec bound is; we want to make sure that you're on the same major.minor release.

giordano:
It's unfortunate, but makes sense to me if that's necessary.

PR #124947, opened 2025-02-12
Update instructions for moving a subdir package

omus:
Tweaks the instructions after working through some issues discussed in Slack: https://julialang.slack.com/archives/C6M4DQA5P/p1739309667085209


CONTRIBUTING.md
```diff
@@ -111,12 +111,16 @@ function check_all_found(table)
 end
 
 function check_packages_versions(pkg_names, repo_url; registry_uuid=GENERAL_UUID, verbose=true, throw=true)
-    dir = mktempdir()
-    run(`$(git()) clone $(repo_url) $dir`)
+    if isdir(repo_url)
+        dir = repo_url
```
omus, line 115: Allows local paths to be tested before pushing

CONTRIBUTING.md
```diff
@@ -111,12 +111,16 @@ function check_all_found(table)
...
+    else
+        dir = mktempdir()
+        run(`$(git()) clone $(repo_url) $dir`)
+    end
 
     registry = only(filter!(r -> r.uuid == registry_uuid, reachable_registries()))
 
-    table = @NamedTuple{pkg_name::String, version::VersionNumber, found::Bool}[]
+    table = @NamedTuple{pkg_name::String, version::VersionNumber, found::Bool, tree_sha::Base.SHA1}[]
```
omus, line 123: For my personal use case my monorepo didn't have tags created for the various releases. I used the returned table from this function to lookup commit SHAs based upon the tree SHAs and create those version tags

README.md
```diff
@@ -247,9 +247,10 @@ Registrator will complain if the second step is skipped.
...
+
+1. Follow Github's documentation on [splitting a subfolder out into a new repository][github-subfolder] an set the subdirectory package to be the new root folder of the new repository (most of these steps are still applicable for packages not hosted on GitHub). In order to avoid having to modify the package's `git-tree-sha1` entries in the registry it is important that subdirectory package's contents remain identical for the registered releases. For packages using GitHub actions this means avoiding transferring the `.github/workflows` directory as doing so would modify the root tree SHA.
```
omus, line 251: Attempted to the extra context required for users to understand what restrictions they have when moving repository content.

ericphanson: I don't quite understand the `.github` comment, can you explain more?

omus: Using a command like this wouldn't work as it causes the tree SHA to differ:
```
git filter-repo --subdirectory-filter CurrentPackageName.jl --path .github/workflows
```
as the package directory (now the root) will contain content it previously did not.

ericphanson: hm, that makes sense, I think if i read this I would be more confused since I would not consider using `--path` to additional transfer more stuff. What about
```suggestion
1. Follow Github's documentation on [splitting a subfolder out into a new repository][github-subfolder] an set the subdirectory package to be the new root folder of the new repository (most of these steps are still applicable for packages not hosted on GitHub). In order to avoid having to modify the package's `git-tree-sha1` entries in the registry it is important that subdirectory package's contents remain identical for the registered releases. This means one should not use `git-filter-repo`'s `--path` functionality to transfer additional top-level directories (like `.github/workflows`) when transferring the history, as this would modify root tree SHA.
```

omus: I've incorporated your suggestions as well as some additional ones. I was attempting to avoid specifically mentioning flags for `git-repo-filter` but I think that is just leading to less understandable instructions.

PR #125642, opened 2025-02-22
Package moved to JuliaPlots

ma-laforge:
Not sure what is needed to finalize the switch.
Let me know what else is required.

Thanks

### DIFF
--- a/N/NumericIO/Package.toml
+++ b/N/NumericIO/Package.toml
@@ -1,3 +1,3 @@
 name = "NumericIO"
 uuid = "6c575b1c-77cb-5640-a5dc-a54116c90507"
-repo = "https://github.com/ma-laforge/NumericIO.jl.git"
+repo = "https://github.com/JuliaPlots/NumericIO.jl.git"


giordano:
As a general remark, after transferring a repo to a different owner/organisation it isn't a great idea to immediately fork it back to the previous location because you break the automatic redirect GitHub creates

ma-laforge:
Right. I had a strange feeling forking back was bad given Github's redirection system.

@giordano : What should I do? I created the fork in order to make fixes for a user.
- Do I need to create a new user just to fork in a safer way?
- Could I have just uploaded my pull requests directly on the JuliaPlots repo without going through the intermediary of my personal fork? (Seems like I have write access to the repo now that I'm part of the group).
- Could/should I have forked the repo under a different name?
- Is what I have done irreversible now? Will things be ok if I delete my fork?

***And of course: Do we have a document explaining best practices when one transfers a repo to a different owner?***

Your help is greatly appreciated.

giordano:
> Could I have just uploaded my pull requests directly on the JuliaPlots repo without going through the intermediary of my personal fork? (Seems like I have write access to the repo now that I'm part of the group).

You must be a member of an organisation to transfer a repo there in the first place. I tend to alwaysffork repos I contribute to, but my usual approach when I move a repo to an org is not to fork it: it's still "my" original repo.

> Is what I have done irreversible now?

Yes.

> Will things be ok if I delete my fork?

No.

To be clear, it's not a huge deal, but keeping the redirect around for some time smoothens transitions.

PR #125971, opened 2025-02-27
Manual: Fix link to naming rules

jpthiele:
The 'guidelines' have been renamed to 'rules' in the documentation.
As I think this was intentional, this fixes the link and also uses 'rules' to stay consistent.


jpthiele:
Since this PR is about the documentation of the general registry the bot has a hard time auto-merging, so it needs manual merging I guess.

PR #125988, opened 2025-02-27
Update incorrect compat bounds for MPSKit 0.3.0

lkdvos:
This old version of MPSKit does not have a correct upper bound on its compat entries.
For example, it lists TensorKit v1 as compatible while this has not even been released.
This was already fixed in MPSKit 0.3.1, but every now and again this old version pops up again.
This should hopefully correct that.


giordano:
Would it be possible to get correct compat bounds rather than yanking the version?

lkdvos:
Apologies this has taken me a while, I wasn't sure how to achieve this, but I think this should now be correctly implemented. (I would appreciate if anyone could double-check this though, since I've never manually had to tweak a registry before.)

Concretely, I removed the (loose) compat bounds that were `"0.x-1"`, and set them based on the compat bounds that followed immediately after, as can be verified from the `0.3.1` compat bounds of MPSKit.

Feel free to let me know if I did something wrong or still have to change something.

giordano:
I don't know what are actual constraints, but syntax is correct (CI is happy about it).  Shall we merge?

lkdvos:
I am quite sure constraints are fine, so if syntax is good this is good to go for me.

PR #126194, opened 2025-03-03
[ITK_jll] Restrict compat of HDF5_jll dep to 1.14.3

divingdiv:
HDF5_jll v1.14.5 broke the ABI compared to previous versions, so we must reflect the compatibility restrictions in the registry. Ref: https://github.com/JuliaPackaging/Yggdrasil/pull/10347#issuecomment-2662923973.


divital-coder:
This is out of sync again, since new packages have been added in the last 2 hours, shall i open another?

PR #128407, opened 2025-04-04
Register CCM_analysis_Julia package

chaseU2:
- Package name: CCM_analysis_Julia
- UUID: 2d4db21e-b325-4375-9406-ee7a0d29fe22
- Repo: https://github.com/chaseU2/CCM_analysis_Julia.git


StefanKarpinski:
What is going on here? Did you make this PR by hand? Also, very unjulian name.

StefanKarpinski:
Several other issues:
1. The package name registered here does not match the package's actual name, which is `CCMAnalysis`.
2. The repo does not have the standard name for the package, which would be `CMMAnalysis.jl`.
3. This appears to be for personal use, not something that would be broadly useful to people.

PR #128480, opened 2025-04-04
yank ProgressMeter 1.10.3

MarcMush:


IanButterworth:
As I said here, I don't think a yank is needed here. 

Also when opening a yank clear explanation of why it's needed should be provided. 

A yank is typically reserved for a  security matter or a compat issues that can only be fixed by revert.

PR #128961, opened 2025-04-12
Cap compatibility of other packages depending on XML2_jll

nhz2:
XML2_jll v2.14 contains breaking API changes by removing some functionality, so this PR adjusts the compat for JuliaIO packages `EzXML` and `LightXML` as well as `Genie`, `at_spi2_atk_jll`, and `XSLT`. Ref: https://github.com/JuliaRegistries/General/pull/128943 and https://github.com/JuliaPackaging/Yggdrasil/pull/10965

CC @VEZY @hhaensel @essenciary


danielwe:
I found three more packages that have XML2_jll in their Deps.toml but weren't addressed in #128943 or here:
* XSLT
  - https://github.com/JuliaRegistries/General/blob/f0c70cc25537601d02e420eef563a7155a65b12c/X/XSLT/Compat.toml
* Genie
  - https://github.com/JuliaRegistries/General/blob/f0c70cc25537601d02e420eef563a7155a65b12c/G/Genie/Compat.toml
* at_spi2_atk_jll
  - https://github.com/JuliaRegistries/General/blob/f0c70cc25537601d02e420eef563a7155a65b12c/jll/A/at_spi2_atk_jll/Compat.toml

Do you want to add them to this PR as well? That way we don't have to add a third PR addressing the same issue

hhaensel:
Stumbled across these changes yesterday when our tests didn't pass any longer. But the failing tests were in fact shortcomings of the old version, e.g. not parsing the '@' character or adding <p></p> around the body.
So, infact, all our tests passed after correcting above mentioned issues.

hhaensel:
This shouldn't replace thorough testing, though, so thanks a lot for your action.
That's really a good service 👏

giordano:
> XML2_jll v2.14 contains breaking API changes

Reference?

danielwe:
https://gitlab.gnome.org/GNOME/libxml2/-/blob/v2.14.1/NEWS#L24
> Binary compatibility is restricted to versions 2.14 or newer.

As discussed in https://github.com/JuliaPackaging/Yggdrasil/pull/10965

giordano:
ABI != API.

giordano:
What was discussed in Yggdrasil was the change of the soname, which affects the ABI of the library, and only matters of JLLs, julia packages calling into the library wouldn't care about the change of the soname only (reason why I only fix the compat of the JLLs in the first place).  I'm asking for evidence of actual breakage for Julia packages.

nhz2:
https://github.com/GNOME/libxml2/blob/master/NEWS contains a list of features and functions that have been removed in 2.14.

> Metadata about the HTML4 content model was removed from the htmlElemDesc
> struct and related functions were deprecated.
> 
> The FTP module and related functions were removed.
> 
> Support for the range and point extensions of the xpointer() scheme
> was removed. The rest of the XPointer implementation isn't affected.
> The xpointer() scheme now behaves like the xpath1() scheme.
> 
> Several legacy symbols and the functions in xmlunicode.h were removed.
> 
> ELF version information was removed.
> 
> The shell was moved from libxml2 to xmllint. Several related functions
> are no longer available.

Tests for EzXML are also now failing. I think it makes sense to restrict compat and allow package maintainers to update when they can find the time to check how these breaking changes will affect their packages.

From the changelog, it also appears that 2.15 will remove more features, so maintainers may want to wait until the libxml API stabilizes before updating from 2.13, or stick with 2.13 if those features are needed.

giordano:
Ok, thanks, that's the kind of reference I was asking for

nhz2:
@hhaensel @essenciary You should have access to EzXML.jl, which hopefully makes it easier for you to update the compatibility or fix other issues.

PR #129058, opened 2025-04-14
Cap Julia compat for Pluto.jl

danielwe:
Julia 1.11.5 changed an internal Pkg API that Pluto has relied on since v0.15.0, so existing releases fail to precompile or load with the new Julia.

There's a fix in the pipeline, so I'm guessing a new compatible Pluto version will be released soon: fonsp/Pluto.jl#3180.

To eliminate the chance of environments resolving to an incompatible Pluto-Julia combination, even after the next release, it seems like capping the Julia compat for existing versions is the right thing to do.

@fonsp perhaps you can confirm whether this is desirable from your side? Registry maintainers, apologies if this is not how things are supposed to be done. Just trying to be helpful


fonsp:
Ah the 1.11.5 release happened faster than I expected! I will just release the new Pluto version now.

fonsp:
You can close this, https://github.com/JuliaRegistries/General/pull/129089 :)

fonsp:
Thanks though! If you're interested in making other Pluto contributions let me know! Take a look at our "Good first issues", there are some Julia-only projects and some Julia+JavaScript ones.

P/Pluto/Compat.toml
```diff
@@ -142,7 +142,7 @@ RelocatableFolders = "0.1-0.3"
 URIs = "1.3.0-1"
 
 ["0.20-0"]
-julia = "1.10.0-1"
+julia = "1.10.0-1.11.4"
```
danielwe, line 145: Need to update this to remove the cap for >=0.20.6

giordano:
>  I hesitate to tag the overextended @giordano, but it would be great to hear from a registry maintainer whether doing this is appropriate?

I'd appreciate asking generally for help in #pkg-registration on Slack.  I don't like being specifically pinged, I'm not on call.

danielwe:
I apologize. #pkg-registration is clearly mentioned in the readme. No excuse for not realizing that that's where to contact maintainers.

danielwe:
I don't think so. The erroring code was only introduced in v0.15.0 with fonsp/Pluto.jl#844

giordano:
Ok, thanks for this!  I'm personally in general in favour of fixing the registry when there known and basically unavoidable incompatibilities that would only lead to bad user experience. I think there's no point in letting the package manager give users broken environments.

fonsp:
Thanks for helping! But I'm a bit surprised that this got merged without approval from Pluto devs. 

Pluto 0.14 (from 2021) and before won't work on Julia 1.11.5 either. But I don't think there is much reason to install old Pluto versions on new Julia versions, so hunting down all the past compatibilities is probably not worth the time.

fonsp:
True! No worries, thanks for helping out!!

PR #130112, opened 2025-05-01
Move `AutoMerge-pr-label` into a separate job, to prevent label events from overriding the AutoMerge CI logs with a "skipped" job

DilumAluthge:


DilumAluthge:
Before this PR: `AutoMerge-regular` and `AutoMerge-label` are a single job, which we simply call `AutoMerge` prior to this PR. Our conditional (`if:`) logic for this single job introduces a race condition, and as a result, the single `AutoMerge` job shows up as "skipped", which makes it very difficult for package authors to find the correct CI logs.

After this PR: `AutoMerge-regular` and `AutoMerge-label` are two different jobs. As you can see from looking at this PR, `AutoMerge-regular` has a green checkmark and a link to the CI logs, while `AutoMerge-label` shows up as "skipped". The "skipped" status of `AutoMerge-label` does not affect `AutoMerge-regular`, which is what we want.

.github/workflows/automerge.yml
```diff
@@ -22,16 +22,110 @@ env:
...
+#    This only includes PR label runs.
+#    It is important to keep this job separate.
+#    If we don't keep this job separate, then any label action will override the main/regular
+#    AutoMerge job with a "skipped" AutoMerge job, thus making it very difficult for PR authors
+#    to view the CI logs (which they need to do if they e.g. want to figure out why `import MyPackageName`
+#    failed during the AutoMerge run.)
+# 3. AutoMerge-stopwatch
 jobs:
-  AutoMerge:
+  AutoMerge-regular:
```
ericphanson, line 37: could we reduce duplication by setting the job name dynamically based on the trigger? https://github.blog/changelog/2022-09-26-github-actions-dynamic-names-for-workflow-runs/

DilumAluthge: Alright, I tried setting the job name dynamically: https://github.com/JuliaRegistries/General/pull/130157

Unfortunately, as can be seen in that PR, GitHub will still overwrite the jobs with each other. The synchronize/close/reopen job will override the label job, and the label job will subsequently overwrite the synchronize/close/reopen job (with a "skipped" status).

ericphanson: Ah ok, thanks for looking into it. That’s too bad, let’s go with this approach then

ericphanson (approved):

PR #130304, opened 2025-05-05
delete Compiler.jl versions v0.0.1~v0.0.3 and register v0.0.0 only

aviatesk:
The management policy for the Compiler.jl stdlib is being changed, so this modification is required accordingly.
For more details, see https://github.com/JuliaLang/julia/pull/57520#issuecomment-2811922086.


GunnarFarneback:
I don't have enough insights about what could or should be done in this case but I will object that the title is misleading. This is not yanking but deletion. Yanking is a specific term which means adding a `yanked = true` field in `Versions.toml`, which means that the version can still be instantiated but no longer is considered by the Pkg resolver.

aviatesk:
Thanks, I changed the title.

DilumAluthge:
I discussed this on Slack with @aviatesk and Kristoffer.

PR #131938, opened 2025-05-29
update the tree hash of Compiler.jl stdlib v0.1.0

aviatesk:
I have modified the implementation of the Compiler.jl stdlib so that it can be installed on v1.10, and updated the tree hash for installation from General.

The Compiler.jl stdlib follows a special versioning policy and, for various reasons, is currently operated to have only v0.1.0 version. Therefore, instead of releasing a patch version, we are making a special accommodation by updating the tree hash on the General side.

I apologize for any recent inconvenience regarding the Compiler.jl stdlib, but such accommodations should not occur in the future.


aviatesk:
As Stefan pointed out, we will proceed with Compiler@0.1.1, and this will be replaced by JuliaRegistries/General#132990.

PR #131829, opened 2025-05-27
New version: ROCmDeviceLibs_jll v6.4.1+0

jlbuild:
Autogenerated JLL package registration

* Registering JLL package ROCmDeviceLibs_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/ROCmDeviceLibs_jll.jl
* Version: v6.4.1+0
* Commit: a5c954437462b8e93408e3dd075c525ca2017af0
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/c9ff61c609ad42a0d8e6010a8a9e0a64bd7ea50d
* Created by: @sheevy

PR #132301, opened 2025-06-04
New version: ROCmDeviceLibs_jll v6.4.1+1

jlbuild:

* Registering JLL package ROCmDeviceLibs_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/ROCmDeviceLibs_jll.jl
* Version: v6.4.1+1
* Commit: eabd2bf562eee65beb4e5257c29ab28fe2a6231d
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/b54d81fc341086de93ba9bbb9b6f7a119474ae23
* Created by: @sheevy


amontoison (approved):

sheevy:
@vchuravy This is a fix to https://github.com/JuliaRegistries/General/pull/131829
ROCmDeviceLibs v6.4.1 uses LLVM 19 which is not available in Julia 1.12. Julia master, uses LLVM 20, so it's safe to bump the compat version to 1.13

vchuravy:
This needs to be a new version, since you are changing compat bounds

giordano:
The better approach is to simply fix the compat bound in the registry.  No need for a new release.

PR #135886, opened 2025-08-04
Update Package.toml

cneverett:
Change of package name to fit with the larger Diplodocus framework that is under development and includes this package as a sub-package


GunnarFarneback:
This is the procedure for name changes: https://github.com/JuliaRegistries/General?tab=readme-ov-file#how-do-i-rename-an-existing-registered-package

cneverett:
Thank you, I have now attempted to follow that procedure

PR #136166, opened 2025-08-07
SpikingNeuralNetworks moved to JuliaSNN organization 

aquaresima:
This package has been moved to the [JuliaSNN](https://github.com/JuliaSNN) organization by @AStupidBear.

See the pull request https://github.com/JuliaRegistries/General/pull/135491 where I discussed this with @goerz.

I am updating this manually because of the JuliaRegistrator comment [here](https://github.com/JuliaSNN/SpikingNeuralNetworks.jl/commit/1a91d4618e41fba7c5c25eaecd2088ba449121fc)

If it s not what I should do, please discard it and comment me what is the right thing to do ;)


goerz:
Apart from maybe renaming the PR to something more descriptive, I think this is the correct way to do it. You may also have to post in the `#pkg-registration` channel on Slack

PR #136859, opened 2025-08-17
CI: Add a new CI job that checks that all treehashes exist in the Git repo

DilumAluthge:
## Motivation

The purpose of this PR is to make sure that if the URL of a package is changed (e.g. if an old package is unmaintained, and we change the URL to point to a maintained fork), that all the existing treehashes can still be found in the new URL.

## Demo

Demo of a succeeding treehash job:
- https://github.com/JuliaRegistries/General/pull/136860

Demo of a failing treehash job (in which I edited a treehash so that it would fail):
- https://github.com/JuliaRegistries/General/pull/136861


giordano:
This would run only when the URL is changed? Otherwise I'm concerned this would add lots of latency to all PRs

DilumAluthge:
~~Right now, it will run whenever any of the package's files (`{Package,Compat,Deps,WeakCompat,WeakDeps,Versions}.toml`) are changed.~~

If desired, I can instead make it so that it only runs if `Package.toml` is changed.

DilumAluthge:
> If desired, I can instead make it so that it only runs if `Package.toml` is changed.

Implemented in https://github.com/JuliaRegistries/General/pull/136859/commits/37f7b9d667cf2db8ae3a7542284cfda202071627

ericphanson:
Could we use `paths: **/Package.toml` in the GHA trigger to prevent even starting Julia if it's not needed? Then most PRs will go through without any jobs. We could leave the check in the Julia code so that if a PR touches multiple packages and only touches 1 Package.toml then only that package gets processed further.

DilumAluthge:
> Could we use `paths: **/Package.toml` in the GHA trigger to prevent even starting Julia if it's not needed? Then most PRs will go through without any jobs. We could leave the check in the Julia code so that if a PR touches multiple packages and only touches 1 Package.toml then only that package gets processed further.

Implemented in https://github.com/JuliaRegistries/General/pull/136859/commits/6023d094c787afe1ce677525a2836f2db0979d6a

GunnarFarneback:
How much latency are we talking about? I think running this on every PR would be healthy. I'm sure there are a number of packages which would fail this check today.

ericphanson:
if a package fails because they are missing some very old version, what should we do?

eg https://github.com/JuliaRegistries/General/pull/133351#issuecomment-2993487468

GunnarFarneback:
Ideally add the tree to the repository (in some disconnected but stable manner) if it can be recovered from the package servers, otherwise deregister the version? (Nobody has any way to instantiate it in that case anyway.)

Possible alternatives are to require a manual merge, do some kind of whitelisting, or ignoring sufficiently old versions.

However, the main value would not be to discover historical sins but to detect new anomalies, which might still be possible to recover properly.

ericphanson:
The storage servers could still have it... so in that case we shouldn't remove the old version, but maybe we can add the tree to the repo via the storage server. Hm, maybe we could have a script to pull old versions and push git tags with them to the repo.

DilumAluthge:
How about we proceed as follows:

1. Start out only running on PRs that touch `Package.toml`.
2. Finish reviewing and merging this PR.
3. See how it goes.
4. Consider subsequently broadening to run on any PR that touches `Versions.toml` (or `Package.toml`).

Initially, I think this CI job should not be a "required status check". Once we have it working well on `master`, we can subsequently make it a "required status check" in the future.

@ericphanson @giordano @GunnarFarneback Can you do another round of review?

.ci/Treecheck/Treecheck.jl
```diff
@@ -0,0 +1,187 @@
...
+                    # For each tree, make sure the tree exists
+                    @testset for (k, v) in pairs(versions_dict)
+                        treehash = v["git-tree-sha1"]
+                        tree_libgit2 = LibGit2.GitTree(gitrepo_libgit2, LibGit2.GitHash(treehash))
+                        @test tree_libgit2 isa LibGit2.GitTree
+                    end
+
+                    treehash_to_commithash = Dict{String, String}()
+
+                    # For each tree, make sure that at least one commit points to that tree
```
GunnarFarneback, line 153: There are a couple problems here.
1. Only commits that can be traced from HEAD are detected. That might be too restrictive.
2. There is no support for subdir packages.
3. For performance it would be better to do the `git log` once and cache the result.

I'm also unsure why we are looking for commit hashes in the first place.

.ci/Treecheck/Treecheck.jl
```diff
@@ -0,0 +1,187 @@
...
+                        commithash = _treehash_to_commithash(treehash)
+                        @test !isempty(strip(commithash))
+                        r = r"^[A-Za-z0-9]*?$"
+                        @test occursin(r, commithash)
+
+                        treehash_to_commithash[treehash] = commithash
+                    end
+
+                    # For each commit:
+                    # 1. Make sure that `git checkout COMMIT` does not error
```
GunnarFarneback, line 165: In which scenario would it error? Why does it matter?

.ci/Treecheck/Treecheck.jl
```diff
@@ -0,0 +1,187 @@
...
+                    # 2. Make sure that the commit points to the expected treehash
```
GunnarFarneback, line 166: Why are we checking this? Don't we trust `git log` to provide coherent information? This also lacks subdir support.

.ci/Treecheck/Treecheck.jl
```diff
@@ -0,0 +1,187 @@
...
+    versions_toml = joinpath(package_abspath, "Versions.toml")
+    package_dict = TOML.parsefile(package_toml)
+    versions_dict = TOML.parsefile(versions_toml)
+    package_git_repo_url = package_dict["repo"]
+    @testset "Treecheck for package: $(package_name)" begin
+        @test !isempty(versions_dict)
+        mktempdir() do tmpdir
+            cd(tmpdir) do
+                run(`git clone "$(package_git_repo_url)" MYCLONEDIR`)
+                cd("MYCLONEDIR") do
```
GunnarFarneback, line 141: It seems easier to just clone into `tmpdir`.

GunnarFarneback (commented):
I'm not sure why so much more work than checking the existence of the tree hash is done.

For what it's worth, to guard against a tree hash being at risk of removal, I would run an explicit `git gc` first.

To check that the tree can really be retrieved I would run `git archive`, although this is on the border of mistrusting git.

.ci/Treecheck/Treecheck.jl
DilumAluthge: Ahhhh, you are right, the commit stuff doesn't work for subdir packages.

In theory we can, for each commit, navigate to the subdir and then check that tree, but the problem is that the registry only stores one subdir, but the true subdir might be different for each commit.

I'm going to drop all of the commit stuff.

DilumAluthge: I've removed the commit stuff in https://github.com/JuliaRegistries/General/pull/136859/commits/3209879f8014c2d606ec3dfedb57b4df67b77673

.ci/Treecheck/Treecheck.jl
DilumAluthge: Moot after https://github.com/JuliaRegistries/General/commit/3209879f8014c2d606ec3dfedb57b4df67b77673, I think

.ci/Treecheck/Treecheck.jl
DilumAluthge: Implemented in https://github.com/JuliaRegistries/General/pull/136859/commits/9622aa35c68fcf8157bfabec50caafa107e8e007

DilumAluthge:
> For what it's worth, to guard against a tree hash being at risk of removal, I would run an explicit `git gc` first.

Implemented in https://github.com/JuliaRegistries/General/pull/136859/commits/e550d8afe4f8314bf3d415e281dbae64f3aa5e00

DilumAluthge:
I do still wish that there were some way to check "the tree is reachable somehow from a commit (ideally a commit on `master`/`main`/whatever the repo's default branch is), and thus the tree won't be GC'd away in the future. But with subdir packages (and with the fact that different versions can have different values for `subdir`, I don't know how to do that robustly.

Does Git have some other notion of a tree being "reachable"?

DilumAluthge:
> I do still wish that there were some way to check "the tree is reachable somehow from a commit (ideally a commit on `master`/`main`/whatever the repo's default branch is), and thus the tree won't be GC'd away in the future. But with subdir packages (and with the fact that different versions can have different values for `subdir`, I don't know how to do that robustly.
> 
> Does Git have some other notion of a tree being "reachable"?

Anyway, I think we can defer the above discussion for a follow-up PR. No sense holding up this PR for that.

DilumAluthge:
While we're waiting for Gunnar's next round of review: @giordano @ericphanson Can you also give this another round of review?

.ci/Treecheck/Project.toml
giordano, line 1: It'd be good to have compat bounds, to avoid having this test break randomly when a new breaking version of some packages (at least JSON3 isn't an stdlib) comes out

.github/workflows/treecheck.yml
```diff
@@ -0,0 +1,90 @@
...
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
+jobs:
+  generate-matrix-treecheck:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
giordano, line 27:
```suggestion
      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```

.github/workflows/treecheck.yml
```diff
@@ -0,0 +1,90 @@
...
+    # Don't error on empty matrix:
+    if: ${{ needs.generate-matrix-treecheck.outputs.matrix != '[]' && needs.generate-matrix-treecheck.outputs.matrix != '' }}
+    needs:
+      - generate-matrix-treecheck
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix: ${{fromJson(needs.generate-matrix-treecheck.outputs.matrix)}}
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
giordano, line 67:
```suggestion
      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```

.ci/Treecheck/Project.toml
DilumAluthge: Done in a23c7846575faa7711abaf35ce83525c358d1d5d

.ci/Treecheck/Treecheck.jl
```diff
@@ -0,0 +1,184 @@
...
+# I don't want this script to randomly break.
+#
+# Note: The function `Tar.tree_hash()` is part of Tar.jl's public API.
+# See: https://github.com/JuliaIO/Tar.jl/blob/9dd8ed1b5f8503804de49da9272150dcc18ca7c7/README.md?plain=1#L17-L32
+function verify_archive_tree_hash(tar_gz::AbstractString, expected_hash::Base.SHA1)
+    # This can fail because unlike sha256 verification of the downloaded
+    # tarball, tree hash verification requires that the file can i) be
+    # decompressed and ii) is a proper archive.
+    calc_hash = try
+        Base.SHA1(open(Tar.tree_hash, `$(exe7z()) x $tar_gz -so`))
```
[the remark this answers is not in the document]
DilumAluthge: > Enclosing the computed and expected hashes in the `Base.SHA1` type is questionable in that it's not a public type. For practical purposes I don't think it adds any more value over direct string comparison than case normalization.

Done in https://github.com/JuliaRegistries/General/pull/136859/commits/3a1565ddf1ff9c607500df381951d73c4695ef7b

GunnarFarneback (approved):

.ci/Treecheck/Project.toml
```diff
@@ -0,0 +1,17 @@
+[deps]
+CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
+JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
+LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
+TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
+Tar = "a4e569a6-e804-4fa4-b0f3-eef7a1d5b13e"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+p7zip_jll = "3f19e933-33d8-53b3-aaab-bd5110c3b7a0"
```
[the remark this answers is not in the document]
DilumAluthge: Done in 30289ed5897305d1589e8b2b35905afd11052dc4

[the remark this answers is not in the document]
GunnarFarneback: Better remove all references to it as well.

DilumAluthge:
@GunnarFarneback @ericphanson @giordano Any additional comments here?

GunnarFarneback:
https://github.com/JuliaRegistries/General/pull/136859#discussion_r2318145280

.ci/Treecheck/Project.toml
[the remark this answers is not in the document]
DilumAluthge: Done in https://github.com/JuliaRegistries/General/commit/73991d47d337161a3c8aaf3325370dad49581fa9

DilumAluthge:
> [#136859 (comment)](https://github.com/JuliaRegistries/General/pull/136859#discussion_r2318145280)

Done in https://github.com/JuliaRegistries/General/commit/73991d47d337161a3c8aaf3325370dad49581fa9

DilumAluthge:
Alright, let's give this a try. If there are further comments, I can address them in a follow-up PR.

PR #137826, opened 2025-09-02
Fix OrdinaryDiffEqCore yank

ChrisRackauckas:
@DilumAluthge the CI didn't seem to catch this?


DilumAluthge:
In the future, we'll catch this with https://github.com/JuliaRegistries/RegistryCI.jl/pull/605

PR #137833, opened 2025-09-02
CI: Add a new CI job that runs `Pkg.add("Example")` on every PR (even PRs that AutoMerge doesn't run on)

DilumAluthge:
I think that this would have caught https://github.com/JuliaRegistries/RegistryCI.jl/issues/604.

Now, the best fix for https://github.com/JuliaRegistries/RegistryCI.jl/issues/604 is to specifically test for it, which is what https://github.com/JuliaRegistries/RegistryCI.jl/pull/605 does.

However, I think this PR is still a good thing to have. The "registry consistency tests" (`RegistryCI.test(reg_path)`) are never going to be guaranteed to cover 100% of potential bugs. So I think that an additional sanity check (this PR) will be useful, even after we merge https://github.com/JuliaRegistries/RegistryCI.jl/pull/605.

### DIFF
--- a/.github/workflows/example_pkg_add.yml
+++ b/.github/workflows/example_pkg_add.yml
@@ -1,0 +1,63 @@
+name: Example `Pkg.add()`
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Hmmm. This doesn't seem sufficient. E.g. this CI job succeeds on #137834.

PR #137834, opened 2025-09-02
Test if #137833 catches bad registry PRs

DilumAluthge:
> Revert "Fix OrdinaryDiffEqCore yank (#137826)"
> This reverts commit 20b778710e3c9a8f4c2045bb25ab92175f781316.

Targets `dpa/pkg-add-example` (#137833).


DilumAluthge:
After some local testing:
1. `Pkg.add("Example")` succeeds
2. `Pkg.add("OrdinaryDiffEqCore")` errors

So it's not sufficient to just always check `Pkg.add("Example")`.

PR #137946, opened 2025-09-04
Update URLs of TensorKit and TensorOperations to reflect transfer

kshyatt:
These have now moved to the @QuantumKitHub org


kshyatt:
cc @jutho @lkdvos

PR #138160, opened 2025-09-08
Make old UnitfulLatexify incompatible with new Unitful

Ickaser:
See discussion at https://github.com/JuliaPhysics/Unitful.jl/pull/808#issuecomment-3263826864. I have been replacing the glue package UnitfulLatexify with a Latexify extension to Unitful; if the glue package also gets loaded, it breaks the extension's functionality, so we need to make sure users don't load both.

In a new release of the glue package, we intend to make loading the glue package simply load the extension with no other code (https://github.com/gustaphe/UnitfulLatexify.jl/pull/22). That should probably get merged and released before this PR, I imagine, to avoid merge conflicts here, but I'm new to the whole registry system and will happily take advice.


U/UnitfulLatexify/Compat.toml
```diff
@@ -15,5 +17,8 @@ Latexify = "0.15.1-0.15"
 ["1.6.3-1.6"]
 Latexify = "0.15.1-0.16"
 
-["1.7-1"]
+["1.7.0-1.7"]
```
Ickaser, line 20:
```suggestion
["1.7-1"]
```

U/UnitfulLatexify/Compat.toml
```diff
@@ -15,5 +17,8 @@ Latexify = "0.15.1-0.15"
...
 Latexify = "0.16.8-0.16"
+
+["1.8-1"]
+Unitful = "1.25-1"
```
Ickaser, line 24:
```suggestion
```

U/UnitfulLatexify/Compat.toml
```diff
@@ -1,11 +1,13 @@
...
 LaTeXStrings = "1.2.0-1"
-Unitful = "1.5.0-1"
 julia = "1"
 
 ["1-1.3"]
 Latexify = "0.14.7-0.14"
 
+["1-1.7"]
+Unitful = "1.5.0-1.24"
+
```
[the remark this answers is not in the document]
Ickaser: Yeah, you're right

PR #139286, opened 2025-09-28
Update URL for rename from IndexedStructVectors.jl to IdTables.jl

ameligrana:
This is in tandem with #139285


goerz:
Sorry, but registered packages cannot be renamed. See the ["How do I rename an existing registered package?" FAQ entry](https://github.com/JuliaRegistries/General?tab=readme-ov-file#how-do-i-rename-an-existing-registered-package)

goerz:
GitHub automatically forwards https://github.com/Tortar/IndexedStructVectors.jl.git, so this doesn't do much.

PR #139294, opened 2025-09-28
Revert "New package: POMLUtils v0.1.0"

DilumAluthge:
Reverts JuliaRegistries/General#138169

---

The package repo was deleted, and it was deleted so soon that the Pkg/storage servers never obtained a copy of the package.

Cross-ref: https://github.com/JuliaRegistries/General/pull/138218#issuecomment-3343030372

---

Note for posterity: We almost never revert or delete packages or package versions. This is a rare special case.


giordano (approved):
This seems to be an exceptional case of a package that was never installable because it was deleted before registration.

PR #140165, opened 2025-10-10
Move ParallelTestRunner to JuliaTesting

vchuravy:


DilumAluthge (approved):
1. I confirmed that the old URL redirects to the new URL.
2. The author of this PR is also the "owner" in the old URL.
3. The PR preserves the trailing `.git`.

PR #140249, opened 2025-10-12
Add Julia 1.12 to registry-consistency

ViralBShah:
Also remove Julia 1.8 and 1.9

cc @DilumAluthge

### DIFF
--- a/.github/workflows/registry-consistency-ci-cron.yml
+++ b/.github/workflows/registry-consistency-ci-cron.yml
@@ -28,7 +28,7 @@ jobs:
           - '1.9'
           - '1.10'
           - '1.11'
-          # - '1.12' # Uncomment this once Julia 1.12 is released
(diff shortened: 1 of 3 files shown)


DilumAluthge (commented):
Let's keep 1.8 and 1.9? We usually keep older Julia versions in the registry consistency tests, due to the significant pain that it would cause users if a buggy PR is merged.

I agree with adding 1.12.

ViralBShah:
I was trying to remove some older ones when adding new. With 1.3 and 1.6 around, isn't that sufficient?

Anyways, added 1.8 and 1.9 back. Don't have a strong view on removing them.

ViralBShah:
Ok, so the 1.12 stuff is having issues.

ViralBShah:
Fixed in #140258

PR #140259, opened 2025-10-12
CI: Start generating an AutoMerge CI manifest for Julia 1.12 (keep the 1.11 manifest also)

DilumAluthge:
To make it smoother to upgrade AutoMerge, when we do so in the future.


DilumAluthge:
This might get confusing. Perhaps we should only have two manifests for the brief period of time when we migrate.

PR #140415, opened 2025-10-14
[HMMER_jll] Update GSL_jll version in Compat.toml

### LINKED ISSUE #5283: [HMMER_jll] GSL dependency error
Trying to run one of the binaries from hmmer (https://github.com/JuliaBinaryWrappers/HMMER_jll.jl), 

```
julia> import HMMER_jll
julia> hmmerpath(name::AbstractString) = joinpath(HMMER_jll.artifact_dir, "bin", name)
julia> hmmercmd(name::String, args...) = `$(hmmerpath(name)) $args`
julia> hmmerexe(name::String, args...) = run(hmmercmd(name, args...))
julia> hmmerexe("hmmemit", "-a data/WW.hmm")
~/.julia/artifacts/6032a5a927fb181407665ca48030f89584d10dfe/bin/hmmemit: error while loading shared libraries: libgsl.so.25: cannot open shared object file: No such file or directory
```

You can check that the HMMER_jll correctly lists GSL as a dependency in the Project.toml, but apparently it's not really calling into it?

P.S.: I'm not sure if this is the correct place to post an issue about a wrapped binary. If not please point me to the correct place.

@marcom maybe you have some ideas?

ViralBShah:
Updated GSL_jll version for compatibility with Julia.

Fix https://github.com/JuliaPackaging/Yggdrasil/issues/5283


ViralBShah:
@giordano Is this right, based on your comment in https://github.com/JuliaPackaging/Yggdrasil/issues/5283#issuecomment-2887781228 ?

PR #140718, opened 2025-10-18
Add LLM guidelines

goerz:
As discussed in https://discourse.julialang.org/t/should-general-have-a-guideline-or-rule-preventing-registration-of-vibe-coded-packages/133205/, which seems to me to have come down on the consensus that LLM-*assisted* coding is fine for registered packages, but blind "vibe-coding" is not.


README.md
```diff
@@ -119,8 +119,35 @@ collections of "utility functions" nor for packages that are only useful for a c
...
+understanding of the generated code.
+
+[It is not okay to submit vibe-coded packages](https://discourse.julialang.org/t/should-general-have-a-guideline-or-rule-preventing-registration-of-vibe-coded-packages/133205),
+defined as "building software with an LLM without reviewing the code it
+writes". Packages that exhibit obvious signs of ["AI slop"](https://en.wikipedia.org/wiki/AI_slop)
+may be blocked from registration. To ensure that a package containing
+LLM-generated components is suitable for this registry, follow best
+practices as much as possible:
+
+* Review and understand all generated code, by hand.
```
ericphanson: let's add Keno's disclosure suggistion (https://discourse.julialang.org/t/should-general-have-a-guideline-or-rule-preventing-registration-of-vibe-coded-packages/133205/16?u=ericphanson). Wording from [base](https://github.com/JuliaLang/julia/blob/master/CONTRIBUTING.md#contributor-checklist)
```suggestion
* if your package contains substantial contributions from a generative AI tool, please disclose so with details in the README
* Review and understand all generated code, by hand.
```

README.md
```diff
@@ -119,8 +119,39 @@ collections of "utility functions" nor for packages that are only useful for a c
...
+* Write READMEs and other parts of the documentation by hand; review and trim
+down any documentation generated by an LLM.
+* Ensure that the [code is tested](https://pkgdocs.julialang.org/v1/creating-packages/#Adding-tests-to-the-package)
+and that these tests run in continuous integration (CI), such as GitHub Actions.
+Collect and track coverage information.
+* Build and deploy documentation in CI, using a system like
+[Documenter.jl](https://documenter.juliadocs.org/stable/) where appropriate.
+
+When communicating about a package registration, never just blindly paste in an
+LLM response to a question raised in the registration PR on GitHub, Slack, etc.
```
ericphanson: Here's another take, maybe a bit too long:
```suggestion
When communicating about a package registration, make sure to communicate _your thoughts_
and ideas, don't rely on an LLM to think for you. We want to hear from you! Occasionally, folks will copy feedback
on a registration directly into an LLM that doesn't have the context or understanding to reply accurately; please don't do this.
```

goerz: Yeah, that looks very good

DilumAluthge (dismissed):
Seems like a good start. We can always make further improvements and refinements in follow-up PRs.

DilumAluthge (dismissed):

README.md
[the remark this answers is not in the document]
goerz: Replaced with Eric's version (lightly edited) in latest push

ericphanson:
let's give this a day or two in case of more review feedback, and plan to merge early next week if there's nothing blocking

andreeco:
I have some concerns about this part:

> Write READMEs and other parts of the documentation **by hand**; review and trim down any documentation generated by an LLM.

While I think that one should check the documentation and refine it manually, it’s overly restricting. Why should one be forced to retype all parts (even the signatures of functions in docstrings) and not just refine and optionally trim it?

Also, be mindful of **information overload** in the README and throughout the package registration process. New contributors may need to learn many new things — Git, GitHub apps (JuliaTeam Registrator), GitHub Actions, GitHub Pages, and more. This can be quite overwhelming.

In line with the KISS principle (Keep It Simple and Straightforward) one could simply write:

> “Review, understand, and refine all LLM-generated code and documentation line by line.”

Also one could provide a separate guide on how to use LLMs in a wise way for code generation in Julia.

If someone has used an LLM in an unprofessional way, one can link the guide to the person and kindly ask them to rework the package before the registration continues.

goerz:
> While I think that one should check the documentation and refine it manually, it’s overly restricting.

I disagree. A README (for most packages that also have a full Documenter-based documentation) should be a very concise, with a paragraph or two to give context for the package, brief installation instructions, maybe a minimal usage example, links to the full documentation and contribution guidelines. These should never be generated by an LLM. A package author should be able to express their own thoughts in a paragraph to describe the purpose of their package.

If they need some help from an LLM, e.g., to overcome a language barrier (translation), fine. If they get a good result out of the LLM, nobody would care or notice. We're not really banning the use of LLM, we're banning overly verbose slop that shows no understanding of what the motivation was for the creation of a package, how it fits into the ecosystem, and all that.

> Why should one be forced to retype all parts (even the signatures of functions in docstrings) and not just refine and optionally trim it?

The signatures of functions should not be written by hand at all. They should be automatically included in Documenter via `@autodocs` etc. Other non-reference parts of a good documentation (I'm thinking here mainly in terms of the [Diataxis philosoph](https://diataxis.fr) with tutorials, explanations, how-to guides) should also be written by hand. These require coherent thought, and any LLM we currently have will be _terrible_ at these. Maybe LLMs will get significantly better and the point will become moot. If people have LLMs assist them in improving their writing (or, again, translation), that's fine too. If nobody notices any LLM slop, it's not a problem. I still would absolutely want to nudge people towards writing documentation by hand. They'll ignore that advice anyway ;-)

In any case, these are guidelines, not hard rules. It's not like we're going to say "Oh, you used Claude to structure your documentation, you are BANNED":

> If someone has used an LLM in an unprofessional way, one can link the guide to the person and kindly ask them to rework the package before the registration continues.

That's exactly what the intent behind these guidelines is.

> Also, be mindful of information overload in the README and throughout the package registration process.

Nobody reads the General registry FAQ unprompted (I think). It's actually a little hard to find, as it is. As you say, this is just something we can link people to when they submit packages that show clear signs of LLM slop.

andreeco:
With function signature, I meant especially the signatures in [docstrings](https://github.com/JuliaLang/julia/blob/74c32ec0b522d528d279efaed502a24e917aeeff/doc/src/manual/documentation.md). It did not come to my mind that people try to manually write documentation without using `@autodocs` or friends.

There is no questions that both code and docs need to be reviewed and refined. For example, in both Julia and Python one should use a one-line, imperative sentence in the beginning:

 - [Julia recommandations](https://github.com/JuliaLang/julia/blob/74c32ec0b522d528d279efaed502a24e917aeeff/doc/src/manual/documentation.md)
- [Python recommandations](https://peps.python.org/pep-0257/)

Yet LLMs often ignore these simple (soft) rules and go for the third person and don't write the one-liner as they should. But I do think that forcing manual typing is restrictive, while it may be a good way to achieve good quality.

One more thing I want to mention is that I looked into the #new-packages-feed, and I appreciate the work you do there. I also see why you are so concerned about this, as I looked into some of the submitted packages. You do a good work there!

goerz:
> With function signature, I meant especially the signatures in [docstrings](https://github.com/JuliaLang/julia/blob/74c32ec0b522d528d279efaed502a24e917aeeff/doc/src/manual/documentation.md).

Yeah, I was mostly referring to all the other parts of the documentation, not docstrings or other "reference" material. Good docstrings have their own challenges, but one could probably use Claude for those, with some care. Most likely, you'd want to point Claude to a docstring-linter (is there a good one for Julia?) that enforces things like "first sentence imperative". I consider docstrings pretty much part of the code, so the "use LLM assistance as needed, and make sure to manually guide and review" applies in that context.

I've added "non-reference" in the [latest commit](https://github.com/JuliaRegistries/General/pull/140718/commits/efc6ac1100b940b5ea375c48143a37a07f1d6ec5), which also gives the opportunity to link to [Diataxis](https://diataxis.fr) (not because that "framework" is entirely without controversy, but because _thinking_ about it will make people write better documentation)

PR #140810, opened 2025-10-20
URL change for the NextLA package

fda-tome:
The package changed repos, but it's still the same. This is needed  for a new version publication.


N/NextLA/Package.toml
```diff
@@ -1,3 +1,3 @@
 name = "NextLA"
 uuid = "d37ed344-79c4-486d-9307-6d11355a15a3"
-repo = "https://github.com/Rabab53/NextLA.jl.git"
+repo = "https://github.com/NextLinearAlgebra/NextLA.jl"
```
giordano:
```suggestion
repo = "https://github.com/NextLinearAlgebra/NextLA.jl.git"
```

PR #140818, opened 2025-10-20
AutoMerge staging: update to new API

ericphanson:
follows https://juliaregistries.github.io/RegistryCI.jl/dev/migration-v1/

closes https://github.com/JuliaRegistries/General/pull/140259

requires https://github.com/JuliaRegistries/RegistryCI.jl/pull/641 to be merged first since I'm using that syntax here


ericphanson:
this is staging workflow only which only has read permissions, so I'm going to merge in order to be able see how it does

PR #140842, opened 2025-10-20
[BinaryBuilderBase] Allow SHA v1

giordano:
At the moment the package is not installable on Julia nightly.


IanButterworth:
FWIW I'm hoping to make this only emit a warning https://github.com/JuliaLang/Pkg.jl/pull/4437

PR #141098, opened 2025-10-23
Add deprecation notice to QuantumCliffordPlots

Krastanov:
Trying out the new capabilities from https://github.com/JuliaLang/Pkg.jl/pull/4433

### DIFF
--- a/Q/QuantumCliffordPlots/Package.toml
+++ b/Q/QuantumCliffordPlots/Package.toml
@@ -1,3 +1,7 @@
 name = "QuantumCliffordPlots"
 uuid = "4eeed724-23be-436c-a322-a8540588cc9f"
 repo = "https://github.com/QuantumSavory/QuantumCliffordPlots.jl.git"
+
+[metadata.deprecated]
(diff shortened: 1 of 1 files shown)


DilumAluthge:
@ericphanson Can we have some checks in RegistryCI to check that the formatting of this kind of metadata is correct?

I guess the first thing to check is that the only top level keys are name, uuid, repo (all of which are required), and then a top-level metadata table, which is optional.

DilumAluthge:
@Krastanov Can I keep this open for now, to use in testing?

DilumAluthge:
@ericphanson For additional checks, should we require that if metadata.deprecated exists, then a reason is provided?

ericphanson:
yeah, might be good. could also check alternative, if provided, is the name of a package in the registry

Krastanov:
> @Krastanov Can I keep this open for now, to use in testing?

Certainly! I just wanted to get the ball rolling. Thanks for taking care of the maintenance of these capabilities!

ericphanson:
Maybe we should merge this and open issues for follow-ups?

DilumAluthge (approved):

DilumAluthge:

PR #141630, opened 2025-10-31
Cap compat for Kmers.jl v0.x at 3.3 to avoid installing old version

BioTurboNick:
Issue is that Kmers 1.x added a too-narrow compat bound on BioSequences, so that when the BioSequences 3.5 was released, Pkg installs BioSequences 3.5 and Kmers 0.1.

### DIFF
--- a/K/Kmers/Compat.toml
+++ b/K/Kmers/Compat.toml
@@ -1,5 +1,5 @@
 [0]
-BioSequences = "3"
+BioSequences = "3-3.3"
 julia = "1.5.0-1"
 
 [1]


DilumAluthge (approved):
Seems reasonable to me. Can you check with other BioJulia folks to see if the narrow compat was intentional?

Note to self: If we later hear otherwise, we can always revert this PR later if need be.

BioTurboNick:
I am - but this PR doesn't actually address the narrow compat per se, it just stops the resolver from producing a pathological result (BS 3.5/Kmers 0.1) instead of a reasonable one (BS 3.4/Kmers 1.2).

PR #141759, opened 2025-11-03
Add deprecation notice to unzip_jll

nhz2:
Added deprecation metadata indicating issues with UTF-8 handling on Windows and suggesting an alternative package.

This PR is based on https://github.com/JuliaRegistries/General/pull/141098

Ref: https://github.com/JuliaPackaging/Yggdrasil/pull/12337 https://github.com/JuliaPackaging/Yggdrasil/issues/7679#issuecomment-3419079074


nhz2:
It turns out that p7zip_jll has a bug on Linux https://github.com/JuliaLang/julia/issues/60071

Maybe we need to wait until the fix in https://github.com/JuliaLang/julia/pull/60025 is backported

nhz2:
This is not stale

DilumAluthge (approved):

DilumAluthge:

PR #141776, opened 2025-11-04
Add CODEOWNERS entry for StaticArrays.jl

DilumAluthge:
Changes here can (and did) break Base Julia CI, which can end up blocking Julia releases. So ping some General registry and/or Pkg folks, to increase visibility of these changes.

I just added myself to start - @KristofferC @fredrikekre let me know if you also want to be added.


DilumAluthge:
As an example, Julia 1.11.x releases are blocked currently (https://github.com/JuliaRegistries/General/pull/140218#issuecomment-3483343775), so the goal is to increase visibility when these kinds of PRs are opened (and to ensure that fixes are backported to appropriate release branches).

For this specific example, I'm going to work on backporting the necessary fix to Julia 1.11.

PR #142106, opened 2025-11-07
Update repository URL for WinReg package

ViralBShah:


DilumAluthge (approved):
1. I verified that the old URL redirects to the new URL.
2. The PR preserves the trailing `.git` at the end of the URL.

PR #142583, opened 2025-11-16
Update AbaqusReader repository URL

ahojukka5:
This PR updates the repository URL for the AbaqusReader package.

## Summary
The AbaqusReader.jl package repository has been transferred from the JuliaFEM organization to ahojukka5's personal account.

## Changes
- **Old URL**: https://github.com/JuliaFEM/AbaqusReader.jl.git
- **New URL**: https://github.com/ahojukka5/AbaqusReader.jl.git

## Background
- The package is maintained by the same author (Jukka Aho, @ahojukka5)
- The repository was transferred to the author's personal account for better maintenance control
- This update is required for JuliaRegistrator to accept new version registrations (v0.2.7)
- The JuliaFEM organization repository still exists but is no longer the primary development location

## Verification
- Package UUID remains unchanged: bc6b9049-e460-56d6-94b4-a597b2c0390d
- All existing versions are accessible from both URLs
- New releases will be published from the ahojukka5 repository

cc: @JuliaRegistrator


ahojukka5:
Nothing seems to happen. Do I need to do something?

ahojukka5:
This should be manually merged as I'm changing the github address of package and automation cannot handle this change.

DilumAluthge (approved):
1. I have verified that the old URL automatically redirects to the new URL in the browser.

DilumAluthge:
I will merge this PR per the request of the package author.

However, I will also make the following comment:

As a general rule, we strongly recommend that repos live in GitHub organizations whenever possible. When a repos lives in a GitHub org, if the original package author stops maintaining the package and becomes unreachable[^1], other people can take over the package and continue maintenance - this is much harder to accomplish if the repos is owned by an individual user.

So, with all that being said, @ahojukka5 I would recommend that you re-consider moving the package back into a GitHub org in the future.

[^1]: This is not a hypothetical situation - it has happened to us multiple times.

PR #143732, opened 2025-12-04
Yank NLPModels.jl 0.22

amontoison:
We have decided to yank release 0.22 of `NLPModels.jl` because the breaking change only affected the removal of a parameter (`x`) from three functions in the public API. This breaking release disrupts compatibility with the existing ecosystem — over 50 registered packages depend on this package.

A non-breaking alternative is possible and straightforward (e.g., deprecating the parameter instead of removing it), which allows us to:

* maintain ecosystem stability;
* avoid unnecessary disruption for users;
* give downstream maintainers time to adapt without urgency.

In hindsight, the release was pushed a bit hastily. Yank 0.22 allows us to preserve compatibility and ensures that breaking releases are made only for substantial, justified changes.


amontoison:
We still want to yank the release 0.22.0 if it is possible.

PR #143667, opened 2025-12-03
New version: CURL_jll v8.17.0+0

jlbuild:
Autogenerated JLL package registration

* Registering JLL package CURL_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/CURL_jll.jl
* Version: v8.17.0+0
* Commit: 248ab19c78b4469a41316d1d7ed17d2c70ecb72e
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/cdb16ff7672ce22cb25c9eb10ebd267a6e5368a1
* Created by: @eschnett

PR #143738, opened 2025-12-04
New version: autoconf_jll v2.72.0+1

jlbuild:

* Registering JLL package autoconf_jll.jl
* Repository: https://github.com/JuliaBinaryWrappers/autoconf_jll.jl
* Version: v2.72.0+1
* Commit: 51a7f8a77f3a9e30dc649265653e01a3bf7cb29a
* Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/1153fde8a78259d32620d1182154cd1a7abdcdb9
* Created by: @fingolfin


jll/A/autoconf_jll/Compat.toml
```diff
@@ -1,10 +1,11 @@
...
-["2-2.71"]
+["2 - 2.71"]
 JLLWrappers = "1.2.0-1"
 
-["2.72-2"]
+["2.72 - 2"]
 Artifacts = ["0.0.0", "1"]
 JLLWrappers = "1.7.0-1"
 Libdl = ["0.0.0", "1"]
+julia = "1.6.0 - 1"
```
fingolfin:
```suggestion
```
Argh where did that come from? 

@giordano can we just drop it? Probably not and the wrapper needs it ... argh 

But I still don't understand: looking at https://github.com/JuliaBinaryWrappers/autoconf_jll.jl/commit/51a7f8a77f3a9e30dc649265653e01a3bf7cb29a it seems this compat requirement was there before already, just not reflected in the registry?!

giordano: I suspect this is a bug in RegistryTools.

fingolfin: Hrm, this looks similar: https://github.com/JuliaRegistries/General/pull/143667/files?diff=unified&w=1

And then @eschnett reported https://github.com/JuliaRegistries/RegistryTools.jl/issues/108 for https://github.com/JuliaRegistries/General/pull/143468

fingolfin: OK based on what I gather from the issue and proposed fix, the patch I propose above seems to actually be a correct workaround for the RegistryTools bug, as the range `julia = "1.6.0-1"` is already present in line 2, and RegistryTool simply mistakenly considers this to be different from `julia = "1.6.0 - 1" .

jll/A/autoconf_jll/Compat.toml
```diff
@@ -1,10 +1,10 @@
 [2]
 julia = "1.6.0-1"
 
-["2-2.71"]
+["2 - 2.71"]
 JLLWrappers = "1.2.0-1"
 
-["2.72-2"]
+["2.72 - 2"]
```
giordano:
```suggestion
["2-2.71"]
JLLWrappers = "1.2.0-1"

["2.72-2"]
```

PR #143821, opened 2025-12-06
[ONNX_jll] Updated dependencies to reflect JuliaBinaryWrappers

stemann:
Follow-up from re-build of the single ONNX_jll version that exists: #143819

Removed protoc_jll dependency from Deps.toml - protoc_jll is no longer a dependency of ONNX_jll: https://github.com/JuliaBinaryWrappers/ONNX_jll.jl

ONNX_jll has zero dependents (so far).


lgoettgens:
This seems like a bad idea to me. Maybe @giordano could elaborate a bit?

stemann:
The registration in #143819 properly added `CompilerSupportLibraries_jll` as a dependency, but the old dependency on `protoc_jll` was not "cleaned up".

lgoettgens:
The problem is that ONNX_jll v1.10.2+0 uses the same declaration of dependencies as the entry you are editing here. This would mean that every manifest containing that version would no longer be able to get instantiated, thus violating the reproducibility of environments.
How it should have happened: Your change in Yggdrasil (that resulted in https://github.com/JuliaRegistries/General/pull/143819) should have changed the version number. That way, the dependencies would have been tracked properly in the registry.
How this particular case should be handled needs to be decided by a full registry maintainer (that also knows about BinaryBuilder stuff). I am deliberately not pinging anyone here, please use the slack channel instead.

imciner2:
The `+1` in the version number is meaningless when it comes to dependency resolution. All versions sharing the same major.minor.patch must have the same dependencies to prevent environments from getting broken. If the dependencies need to change (especially removing one), then it needs a new major.minor.patch version, not a new build number.

stemann:
I was just hoping for an exception in this case - where there are zero public dependents. Sure there could theoretically be a private one out there, but I’d say the chances of that are pretty slim - for a C++ library in a Julia content. The primary potential downstream consumers are Torch and ONNXRuntime jll’s - both of which I also maintain.

stemann:
@imciner2 @giordano Please give this a second thought:

ONNX_jll v1.10.2 was re-built, re-defining ONNX_jll to provide dynamic shared libraries - moving the static library in ONNX_jll v1.10.2+0 to ONNX_static_jll. This broke all non-Manifest applications (notably all of them as there was only ever one version of ONNX_jll).

Does it make sense (in this case) to safe-guard the (likely non-existent) Manifest-defined environments out there?

stemann:
... and it might be the issue I cannot get Torch to build, due to presence of a wrong version of protobuf present while building, e.g.,
```
┌ Warning: Symlink protoc from artifact 3256ac98b9df538966a7cd30f40e42482ff46ada already exists in artifact 45b6079c3d821ea09fcb5123ea0a87902f802735
```
https://buildkite.com/julialang/yggdrasil/builds/25881/steps/canvas?sid=019b1799-4b7f-4507-a8fd-44e81a782df0

PR #144324, opened 2025-12-12
OpenMPI_jll: Restrict julia compat

eschnett:
`OpenMPI_jll 5.0.9` was only built with `libgfortran5` and is not compatible any more with earlier Julia versions.


giordano:
We can't change compat bounds

PR #145354, opened 2026-01-01
[JET] Add upper bounds to julia compat entries

ChrisRackauckas-Claude:
## Summary

JET.jl has strict Julia version requirements that were not properly reflected in the registry compat entries:

- **JET 0.9.x** (v0.8.28-0.9.18) requires Julia 1.10 only (not 1.11+)
- **JET 0.9.19+** requires Julia 1.11 only (not 1.12+)
- **JET 0.10.x and 0.11.x** require Julia 1.12 only

The previous entries used unbounded ranges like `julia = "1.10.0-1"` which allowed incompatible Julia versions. This caused issues when the resolver would select JET versions that don't work with the user's Julia version.

For example, on Julia 1.11, the resolver could previously select JET v0.9.0-0.9.18 (which only works on Julia 1.10), causing cryptic failures.

## Changes

| JET Version Range | Before | After |
|---|---|---|
| 0.8.28 - 0.9.18 | `julia = "1.10.0-1"` | `julia = "1.10"` |
| 0.9.19 | `julia = "1.11.0-1"` | `julia = "1.11"` |
| 0.9.20 - 0.9.x | `julia = "1.11.0 - 1"` | `julia = "1.11"` |
| 0.10 - 0.10.6 | `julia = "1.12.0-1"` | `julia = "1.12"` |
| 0.10.7+ | `julia = "1.12.0 - 1"` | `julia = "1.12"` |

## Verification

The upstream JET.jl Project.toml files confirm these constraints:
- [JET v0.9.14 Project.toml](https://github.com/aviatesk/JET.jl/blob/v0.9.14/Project.toml): `julia = "1.10"`
- [JET v0.10.14 Project.toml](https://github.com/aviatesk/JET.jl/blob/v0.10.14/Project.toml): `julia = "1.12"`
- [JET v0.11.3 Project.toml](https://github.com/aviatesk/JET.jl/blob/v0.11.3/Project.toml): `julia = "1.12"`

## Reference

See discussion at: https://github.com/SciML/RootedTrees.jl/pull/209#issuecomment-3703421412

### DIFF
--- a/J/JET/Compat.toml
+++ b/J/JET/Compat.toml
@@ -1,5 +1,5 @@
 ["0.0"]
-julia = "1-1.10"
+julia = "1-1.6"
 
 ["0.1"]
 LoweredCodeUtils = "1.2.4-1"
@@ -96,7 +96,7 @@ JuliaInterpreter = "0.8.16-0.8"
 LoweredCodeUtils = "2.1"
 
 ["0.5 - 0.6.1"]
-julia = "1.7.0-1.10"
+julia = "1.7.0-1.8"
 
 ["0.5.4 - 0.8"]
 Revise = "3.3.0-3"
@@ -111,10 +111,10 @@ JuliaInterpreter = "0.9"
 julia = "1.8.1-1.8.3"
 
 ["0.6.2 - 0.6.15"]
-julia = "1.8.0-1.10"
+julia = "1.8"
 
 ["0.6.21 - 0.7.14"]
-julia = "1.8.1-1.10"
+julia = "1.8"
 
 ["0.7.11 - 0.7.13"]
 Preferences = "1.3.0-1"
@@ -126,7 +126,7 @@ PrecompileTools = "1"
 Preferences = "1.4.0-1"
 
 ["0.7.15 - 0.8.21"]
-julia = "1.9.0-1.10"
+julia = "1.9"
 
 ["0.7.2 - 0.7.11"]
 SnoopPrecompile = "1"
--- a/J/JET/Compat.toml
+++ b/J/JET/Compat.toml
@@ -32,7 +32,7 @@ MacroTools = "0.5.6-0.5"
 Pkg = "1.10.0-1"
 Preferences = "1.4.0-1"
 Test = "1.10.0-1"
-julia = "1.12.0-1"
+julia = "1.12"
 
 ["0.10.11 - 0"]
 JuliaSyntax = "1 - 2"
@@ -65,7 +65,7 @@ Pkg = "1.10.0 - 1"
 PrecompileTools = "1.3.2 - 1"
 Preferences = "1.4.0 - 1"
 Test = "1.10.0 - 1"
-julia = "1.12.0 - 1"
+julia = "1.12"
 
 ["0.10.7 - 0.10.10"]
 JuliaSyntax = "1"
@@ -148,7 +148,7 @@ Test = "1.10.0-1"
 LoweredCodeUtils = "2.2-2.3"
 
 ["0.8.28 - 0.9.18"]
-julia = "1.10.0-1"
+julia = "1.10"
 
 ["0.9 - 0.9.6"]
 LoweredCodeUtils = "2.4.5-2"
@@ -163,7 +163,7 @@ LoweredCodeUtils = "3.0.2-3"
 JuliaSyntax = "0.4.10-0.4"
 
 ["0.9.19"]
-julia = "1.11.0-1"
+julia = "1.11"
 
 ["0.9.20 - 0.9"]
 CodeTracking = "1.3.1 - 1"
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Added a second commit to also cap the older JET versions for better resolver behavior:

| JET Version Range | Before | After |
|---|---|---|
| 0.0.x | `julia = "1-1.10"` | `julia = "1-1.6"` |
| 0.5 - 0.6.1 | `julia = "1.7.0-1.10"` | `julia = "1.7.0-1.8"` |
| 0.6.2 - 0.6.15 | `julia = "1.8.0-1.10"` | `julia = "1.8"` |
| 0.6.21 - 0.7.14 | `julia = "1.8.1-1.10"` | `julia = "1.8"` |
| 0.7.15 - 0.8.21 | `julia = "1.9.0-1.10"` | `julia = "1.9"` |

This ensures the resolver picks the appropriate JET version for each Julia version (e.g., JET 0.8.22+ on Julia 1.10) instead of potentially selecting older, less optimal versions.

DilumAluthge:
These are always tough to review manually. CI is passing, which is good, but the CI checks only test that the compat entry is of a valid format - they don't check that the meaning of the compat entry is correct.

Ideally, we'd do a quick test before merging. E.g., check out this registry branch locally, and test that:
1. JET 0.9.0 is installable on Julia 1.10
2. JET 0.9.0 is not installable on Julia 1.11

ChrisRackauckas:
@aviatesk I found that these version caps are necessary because older JET just fails on newer Julia versions due to method table changes.

ChrisRackauckas:
> JET 0.9.0 is installable on Julia 1.10
> JET 0.9.0 is not install on Julia 1.11

It's not whether it installs, it's whether it's usable. JET v0.9 for example uses the old method table format. On Julia v1.12 this changed to `.mt` instead of whatever it was before, so JET v0.9 installs and loads on Julia v1.12 but any analysis will fail with this getproperty access missing because of the compiler change, so it's effectively unusable there. From Slack discussions @aviatesk noted that we were "using the wrong version" of JET in one of our tests and that's the reason for the failure, which we were then able to confirm. This change to the registry is then solidifying the upper bounds so that it's not possible for tests to use the wrong versions, which will help people diagnose the issues and is a requirement for Downgrade CI to work with JET.

ChrisRackauckas:
@ranocha this is the PR I mentioned for the downgrade.

DilumAluthge:
Sorry, I should clarify my ask. I want to test that this PR is working correctly. So by "installable", all I mean is whether or not Pkg allows the installation of that version of JET on that version to Julia, when you use the registry branch for this PR.

So "installable" (in my previous comment) means "Pkg installs the given version", and "not installable" means that "Pkg throws a resolver error".

ChrisRackauckas:
I'll get it tested.

ChrisRackauckas-Claude:
## Testing Results

I tested this PR with the modified registry on Julia 1.11.8.

### Test Setup
- Cloned the registry from `ChrisRackauckas-Claude/General:jet-julia-compat-caps`
- Created a fresh Julia depot with only the modified registry
- Tested JET installation and version resolution

### Results

**1. Version Resolution ✓**
```
Julia version: 1.11.8
Attempting to add JET...
   Resolving package versions...
⌅ [c3a54625] + JET v0.9.20
```
JET v0.9.20 was correctly selected, matching the constraint that v0.9.19+ requires Julia 1.11.

**2. Constraint Enforcement ✓**

Attempting to install JET v0.10 (which requires Julia 1.12) correctly fails:
```
Pkg.Resolve.ResolverError: restricted by julia compatibility requirements to versions: 0.8.28 - 0.9.20 or uninstalled — no versions left
```

Attempting to install JET v0.11 also correctly fails with incompatible dependency chain.

**3. Functionality ✓**
JET v0.9.20 loads and `@report_call` runs successfully.

### Verified Compat Changes
The key bounds in `Compat.toml` are working as intended:
- `["0.8.28 - 0.9.18"]`: `julia = "1.10"` (capped at 1.10)
- `["0.9.19"]` and `["0.9.20 - 0.9"]`: `julia = "1.11"` (capped at 1.11)  
- `["0.10.7 - 0"]`: `julia = "1.12"` (requires 1.12)

The PR correctly prevents the resolver from selecting incompatible JET versions. ✅

aviatesk:
LGTM. Thank you for investigating this compat bounds.

lgoettgens:
JET.jl commit https://github.com/aviatesk/JET.jl/commit/377fa5303a5f532755ece84c65fd1878b8fc2ab6 (which is equivalent to v0.11.3, but without the julia compat post-change) seems to work fine on julia 1.13.0-beta1. So maybe this restriction here is a bit too strict.
Please also note, that now there is no JET version usable on julia pre-releases (cf. https://github.com/aviatesk/JET.jl/issues/796, https://github.com/JuliaStats/StatsBase.jl/pull/992).

PR #145437, opened 2026-01-03
Add instructions for moving a package into a subdirectory

GunnarFarneback:
It happens now and then that people move packages into subdirectories and best practice should be documented.


giordano (dismissed):
Looks good, thanks.  Minor comment: it'd be caught by the automerge checks anyway, but is it worth pointing out that the new subdir needs to contain the license file in its top level?

GunnarFarneback:
Sounds like a good idea. Added.

giordano (dismissed):
I'll merge it tomorrow if there are no objections.

ericphanson (approved):

PR #145500, opened 2026-01-04
Deprecate CodecBGZF

jakobnissen:
This package is deprecated in favor of BGZFLib. The reason BGZFLib is not simply a breaking change of this CodecBGZF is that the "Codec*" prefix refers to the TranscodingStreams API, which the BGZF format does not easily fit into, which is what ultimately caused the hard-to-fix bugs.


DilumAluthge (approved):

DilumAluthge:

PR #146388, opened 2026-01-17
Revert "New version: ALPS_jll v1.1.2+0"

### LINKED ISSUE #12966: Duplicate name for ALPS
It turned out that this is another `ALPS_jll` here https://github.com/JuliaPackaging/Yggdrasil/blob/master/C/Coin-OR/ALPS/build_tarballs.jl beside https://github.com/JuliaPackaging/Yggdrasil/blob/master/A/ALPS/build_tarballs.jl

And this caused the conflicts in `JuliaRegistries`
https://github.com/JuliaRegistries/General/pull/146385/changes.

I am not sure how to solve the problem. And maybe in the future we should check the names conflicts?

Beforerr:
fix https://github.com/JuliaPackaging/Yggdrasil/issues/12966
see also https://github.com/JuliaPackaging/Yggdrasil/pull/12967


giordano:
Confirmed this was a mistake and no one should have used this version of this package in the short time it existed.

PR #147027, opened 2026-01-27
Update repo URL for Landfire.jl, to reflect repo transfer to org

joshday:
Changing the repo of Landfire.jl to where it now lives.


DilumAluthge (approved):
1. I have verified that the old URL redirects to the new URL in the browser.
3. The author of this PR is the same username that owns the old repo.

PR #147056, opened 2026-01-28
Update repo URL for JACC.jl package, to reflect transfer to different org

PhilipFackler:


DilumAluthge (approved):
1. I have verified that the old URL redirects to the new URL in the browser.
3. The Treecheck CI job is passing.

PR #39292, opened 2021-06-21
Cap the compatibility of all versions of StatPlots to Julia ≤ 1.6

DilumAluthge:
Fixes https://github.com/JuliaPlots/Plots.jl/issues/5518
Closes #38626 

This is an alternative to #38626.

This pull request caps the compatibility of all versions of `StatPlots` so that they are only compatible with Julia versions ≤ 1.6.

PR #147710, opened 2026-02-06
Metadata: Deprecate StatPlots (singular) in favor of StatsPlots (plural)

DilumAluthge:
Follow-up to #39292

See also: https://github.com/JuliaPlots/Plots.jl/issues/5518


DilumAluthge:
We already did a precursor to this (in #39292), but it would still be ideal to get a thumbs-up from a StatPlots.jl maintainer.

@mkborregaard @BeastyBlacksmith @daschw @piever @t-bltg Any objections?

BeastyBlacksmith:
That is a good idea, thank you

PR #147889, opened 2026-02-08
transfer NIDAQ.jl to JuliaDAQ

bjarthur:
see https://github.com/JuliaRegistries/General/pull/147227#issuecomment-3836066501

@goerz


goerz:
I don’t have merge permissions myself, but I’ll forward this to the Slack channel

PR #147891, opened 2026-02-08
JuliaFormatter: Yank versions 2.2.1 and 2.2.2

domluna:
Mark versions 2.2.1 and 2.2.2 as yanked in Versions.toml.

These two releases break precommit CI

### DIFF
--- a/J/JuliaFormatter/Versions.toml
+++ b/J/JuliaFormatter/Versions.toml
@@ -751,9 +751,11 @@ git-tree-sha1 = "ca9470360f51697fc1bde747760b2d99936619ea"
 
 ["2.2.1"]
 git-tree-sha1 = "eeb150b63d8bfdca04a0c367a8b95d3863fdc892"
+yanked = true
 
 ["2.2.2"]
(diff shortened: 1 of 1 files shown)


DilumAluthge:
So, in previous times, I would just have merged this.

However, recently, I have heard feedback from multiple folks that they find yanks to be disruptive to their workflows. A common form of this feedback is along the lines of this: a user has a CI workflow that uses a manifest file checked into source control, and the CI job was working fine yesterday, and today CI has suddenly broken, with no changes made by the user, and the culprit turns out to be that a yank was recently merged into the General registry.

@domluna In this case, are there other (non-yanking) steps that we could take, in order to achieve your goals?

DilumAluthge:
[for now, I'll convert this to "draft" status, to prevent accidental merging]

DilumAluthge:
Bump @domluna - can you take a look at my comment above?

domluna:
2.3 is tagged with the breaking changes removed so we can probably just close this.

PR #148150, opened 2026-02-11
Update Powsybl package github URL

obrix:
Update the Powsybl package github URL. 
Just remove the upper case P and replace it by a lower case. This is blocking the registrator bot who is detecting an URL change.


goerz:
You should rename the repo instead, so that it correctly matches the package name (with the uppercase).

I’m going to tentatively close this

PR #149351, opened 2026-02-26
Move the naming guidelines into this repo (from the Pkg docs)

DilumAluthge:
Here's the corresponding PR to remove the naming guidelines from the Pkg docs. See that PR for the motivation.
- [x] https://github.com/JuliaLang/Pkg.jl/pull/4626

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ You can register your package!
 See [Registering a package in General](https://github.com/JuliaRegistries/General#registering-a-package-in-general) in the README for how to do that.
 The "FAQ" section in the README helps answer many more questions, like [do I need to register a package to install it?](https://github.com/JuliaRegistries/General#do-i-need-to-register-a-package-to-install-it), [should I register my package?](https://github.com/JuliaRegistries/General#should-i-register-my-package), and more.
 
-* Please be aware of the [package naming guidelines](https://pkgdocs.julialang.org/v1/creating-packages/#Package-naming-rules)
+* Please be aware of the [package naming guidelines][naming-guidelines]
 * We strongly encourage authors to follow best practices like having documentation (or a descriptive README), tests, and continuous integration.
 
 ## As a Julia community member
@@ -50,7 +50,7 @@ Any comment without `[noblock]` included in it will block AutoMerge from automat
     Since registration is a mechanism to share code with the whole Julia community, such a description is important for the package to be useful.
     While we don't strictly require such documentation, it can help to give a polite and gentle nudge in the PR comments, or show folks how to write documentation and/or what is helpful to include in a README.
     We want to encourage best practices (in an inclusive and friendly way!) even when they are not strict requirements.
-    * Sometimes package names are possibly confusing or don't conform to our [naming guidelines](naming-guidelines), but AutoMerge does not detect this.
+    * Sometimes package names are possibly confusing or don't conform to our [naming guidelines][naming-guidelines], but AutoMerge does not detect this.
     Feel free to comment, describing what you think is confusing or non-compliant about the current name, and any suggestions you have for a more clear name.
 
 ### Other PRs to General
@@ -166,6 +166,6 @@ check_package_versions("FastParzenWindows", ".")
 ```
 
 [FAQ]: https://github.com/JuliaRegistries/General#faq]
-[naming-guidelines]: https://pkgdocs.julialang.org/dev/creating-packages/#Package-naming-guidelines-1
+[naming-guidelines]: ./NAMING_GUIDELINES.md
 [permissions]: https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-permission-levels-for-an-organization#permission-levels-for-repositories-owned-by-an-organization
 [RegistryCI]: https://github.com/JuliaRegistries/RegistryCI.jl/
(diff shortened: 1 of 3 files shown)


DilumAluthge:
We'll merge this PR first, and then we'll merge the Pkg PR.

goerz:
The list items in `NAMING_GUIDELINES.md` would benefit from being individually linkable (e.g. for use in review comments like "see rule 6"). Since turning them into headings would break the document structure, a lightweight approach might be to add an empty inline anchor before each item:

```markdown
1. <a id="rule-avoid-jargon"></a>Avoid jargon. In particular...
2. <a id="rule-avoid-julia"></a>Avoid using `Julia` in your package name...
3. <a id="rule-pluralize"></a>Packages that provide most of their functionality...
4. <a id="rule-clarity"></a>Err on the side of clarity...
5. <a id="rule-less-systematic"></a>A less systematic name may suit...
6. <a id="rule-wrappers"></a>Packages that wrap external libraries...
7. <a id="rule-avoid-close-names"></a>Avoid naming a package closely...
8. <a id="rule-avoid-distinctive-names"></a>Avoid using a distinctive name...
9. <a id="rule-style"></a>Packages should follow the Stylistic Conventions...
10. <a id="rule-automerge"></a>For the complete list of rules...
```

I _think_ this works on GitHub, and that it renders `<a id="...">` tags in Markdown, so each rule would then be directly linkable as e.g. `NAMING_GUIDELINES.md#rule-wrappers`.

Do I have permissions to add a commit to this PR? If so, let me try out if this works...

DilumAluthge:
Yeah, go ahead and push to my branch

goerz:
Pushed, alongside some typo/formatting fixes. But I'm not sure if it actually works, or if GitHub strips out the anchors… At least https://github.com/JuliaRegistries/General/blob/9b31e859965a5b13183f856fb23bceea002198a6/CONTRIBUTING.md#rule-clarity doesn't jump to the right location to me. But let me test this out in a proper repo… (stay tuned)

DilumAluthge:
> At least https://github.com/JuliaRegistries/General/blob/9b31e859965a5b13183f856fb23bceea002198a6/CONTRIBUTING.md#rule-clarity doesn't jump to the right location to me.

You're in the wrong file - you need to link to `NAMING_GUIDELINES.md`

goerz:
Duh.. right. Jumping _does_ work (also in my proper test at https://github.com/goerz-testing/JuliaPackageNamingGuidelines/blob/master/NAMING_GUIDELINES.md#rule-clarity). But there's something off with how the item numbers are rendered (not in front of the first line): 
CleanShot 2026-03-04 at 14 51 07@2x


**EDIT:** That's in Safari. Chrome seems to render correctly.

DilumAluthge:
Hmm. Can you try in a different browser? It looks fine to me on Firefox:

Screenshot 2026-03-04 at 14 53 01

goerz:
Yes, you're right: This seems to be specific to Safari. Still annoying, let me see if it's fixable

DilumAluthge:
Okay I see the problem in Safari too.

goerz:
Alright, so there seem to be a few approaches to fix this. One is [interspersing the anchors](https://github.com/goerz-testing/JuliaPackageNamingGuidelines/commit/d0dcd1bfcb559a15afcfc0e49c99aa4ad2317592):

https://github.com/goerz-testing/JuliaPackageNamingGuidelines/blob/d0dcd1bfcb559a15afcfc0e49c99aa4ad2317592/NAMING_GUIDELINES.md#rule-clarity

A bit hacky, but seems to work in all browsers. This is the "silent anchor" that I was going for originally.

A possibly better alternative would be to make the anchors visible:

https://github.com/goerz-testing/JuliaPackageNamingGuidelines/blob/26850cf6833b178e372dd83ec21cc01f5d129ce4/NAMING_GUIDELINES.md#rule-clarity

I can use other symbols besides `§`. I was considering 🔗 because I thought that's what GitHub uses for hovering over titles, but unfortunately, what they're using is an SVG instead. It would have been nice to put the `§` in front of the item numbers, but that doesn't seem possible. I could also put it at the end of the first paragraph. Or, we could linkify the first sentence… Maybe let me play around with that, too.

@DilumAluthge Any thoughts / preferences?

DilumAluthge:
https://github.com/goerz-testing/JuliaPackageNamingGuidelines/blob/26850cf6833b178e372dd83ec21cc01f5d129ce4/NAMING_GUIDELINES.md#rule-clarity (with the § symbol) looks fine to me.

goerz:
What do you think about https://github.com/goerz-testing/JuliaPackageNamingGuidelines/blob/7cafbeb8b4a38e7f5471a7c943ef578db26124e7/NAMING_GUIDELINES.md

Better or worse than the linkified §  symbol?

DilumAluthge:
Hmmm. Maybe a little worse, because it makes it look like those are clickable links, but they don't go anywhere.

goerz:
They go to themselves (same as the `$` links)

The point being that this provides a way to _get_ the link in the first place without inspecting the source

But I'm definitely open to committing the `$` version. Kinda on the fence myself.

DilumAluthge:
I guess I just visually prefer the `§` version - with those, it's more clear to me that it'll just jump to itself.

I don't feel too strongly either way.

goerz:
Alright, done! If other reviewers have strong opinions on this, we can revise it further.

DilumAluthge:
Yeah we can always make follow-up PRs.

goerz (dismissed):

goerz (approved):

DilumAluthge:
It would be ideal to have another approval as well, since you and I both committed to this PR.

ericphanson (approved):
lgtm

PR #150265, opened 2026-03-11
Update BlackBoxOptim.jl's url to SciML

ChrisRackauckas:


oscardssmith (approved):

goerz:
@robertfeldt Can you confirm?

oscardssmith:
It should be noted that https://github.com/robertfeldt/BlackBoxOptim.jl was transferred to sciml (i.e. this isn't a fork, it's the original repository moved to an organization which requires the original author to approve)

DilumAluthge (approved):
Here is the checklist that I usually use for PRs like this:

1. [x] I have confirmed that the old URL automatically redirects to the new URL in the web browser.
2. [x] The PR preserves the trailing `.git` at the end of the URL.
3. [x] The Treecheck CI job ran on this PR and is green (indicating that all treehashes are available from the new repo URL).

goerz (approved):

DilumAluthge:
Yes, in the case where the old URL automatically redirects to the new URL in the web browser (as in this case), and where the old URL was owned by a user (not an organization), I usually don't wait for confirmation from the owner of the old URL.

In this case, my justification is something like this: The old URL was owned by a user (not an organization). The fact that GitHub automatically redirects the old URL to the new URL indicates that the repo was transferred from the user to the organization. On GitHub, when a repo is owned by a user, the only person that can transfer the repo is the old owner.[^1]

So this tells us that @robertfeldt was the one that chose to transfer the repo to the SciML org, and thus we already know that they chose to make this change.

With that all being said, if @robertfeldt were to come into this PR and comment that they didn't actually want to transfer the repo, we of course have the ability to revert this PR, if necessary.

[^1]: When a repo is owned by a user, then only the owner can transfer the repo. Other users that have commit (write) access will not be able to transfer it.

DilumAluthge:
It might be helpful for me to document my checklist, so that other registry maintainers can use it too: https://github.com/JuliaRegistries/General/pull/150273

We can also edit/expand the checklist over time.

robertfeldt:
I confirm/approve 👍

PR #150490, opened 2026-03-14
UnicodePlots: Retroactively set Unitful compat on versions 2.9 – 3.8.0

jmert:
As mentioned in JuliaPlots/UnicodePlots.jl#430, when Unitful support was added, the implementation accidentally depended on functionality that assumed Unitful v1.6+. (JuliaPlots/UnicodePlots.jl#431 adjusts the implementation to restore compatibility with earlier versions of Unitful.)

In order to support resolving valid/functional environments (e.g with the [julia-downgrade-compat][] action) across all versions of UnicodePlots, we can retroactively correct the compat range to reflect the true package requirements.

[julia-downgrade-compat]: https://github.com/julia-actions/julia-downgrade-compat

@t-bltg


t-bltg:
I'm in favor of this PR, but I have no idea if this is the correct approach or not.

Maybe @giordano can review / comment ?

DilumAluthge (approved):
On manual review, this looks okay. However, manual review is usually insufficient for these kinds of PRs. And our existing CI checks are also usually insufficient for these kinds of PRs.

The only way to really know if this PR works is to try the branch out locally, to see if the desired effect is indeed observed when using this branch locally.

Alternatively, if manaul review looks okay, and if CI is green, we can try merging. And we can always undo the changes if we observe breakages in the wild.

DilumAluthge:
Let's give this a try. If we notice breakage in the wild, we can undo these changes.

PR #150497, opened 2026-03-14
Revert "New package: QRacahSymbols v1.0.0"

goerz:
Reverts JuliaRegistries/General#150218


goerz:
That package wasn’t supposed to be registered. I had forgotten about a “ignore blocking comments” override. This is especially unfortunate for a 1.0 registration, so I hope we can fix this by reverting the registration.

goerz:
The registration has already made it onto the package servers, so reverting is no longer a good option.

Closing in favor of https://github.com/JuliaRegistries/General/pull/150509 (yanking)

PR #150509, opened 2026-03-14
Yank QRacahSymbols v1.0.0

goerz:
See https://github.com/JuliaRegistries/General/pull/150218

Since the `v1.0.0` release has already made it to the package servers, we can no longer revert. The result of a discussion on Slack was that yanking was the best course of action.


DilumAluthge (approved):

goerz:
@sethkasante Let me know if you want me to merge this yank, cf. https://github.com/JuliaRegistries/General/pull/150218#issuecomment-4059426652

goerz:
Merging, as requested and confirmed in https://github.com/JuliaRegistries/General/pull/150218#issuecomment-4059571105

PR #150728, opened 2026-03-17
Yank Bioinformatics v0.1.0

mbauman:
This git tree sha is not in either https://github.com/mrtkp9993/Bioinformatics.jl or https://github.com/UnofficialJuliaMirrorSnapshots/Bioinformatics.jl-c429c406-9079-5a14-8339-dedc8835de90 or in the public storage servers, so it's not installable in any case.

### DIFF
--- a/B/Bioinformatics/Versions.toml
+++ b/B/Bioinformatics/Versions.toml
@@ -1,5 +1,6 @@
 ["0.1.0"]
 git-tree-sha1 = "6b879cd7f84a18d845ba6dbebaa7df6e4dd8e440"
+yanked = true
 
 ["0.1.1"]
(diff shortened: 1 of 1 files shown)


mbauman:
Oh interesting, I can't install v0.1.1 either.  I think the reconstruction of the git repo here ended up breaking these tree shas?

```
(@temp) pkg> add Bioinformatics
   Resolving package versions...
    Updating git-repo `https://github.com/mrtkp9993/Bioinformatics.jl.git`
ERROR: Bioinformatics: git object 6f4646f277261658185f474e6e826cad14301b2a could not be found
```

mbauman:
This is a bigger problem than I intended to bite off here.  There are many such versions/packages, and if we're gonna take action we should probably do so holistically.

PR #150798, opened 2026-03-18
Yank JumpProcesses 9.23.2

isaacsas:
This should not have been released as it contains breaking changes for v10.0 that are not completed yet.


DilumAluthge:
In the past, I would have just merged this. But recently, folks in the community have given feedback that yanks can be disruptive to them.

Is there another way that we can accomplish the same goal? For example, could you revert all the breaking changes and make a patch release 9.32.3?

PR #150984, opened 2026-03-21
Trigger Treecheck on `NeuroAnalyzer.jl`

DilumAluthge:


DilumAluthge:
This will need to be fixed in the package's repo: https://codeberg.org/AdamWysokinski/NeuroAnalyzer.jl/pulls/33

PR #151660, opened 2026-03-30
CI: Remove the old manifests (`Manifest.1.x.toml`), but keep the new manifests (`Manifest-v1.x.toml`)

DilumAluthge:


DilumAluthge:
[not stale]

PR #151856, opened 2026-04-01
Migrate NaturallyUnitful.jl to JuliaPhysics organization

MasonProtter:
I am the original author of NaturallyUnitful, but we're migrating it for maintainability reasons.


goerz (changes requested):
Usually, the actual repo should be transferred

MasonProtter:
@goerz okay, I have changed it around so now it's a transfer of the repo.

goerz (approved):

PR #151981, opened 2026-04-02
Retroactively tighten OrdinaryDiffEqCore compat for LowOrderRK and Verner

ChrisRackauckas:
## Summary

Retroactively tighten `OrdinaryDiffEqCore` compat bounds for versions of `OrdinaryDiffEqLowOrderRK` and `OrdinaryDiffEqVerner` that introduced `lazy::Val{Bool}` (from PR SciML/OrdinaryDiffEq.jl#3044) but didn't require Core >= 3.5.2 (which added `_unwrap_val`).

Without this fix, the resolver can pick Core < 3.5.2 with these versions, producing:
```
MethodError: no method matching !(::Val{true})
```

See SciML/OrdinaryDiffEq.jl#3253 for the original issue and SciML/OrdinaryDiffEq.jl#3327 for the source fix.

### Changes

| Package | Versions affected | Change |
|---------|-------------------|--------|
| OrdinaryDiffEqLowOrderRK | 1.11.0, 1.12.0 | `OrdinaryDiffEqCore` tightened from `"3"` to `"3.5.2 - 3"` |
| OrdinaryDiffEqVerner | 1.12.0, 1.13.0 | `OrdinaryDiffEqCore` tightened from `"3"` to `"3.5.2 - 3"` |

New versions (LowOrderRK 1.13.0 and Verner 1.14.0) with correct bounds have been registered via SciML/OrdinaryDiffEq.jl#3327.

### DIFF
--- a/O/OrdinaryDiffEqLowOrderRK/Compat.toml
+++ b/O/OrdinaryDiffEqLowOrderRK/Compat.toml
@@ -17,11 +17,12 @@ OrdinaryDiffEqCore = ["0.0.0", "1"]
 ["1.1 - 1.2"]
 OrdinaryDiffEqCore = "1.1.0-1"
 
-["1.10 - 1"]
+["1.10"]
 OrdinaryDiffEqCore = "3"
 
 ["1.11 - 1"]
 DiffEqBase = "6.194.0 - 6"
+OrdinaryDiffEqCore = "3.5.2 - 3"
 SciMLBase = "2.116.0 - 2"
 
 ["1.12 - 1"]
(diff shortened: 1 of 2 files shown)


ViralBShah (approved):

DilumAluthge (approved):
On manual review, this looks okay. However, manual review is usually insufficient for these kinds of PRs. And our existing CI checks are also usually insufficient for these kinds of PRs.

The only way to really know if this PR works is to try the branch out locally, to see if the desired effect is indeed observed when using this branch locally.

Alternatively, if manaul review looks okay, and if CI is green, we can try merging. And we can always undo the changes if we observe breakages in the wild.

PR #152434, opened 2026-04-08
Move ThermoCycleGlides from Sush1090 to ClapeyronThermo

longemen3000:
CC @sush1090

### DIFF
--- a/T/ThermoCycleGlides/Package.toml
+++ b/T/ThermoCycleGlides/Package.toml
@@ -1,3 +1,3 @@
 name = "ThermoCycleGlides"
 uuid = "58ad84e6-b92d-4139-8d7e-462257ec6cf0"
-repo = "https://github.com/Sush1090/ThermoCycleGlides.jl.git"
+repo = "https://github.com/ClapeyronThermo/ThermoCycleGlides.jl.git"


goerz:
It would be good if https://github.com/Sush1090/ThermoCycleGlides.jl.git forwarded to https://github.com/ClapeyronThermo/ThermoCycleGlides.jl, as should be the case when a repo is moved. Hopefully, if @Sush1090 deletes their fork, that will restore the forward.

@Sush1090 should have commit access to https://github.com/ClapeyronThermo/ThermoCycleGlides.jl, so that they can push to branches in the original repo, instead of via a fork

Sush1090:
Hello @goerz,

I have deleted my fork so now the forward should be automatic? Or I need to do something else as well?

goerz:
I had hoped deleting the fork would restore the forwarding, but unfortunately it seems like it doesn't. The original https://github.com/Sush1090/ThermoCycleGlides.jl.git now just gives a 404. Seems like forking back a repo is the same as creating a new project with the same name, which breaks the forwarding. I suppose the lesson is that when someone transfers a project to an org, they absolutely should not fork it back to their account (not that you could have known this).

Oh well… So I don't think there's an easy way to fix this. The only possibility might be for @Sush1090 to contact GitHub support and ask them to restore the forwarding.

But it any case, I can just merge this PR now.

If you don't want to contact GitHub supports, I don't think there's any harm if you want to re-fork the project; maybe that's even better than having a 404, since at least it allows visitor's to manually click to the new "official" repo in the org.

goerz (approved):

PR #152668, opened 2026-04-11
Yank RestrictedBoltzmannMachines v0.5.0–v0.33.1 (bloated git history)

cossio:
## Summary

Yank all versions of RestrictedBoltzmannMachines from v0.5.0 to v0.33.1 (inclusive).

These versions contain a bloated git history (large binary/data files were committed to the repository), causing `Pkg.add` / `Pkg.update` to download an unnecessarily large amount of data. The issue was fixed starting from v0.34.0.

Yanking these versions will steer users toward the newer, lighter versions while keeping the registry entries available for reproducibility.

## Changes

- Added `yanked = true` to 58 version entries in `R/RestrictedBoltzmannMachines/Versions.toml` (v0.5.0 through v0.33.1; v0.14.0 was already yanked).


cossio:
I have checked that no registered packages (even older versions) depend on the yanked versions in this PR.

cossio:
UPDATE: do not merge yet. I'm still discussing with @goerz what's the best approach here.

PR #152830, opened 2026-04-13
HDF5_jll: Set Julia compat to 1.10 for HDF5_jll 2.0.0

mkitti:
Update Julia version compatibility from 1.6.0 to 1.10.0 for HDF5_jll v2.0.0+

On macOS and Linux:
```
$ julia +1.6 --project=. -E "using Pkg; Pkg.add(name=\"HDF5_jll\", version=\"2.0.0\"); using HDF5_jll; HDF5_jll.is_available()"
false
```

Only Windows seems to be supported on HDF5_jll v2 and Julia 1.6:
https://github.com/JuliaIO/HDF5.jl/actions/runs/24325249609/job/71019153490

cc: @eschnett


DilumAluthge:
I'm a little confused by the CI job you linked there. Did you mean to link a different job? That job just says that Julia 1.6 doesn't have native builds for Apple Silicon. You can fix that error by specifically running the job on Intel macOS, e.g. `runs-on: macos-15-intel`

DilumAluthge:
Did you mean to link https://github.com/JuliaIO/HDF5.jl/actions/runs/24325249609/job/71019153478 instead?

mkitti:
Yes. I'm starting to question this course of action since technically HDF5_jll v2.0.0 does work on Windows with Julia 1.6.

eschnett:
We built `HDF5_jll 2.0.0` only for Julia 1.10 and upwards. Julia 1.6 is not support by this jll library. I don't recall the exact reason, but it had to do with some of the dependencies – something like the fact that we don't build for libgfortran and libgfortran4 and cxx03 any more. I would probably be possible to allow Julia 1.9 or 1.8, but there is a general trend towards supporting only 1.10 when building jlls if supporting older Julia versions requires a significant effort.

mkitti:
The problem that I am trying to fix here is that HDF5_jll v2.0.0 says it is compatible with Julia 1.6.

```
% julia +1.6 -e "using Pkg; Pkg.add(name=\"HDF5_jll\", version=\"2\"); using HDF5_jll; println(HDF5_jll.is_available())"
   Resolving package versions...
   Installed Xorg_libpciaccess_jll ─ v0.18.1+0
   Installed HDF5_jll ────────────── v2.0.0+2
# ...
false
```

mkitti:
The compat with Julia 1.10 was set in https://github.com/eschnett/Yggdrasil/commit/61a0109b80d702f00124ee4ad2ddea1fe148507b 

This mean that the initial HDF5_jll 2.0.0 builds had Julia 1.6 compat. However, they did not actually provide compatible artifacts except on Windows.

For HDF5_jll v2.0.0+2, there is a `julia = "1.6"` compat declaration: 
https://github.com/JuliaBinaryWrappers/HDF5_jll.jl/blob/8187d2936969b5b5c068207f893d8b4e88008edd/Project.toml#L34

@eschnett, can you confirm the Julia 1.6 compat for HDF5_jll v2.0.0 was not your intention?

eschnett:
I don't recall the details. I think that `HDF5_jll 2.0.0` has artifacts that are in principle compatible with Julia 1.6, but only if one builds Julia 1.6 from source using libgfortran5. If I recall correctly, a standard Julia binary from `juliaup` would require an artifact that we did not build. It's unfortunate that the automatic checks don't catch these "minor details". If the required artifacts are not present then the best way forward is to add a `julia="1.10"` compat entry for this jll version.

giordano:
Looks good, but should afterwards mirrored in Yggdrasil with a [skip ci] PR, otherwise this change will be overrriden.

mkitti:
@giordano the current recipe already has a 1.10 lower bound:
https://github.com/JuliaPackaging/Yggdrasil/blob/5503e8486ee98ef3d63fd731c96c4cc043d8920f/H%2FHDF5%2Fbuild_tarballs.jl#L335

mkitti:
@DilumAluthge could we merge this?

DilumAluthge (approved):

PR #154175, opened 2026-04-28
Narrow OrdinaryDiffEqDifferentiation 2.8 floor to OrdinaryDiffEqRosenbrock 1.30+

devmotion:
## Summary

Follow-up to #154061. That PR raised `OrdinaryDiffEqDifferentiation` to `"2.8"` for the entire `OrdinaryDiffEqRosenbrock` `["1.23 - 1"]` range, but the OOP `calc_rosenbrock_differentiation` symbol that requires `OrdinaryDiffEqDifferentiation 2.8.0` is only **called** starting in `OrdinaryDiffEqRosenbrock 1.30.0`. Versions 1.23.0–1.29.0 only call the bang variant `calc_rosenbrock_differentiation!`, which is defined in every 2.x release of `OrdinaryDiffEqDifferentiation`.

This narrows the `"2.8"` floor to apply only to Rosenbrock 1.30+ and restores the original `"2"` floor for 1.23–1.29.

```diff
 ["1.23 - 1"]
 DiffEqBase = "6.194.0 - 6"
-OrdinaryDiffEqDifferentiation = "2.8"
 SciMLBase = "2.116.0 - 2"
+
+["1.23 - 1.29"]
+OrdinaryDiffEqDifferentiation = "2"
+
…
+["1.30 - 1"]
+OrdinaryDiffEqDifferentiation = "2.8"
```

`DiffEqBase` and `SciMLBase` floors are unchanged across the range, so they stay in the original `["1.23 - 1"]` block; only the `OrdinaryDiffEqDifferentiation` axis splits at the 1.29/1.30 boundary.

## Verification

OOP `calc_rosenbrock_differentiation` was introduced by [SciML/OrdinaryDiffEq.jl#3075](https://github.com/SciML/OrdinaryDiffEq.jl/pull/3075) (commit `689d0b79`, 2026-04-11), which **adds** both the OOP definition (in `lib/OrdinaryDiffEqDifferentiation`) and three OOP call sites (in `lib/OrdinaryDiffEqRosenbrock`) — additive, not replacing the bang variant. The coordinated version-bump commit `dff8c622` shipped this as **Differentiation 2.8.0** + **Rosenbrock 1.30.0**, so the OOP call only exists from Rosenbrock 1.30.0 onwards.

`OrdinaryDiffEqDifferentiation 2.8.0` defines both the bang and OOP variants (additive change in #3075), so 1.23–1.29 callers (bang-only) are compatible with both `2` (pre-2.8) and `2.8+` of Differentiation. The narrower split preserves resolvability for downstream packages that pin Rosenbrock in 1.23–1.29 without weakening protection for 1.30+ where the OOP call actually exists.

## Impact

This affects downstream packages that pin `OrdinaryDiffEqRosenbrock` in the 1.23–1.29 range with combinations that are runtime-correct but newly unresolvable after #154061. [Pumas](https://github.com/PumasAI/Pumas.jl) had its CI broken on every shard at `Pkg.instantiate()` with `Unsatisfiable requirements detected for package OrdinaryDiffEqDifferentiation`, **and** an in-progress development release of Pumas became uninstallable for the same reason. This narrowing restores resolvability for both.

cc @DilumAluthge @ChrisRackauckas

---

This PR was prepared with the assistance of generative AI; the analysis (commit-walk on SciML/OrdinaryDiffEq.jl, identification of the introducing version) and the proposed split were AI-generated and human-reviewed.

### DIFF
--- a/O/OrdinaryDiffEqRosenbrock/Compat.toml
+++ b/O/OrdinaryDiffEqRosenbrock/Compat.toml
@@ -117,14 +117,19 @@ OrdinaryDiffEqCore = "3"
 
 ["1.23 - 1"]
 DiffEqBase = "6.194.0 - 6"
-OrdinaryDiffEqDifferentiation = "2.8"
 SciMLBase = "2.116.0 - 2"
+
+["1.23 - 1.29"]
+OrdinaryDiffEqDifferentiation = "2"
 
 ["1.28 - 1"]
 RecursiveArrayTools = "3.52.0 - 3"
 
 ["1.28 - 2"]
 FastBroadcast = "1.3.0 - 1"
+
+["1.30 - 1"]
+OrdinaryDiffEqDifferentiation = "2.8"
 
 ["1.4 - 1.11"]
 ADTypes = "1.11.0-1"


ChrisRackauckas (approved):

ChrisRackauckas:
Oddly specific. Let me know if that didn't work

devmotion:
It seems to work.

We pin all SciML packages to a snapshot (currently end of March) to avoid breakage due to SciML updates which happened multiple times in the past. Since it's so disruptive even for existing releases of downstream packages and existing environments, IMO yanking releases or changing compat entries in the registry should really be a last resort - and better be done correctly.

O/OrdinaryDiffEqRosenbrock/Compat.toml
```diff
@@ -117,15 +117,20 @@ OrdinaryDiffEqCore = "3"
...
+OrdinaryDiffEqDifferentiation = "2"
+
 ["1.28 - 1"]
 RecursiveArrayTools = "3.52.0 - 3"
 
 ["1.28 - 2"]
 FastBroadcast = "1.3.0 - 1"
 
+["1.30 - 1"]
+OrdinaryDiffEqDifferentiation = "2.8"
```
bauglir, line 132: It seems like this should have been
```suggestion
OrdinaryDiffEqDifferentiation = "2.8 - 2"
```

This current value which gets parsed through `VersionSpec` into a `VersionRange` into a `VersionBound` results in only v2.8.z versions being allowed when resolving, thereby excluding v2.9.0 which is also available.

This is different behavior from what gets defined in a `Project.toml` which parses through `semver_spec` and would have translated into "2.8 - 2" under the hood.

DilumAluthge: @ChrisRackauckas @devmotion 

If allowing 2.9.0 is desirable, I can merge a PR.

ChrisRackauckas: I was assuming @devmotion did a thorough history analysis to end up at the 2.8 number, so I'll wait for him to respond why Joris is saying 2.9 and he's saying 2.8

bauglir: @devmotion didn't come up with the 2.8 number, he finetuned the bounds on `DiffEqRosenbrock` to relax the constraint for versions where the more tight requirement was not necessary. The 2.8 number comes from your original PR #154061. This states

> ## Fix
> Tighten the lower bound from "2" to "2.8":
>
> ```diff
>  ["1.23 - 1"]
>  DiffEqBase = "6.194.0 - 6"
> -OrdinaryDiffEqDifferentiation = "2"
> +OrdinaryDiffEqDifferentiation = "2.8"
>  SciMLBase = "2.116.0 - 2"
> ```
> This is a retroactive compat tightening only — no versions are yanked. The resolver will simply exclude OrdinaryDiffEqDifferentiation v2.0–v2.7 when selecting Rosenbrock v1.23+.

So the stated goal here is to tighten the lower bound to v2.8 and to eliminate v2.0-v2.7 from being selected for versions of `OrdinaryDiffEqRosenbrock` after v1.30. However, that syntax does _not_ implement that goal. It bounds the allowed versions to v2.8.z which doesn't allow later versions in the v2 range and therefore removes v2.9.0 from the resolvable list which breaks version resolution in our product.

devmotion: My PR fixed the incorrectly restrictive compat bounds for versions < 1.30. I didn't touch this bound (and generally tried to keep changes as minimal as possible). AFAICT there is no reason to restrict >=1.30 to 2.8.x, it should be 2.8 - 2 for >= 1.30.

asinghvi17: Fixed in https://github.com/JuliaRegistries/General/pull/154451

PR #154190, opened 2026-04-28
Yank t8code v0.7.5

benegee:
This should have been a breaking release.


JoshuaLampert:
Instead of yanking the version, let's revert the change, release a v0.7.6 (which is the same as v0.7.4), apply the change again, and release a v0.8.0. So this can be closed.

PR #154474, opened 2026-04-30
Yank ReachabilityAnalysis v0.30.3

schillic:
x-ref https://github.com/JuliaRegistries/General/pull/154473 / https://github.com/JuliaReach/ReachabilityAnalysis.jl/pull/1077


schillic:
As per discussion on slack, the suggestion is to rather make a new minor release 0.30.4 that reverts the offending change.

PR #154562, opened 2026-05-01
EarthBox.jl: Update Package.toml for EarthBox.jl after transfer

geonomylabs-dev:
Updates the registered repository URL for EarthBox.jl after transferring the repository from eakneller to geonomylabs.

The old URL redirects, but this makes the General registry metadata point to the canonical repository location.


goerz:
Can you make sure that the owners of the geonomylabs org [set their membership visibility to "public"](https://docs.github.com/en/account-and-profile/how-tos/organization-membership/publicizing-or-hiding-organization-membership)? This is so that there is a point of contact for issues with packages in the org. Ideally, an org should have two or more owners to increase the [bus factor](https://en.wikipedia.org/wiki/Bus_factor)

goerz (approved):

DilumAluthge (approved):
Applying my usual checklist:

1. [x] I have confirmed that the old URL automatically redirects to the new URL in the web browser.
2. [x] The PR preserves the trailing `.git` at the end of the URL.
3. [x] The Treecheck CI job ran on this PR and is green.

eakneller:
Thank you for  looking into this. I have made the geonomylabs org owner membership public.

PR #155435, opened 2026-05-14
CI: Remove the "Manual PRs feed" workflow

DilumAluthge:


goerz:
I don't think I have any context for this… What's this about?

goerz (approved):

DilumAluthge:
[discussed out of band]

DilumAluthge:
Note: This only affects the "manual PRs" feed (`#pkg-registration-manual-prs`).

It does not affect the `#new-packages-feed`, which is what I think most people are interested in.

imciner2:
Out of band discussion doesn't really help explain to those of us who have watched that channel why this was removed and what is going to happen now. Can this please be explained better?

goerz:
There was a security concern that might have allowed an attacker to compromise the registry. Even if unlikely, @DilumAluthge felt that it was better to err on the side of caution and remove this particular workflow. The "out of band" was because we didn't want to discuss the potential security issue in public before closing it. @DilumAluthge can probably get into more details on this, now that it's closed.

Potentially, we could restore the feed if it can be done securely

DilumAluthge:
Yeah, that's basically it. There have been several notable repository compromises recently, all of which seem to have started by exploiting `pull_request_target`.

To be clear, I think that our usage (in the General registry) of `pull_request_target` was safe. However, given all of the recent compromises, I think `pull_request_target` is too much of a footgun, and I don't think we can justify continuing to use it. That's why I wanted to remove the usage of `pull_request_target` from this repo.

It would be great to figure out a different way to keep the "manual PRs" feed (`#pkg-registration-manual-prs`) working.

PR #155453, opened 2026-05-14
CUDA: Add artificial weakdep to prevent incompatible installations.

### LINKED ISSUE #3130: cuDNN 6 + CUDA 5 resolves but breaks at runtime: `expected CUDACore.TaskLocalState, got CUDA.TaskLocalState`
## Summary

`cuDNN` 6.x has no `CUDA` compat entry, so the Julia resolver happily installs **`CUDA` 5.x + `cuDNN` 6.x** side-by-side. At runtime that combination errors with `expected CUDACore.TaskLocalState, got a value of type CUDA.TaskLocalState`, because CUDA 5 and CUDACore 6 each define their own nominally-distinct `TaskLocalState` struct but share the same `:CUDA` task-local-storage key.

Two user reports in the wild that I know of:

- https://github.com/SciML/NeuralOperators.jl/issues/125#issuecomment-4355798562
- https://github.com/LuxDL/Lux.jl/issues/1245

## Reproduction

User environment from NeuralOperators.jl#125:

```
[052768ef] CUDA v5.10.1
[02a925ec] cuDNN v6.0.0
[b2108857] Lux v1.31.3
[872c559c] NNlib v0.9.33
```

Calling a Lux model that touches NNlib's cuDNN-backed conv:

```julia
using CUDA, cuDNN, Lux
fno = FourierNeuralOperator(gelu; chs=(1, 64, 64, 128, 1), modes=(16,))
ps, st = Lux.setup(rng, fno) |> cu
Lux.apply(fno, m_sample, ps, st)
```

errors with:

```
ERROR: TypeError: in typeassert, expected CUDACore.TaskLocalState,
                  got a value of type CUDA.TaskLocalState
Stacktrace:
  [1] task_local_state!() @ CUDACore .../lib/cudadrv/state.jl:73
  [2] math_mode
  ...
  [4] cuDNN.cudnnConvolutionDescriptor(...) @ NNlibCUDACUDNNExt
```

Downgrading `cuDNN` to 1.x works around it.

## Mechanism

`CUDACore/lib/cudadrv/state.jl` (master) and `lib/cudadrv/state.jl` at `v5.10.1` define essentially the same code:

```julia
mutable struct TaskLocalState
    device::CuDevice
    context::CuContext
    streams::Vector{Union{Nothing,CuStream}}
    math_mode::MathMode
    math_precision::Symbol
    ...
end

function task_local_state!(args...)
    tls = task_local_storage()
    if haskey(tls, :CUDA)
        validate_task_local_state(@inbounds(tls[:CUDA])::TaskLocalState)
    else
        ...
        tls[:CUDA] = TaskLocalState(args...)
    end::TaskLocalState
end
```

Same TLS key `:CUDA`, but the struct is defined in different modules post-split: `CUDA.TaskLocalState` (CUDA 5.x) vs `CUDACore.TaskLocalState` (CUDACore 6.x). Nominally distinct types in Julia.

`cuDNN` 6 imports `using CUDACore` (no `using CUDA`) and calls `CUDACore.math_mode()` / `CUDACore.active_state()` — see `lib/cudnn/src/cuDNN.jl:50` and `:80`. `CUDACore.active_state` calls `task_local_state!()`, which is line 73 in the user's stack trace.

When the project has CUDA 5.10.1 already loaded:

1. CUDA 5.10.1 fills `task_local_storage()[:CUDA]` with a `CUDA.TaskLocalState`.
2. cuDNN 6 → CUDACore reads `tls[:CUDA]`, typeasserts it as `CUDACore.TaskLocalState`. Fails.

`CUDA` 6.x re-exports from `CUDACore` so `CUDA.TaskLocalState === CUDACore.TaskLocalState`; the bug only manifests when `CUDA` 5.x is the resolved version.

## Why the resolver allows the broken combo

`lib/cudnn/Project.toml` at master:

```toml
[deps]
CEnum = ...
CUDACore = ...
CUDA_Runtime_Discovery = ...
CUDNN_jll = ...

[compat]
CEnum = "0.2, 0.3, 0.4, 0.5"
CUDACore = "=6.1.0"
CUDA_Runtime_Discovery = "2"
CUDNN_jll = "9.20"
julia = "1.10"
```

No `CUDA` entry. `JuliaRegistries/General/C/cuDNN/Compat.toml` mirrors this:

```toml
[6]
CEnum = "0.2 - 0.5"
CUDNN_jll = "9.20.0 - 9"
julia = "1.10.0 - 1"

["6.1 - 6"]
CUDACore = "6.1.0"
CUDA_Runtime_Discovery = "2"
```

No `CUDA` row anywhere on `cuDNN` 6. Compare to `cuDNN` `1.4.7 - 1`, which does have `CUDA = "5.11"`. So `cuDNN` 1.x is paired with `CUDA` 5.x, but `cuDNN` 6.x is unconstrained against `CUDA`, leaving the resolver free to land `CUDA 5.x + cuDNN 6.x`.

## Possible fixes

1. **Add `CUDA` as a hard dep of `cuDNN` 6 with `[compat] CUDA = "6"`**. Simplest, but reintroduces a dep that the 6.0 split deliberately removed.

2. **Add `CUDA` as a `[weakdeps]` entry on `cuDNN` 6 with `[compat] CUDA = "6"`** (no extension body). Resolver enforces only when `CUDA` is in the env — which is exactly the broken case, since `NNlibCUDACUDNNExt = ["CUDA", "cuDNN"]` only loads when both are present. Preserves the split.

3. **Retroactive registry compat**. Edit `C/cuDNN/Compat.toml` to add `CUDA = "6"` for `[6]`. Affects already-released 6.0.0 and 6.1.0; can break existing user Manifests, so probably less desirable than just cutting a `cuDNN` 6.1.1 with the source fix.

Happy to open the source PR for whichever shape (1) or (2) you prefer — wanted to surface the diagnosis here first.

cc @maleadt

maleadt:
Versions 5 and 6 if the CUDA stack are incompatible, however, it's currently possible to co-install cuDNN@6, which depends on only CUDACore (things have been split up), with CUDA@5 which doesn't have a CUDACore dependency yet (pre-split).

Avoid this situation by adding an artificial weak dependency on CUDACore to old CUDA.jl versions, preventing this incompatible combination.

Fixes https://github.com/JuliaGPU/CUDA.jl/issues/3130


maleadt:
Looks like this causes installation issues on 1.10 and 1.11...

PR #156488, opened 2026-05-28
Fix Oceananigans version range for Breeze 0.5.1

navidcy:
thanks @giordano

Close https://github.com/NumericalEarth/Breeze.jl/pull/744

PR #156490, opened 2026-05-28
Fix Oceananigans version range for NumericalEarth 0.5.3

navidcy:


giordano (dismissed):
Similar to #156488, [Oceananigans v0.108.1](https://github.com/CliMA/Oceananigans.jl/releases/tag/v0.108.1) changed the syntax of a function which is used by NumericalEarth.  A new breaking version, [v0.109.0](https://github.com/CliMA/Oceananigans.jl/releases/tag/v0.109.0), was tagged, but we still need not to allow NumericalEarth to use incompatible versions.

giordano (approved):

PR #156977, opened 2026-06-03
Reactant: Upper bound GPUCompiler compat

wsmoses:
Later GPUCompiler adds instcombine into the cuda compilation pipeline, which breaks kernel raising. see https://github.com/EnzymeAD/Enzyme-JAX/issues/2522

cc @giordano @vchuravy @maleadt @gbaraldi


maleadt:
If Reactant is so sensitive to GPUCompiler internals, consider using tighter bounds.

maleadt (dismissed):

giordano:
This isn't sufficient, you need to cap all GPUCompiler entries, also at https://github.com/JuliaRegistries/General/blob/1f3380140217f6918847618db1f54b56df902593/R/Reactant/Compat.toml#L390 and https://github.com/JuliaRegistries/General/blob/1f3380140217f6918847618db1f54b56df902593/R/Reactant/Compat.toml#L411

wsmoses:
@giordano okay fixed that

giordano (approved):

PR #157792, opened 2026-06-11
Update SymbolicIntegrationMaxima source

Amin-El-Sayed:
Updates the `SymbolicIntegrationMaxima` registry metadata after the package moved into `JuliaSymbolics/SymbolicIntegration.jl` as the `lib/SymbolicIntegrationMaxima` subpackage.

This changes:
- `repo` to `https://github.com/JuliaSymbolics/SymbolicIntegration.jl.git`
- adds `subdir = "lib/SymbolicIntegrationMaxima"`

The package UUID and registered versions are unchanged.

### DIFF
--- a/S/SymbolicIntegrationMaxima/Package.toml
+++ b/S/SymbolicIntegrationMaxima/Package.toml
@@ -1,3 +1,4 @@
 name = "SymbolicIntegrationMaxima"
 uuid = "6f8f0c92-4498-4b29-a8cc-7d11f58db8c9"
-repo = "https://github.com/Amin-El-Sayed/SymbolicIntegrationMaxima.jl.git"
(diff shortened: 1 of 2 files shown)


GunnarFarneback:
> registered versions are unchanged.

No, they aren't. The PR includes a changed tree hash, which is not allowed.

Amin-El-Sayed:
You are right, thanks for catching this. I misunderstood the registry invariant here: changing the tree hash would mutate the already registered v0.1.1, which is not allowed.

I am going to close this PR. The correct path is to leave v0.1.1 unchanged and register a new SymbolicIntegrationMaxima version from the SymbolicIntegration.jl subdirectory after the package version is bumped there.

Amin-El-Sayed:
Closing in favor of a new-version registration flow. Existing registered versions and their tree hashes should remain unchanged.

PR #157885, opened 2026-06-12
[Reactant] Fix compat with ReactantCore

giordano:
[Reactant@0.2.265 effectively requires ReactantCore@0.1.20](https://github.com/EnzymeAD/Reactant.jl/pull/2972), but when that was released ReactantCore@0.1.20 was actually not released yet and the compat bound wasn't updated accordingly. [ReactantCore@0.1.20 is now out](https://github.com/JuliaRegistries/General/pull/157838) and this change is reflecting the real compat bound.

CC @pangoraw @wsmoses for your information.


Pangoraw (approved):
Thank you!

DilumAluthge (approved):

PR #158518, opened 2026-06-19
Yank CSL 1.5.4

Keno:
MinGW 13 has an ABI break and we're backing out the bump. See https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/475.


giordano (approved):
This should be pretty safe because CompilerSupportLibraries_jll is a non-upgradable standard library, so no one should be affect by this change in practice

PR #160719, opened 2026-07-10
Update HypercubeTransform deps

ptiede:
This is needed because Distributions changed the type signature in 0.25.129 for an alias. HypercubeTransform patched this, but I forgot to update the dependency to only support versions 0.225.129 and above. 

Patch assisted with Claude code.


GunnarFarneback (approved):
$ manual_merge_analysis 160719
Analyzing PR 160719 by ptiede
Title: Update HypercubeTransform deps

Identified as a retro compat change PR for HypercubeTransform.

? Compat.toml does not roundtrip through compressed load+save up to space differences. This is probably ok.
0.4.16 Distributions 0.24 - 0.25 => 0.25.129 - 0.25

PR #162304, opened 2026-07-25
Add missing upper compat bounds for JasPer_jll

eschnett:
`JasPer` 4.0 has been released and is not ABI compatible with older versions. Add missing upper compat bounds for all current users of `JasPer_jll`.


imciner2:
@eschnett Can you rebase this, or is this no longer needed?

[noblock]

eschnett:
Still needed. Rebased.

giordano (approved):

PR #162780, opened 2026-07-30
Correct OrdinaryDiffEqDefault 2.4.1 Core floor

ChrisRackauckas-Claude:
## Summary

- require OrdinaryDiffEqCore 4.11 or later for OrdinaryDiffEqDefault 2.4.1 and later, by bumping the existing `["2.4.1 - 2"]` entry

## Why

OrdinaryDiffEqDefault 2.4.1 uses `AutoDePSpecialize` and the `lorenz_pref` precompile workload helpers. OrdinaryDiffEqCore 4.11.0 is the first registered Core release containing all of those names.

Without this correction, the registry can resolve 2.4.1 with Core 4.10.0 and fail while precompiling with `UndefVarError: lorenz_pref not defined in OrdinaryDiffEqCore`.

## Local verification

- reproduced the precompile failure with unmodified registry metadata and Core 4.10.0
- applied this metadata change and resolved the exact OrdinaryDiffEqDefault 2.4.1 release with Core 4.11.0
- loaded the corrected environment and passed all 5 targeted compatibility checks

Please ignore this draft until it has been reviewed by @ChrisRackauckas.

### DIFF
--- a/O/OrdinaryDiffEqDefault/Compat.toml
+++ b/O/OrdinaryDiffEqDefault/Compat.toml
@@ -123,6 +123,8 @@ SciMLBase = "3.35.1 - 3"
 ["2.4 - 2"]
 LinearSolve = "5.1.0 - 5"
 
+["2.4.1"]
+OrdinaryDiffEqCore = "4.11.0 - 4"
+
 ["2.4.1 - 2"]
-OrdinaryDiffEqCore = "4.8.0 - 4"
 SciMLBase = "3.39.0 - 3"
--- a/O/OrdinaryDiffEqDefault/Compat.toml
+++ b/O/OrdinaryDiffEqDefault/Compat.toml
@@ -123,8 +123,6 @@ SciMLBase = "3.35.1 - 3"
 ["2.4 - 2"]
 LinearSolve = "5.1.0 - 5"
 
-["2.4.1"]
-OrdinaryDiffEqCore = "4.11.0 - 4"
-
 ["2.4.1 - 2"]
+OrdinaryDiffEqCore = "4.11.0 - 4"
 SciMLBase = "3.39.0 - 3"


ChrisRackauckas (dismissed):

O/OrdinaryDiffEqDefault/Compat.toml
```diff
@@ -123,6 +123,8 @@ SciMLBase = "3.35.1 - 3"
 ["2.4 - 2"]
 LinearSolve = "5.1.0 - 5"
 
+["2.4.1"]
+OrdinaryDiffEqCore = "4.11.0 - 4"
+
 ["2.4.1 - 2"]
-OrdinaryDiffEqCore = "4.8.0 - 4"
```
giordano:
```toml
["2.4.1 - 2"]
OrdinaryDiffEqCore = "4.11.0 - 4"
```
would be more standard

> leave future OrdinaryDiffEqDefault releases free to declare a different compatibility floor

This doesn't make much sense: future versions would change the compat bound anyway automatically.

ChrisRackauckas-Claude: Done — bumped the existing `["2.4.1 - 2"]` entry to `"4.11.0 - 4"` instead of adding an exact-version section, and dropped that (incorrect) bullet from the description. 2.4.1 is currently the latest registered version, so the two forms are equivalent today.

giordano (approved):

giordano:
But there's a conflict now

ChrisRackauckas-Claude:
Superseded — closing.

This change already landed in master via #162878 (merged 2026-08-01), which raised the `OrdinaryDiffEqCore` floor to `4.11.0 - 4` for the same five sublibrary versions, including OrdinaryDiffEqDefault 2.4.1. Current `O/OrdinaryDiffEqDefault/Compat.toml` on master:

```toml
["2.4.1"]
OrdinaryDiffEqCore = "4.11.0 - 4"

["2.4.1 - 2"]
SciMLBase = "3.39.0 - 3"

["2.4.2 - 2"]
OrdinaryDiffEqCore = "4.11.0 - 4"
```

2.4.1 and 2.4.2 are the only registered versions at or above 2.4.1, so the floor is fully covered.

Rebasing this branch onto master drops the substantive commit as already-applied, leaving only a cosmetic merge of the `["2.4.1"]` block into `["2.4.1 - 2"]` — no resolver behavior change. Not worth the churn, so closing rather than pushing a no-op.

Sorry for the noise, @giordano.

PR #163056, opened 2026-08-01
New package: KaimonSlate v1.0.0

kahliburke:
- Registering package: KaimonSlate
- Repository: https://github.com/kahliburke/KaimonSlate.jl
- Created by: @kahliburke
- Version: v1.0.0
- Commit: 21c0b7f00e8035b055bfce043e125fcdca985275


GunnarFarneback:
Please register through the Registrator comment or web interfaces. Manual registration of new packages is only an option for git hosting services without Registrator support.

If your intention is to let the waiting time run in parallel for this package and a dependency, it's possible to do so by the following dirty workaround:

1. Remove the not yet registered dependency from this package.
2. Register with Registrator. It will accept the registration but automerge will complain about the package being broken.
3. Wait for the dependency to be registered.
4. Add back the dependency to this package. **Do not change the version number.**
5. Re-register with Registrator. The same PR will be updated without resetting the timer. Automerge will find that it's no longer broken.

The simpler alternative is of course to just register the packages sequentially.

kahliburke:
@GunnarFarneback Or, you could have just not closed this?

GunnarFarneback:
As you're anxious to have it open I'll reopen it but don't expect that it will make the process faster. Automerge is the fast lane.

kahliburke:
@GunnarFarneback -- "The simpler alternative is of course to just register the packages sequentially."

Ok, so what you're saying is that if I want to properly structure packages in Julia, then the 'simpler' way is to incur an extra 3 day overhead for each layered dependency? Doesn't that actively discourage good package structure?

And then, when I initiate a method that will allow me to avoid paying an extra 3 day overhead tax for no discernible benefit, your response is to close the PR to prevent me from doing so?

But instead, I need to use some different "dirty" workaround? 

Umm ... why?

GunnarFarneback:
I agree that the dirty workaround isn't great but that's the best option the tooling currently provides.

kahliburke:
Does this PR not satisfy the 3 day timeout, and then I could get the automerge process to utilize that?

I'm simply trying to release a package and I have a thin interface package that's in the same repository that I've already initiated, but this 3 day waiting period plus an additional 3 day waiting period just slows things down so much. I really think some changes need to be made here.

kahliburke:
Ok, I checked again and you're probably correct that this doesn't help the timer issue. I just really think there needs to be a better way to do this. I'll do the dirty workaround you suggest.

kahliburke:
@GunnarFarneback As far as I can discern, Julia is the only language I'm aware of that enforces a 3 day waiting period on package registration. The fact that this compounds with any layered package structure, which in some cases is arguably the correct approach (as I would say in my current case), seems like it needs improvement.

The pain doesn't stop at registration either, because updates to packages with this layered structure are also impacted, it takes far longer than it should to simply release an update to a base package and the affected downstream packages. 

I really feel like this is an area in need of improvement.

goerz:
> I'll do the dirty workaround you suggest.

It’s the workaround I already told you about yesterday, including the fact that this is the best the tooling currently provides 🤷‍♂️

> Julia is the only language I'm aware of that enforces a 3 day waiting period on package registration.

That’s not going to change. Community review is a cornerstone of the General registry. But as far as the broader landscape of scientific computing goes, it’s pretty easy-going: [CRAN has mandatory reviews](https://cran.r-project.org/submit.html), so that’s an indeterminate waiting period!

Godot2026:
Hi, @kahliburke ! Please come back! I really need your apps!!! I really need Kaimon and KaimonSlate!!!!!!!

Godot2026:
Please come back to the community.

PR #163390, opened 2026-08-04
Point the five Khepri packages at the aptmcl/Khepri monorepo

aptmcl:
The packages moved from individual repositories into a monorepo. Each entry now names that repository and the subdirectory the package occupies.

Versions.toml, Deps.toml and Compat.toml are deliberately untouched: the git-tree-sha1 recorded for every already-registered version refers to a tree in the original per-package repository, and those repositories are archived read-only rather than deleted, so existing versions continue to resolve.


GunnarFarneback:
Closed in favor of #163667.

PR #164788, opened 2026-08-17
[libgiac_julia_jll] cap v0.5.0 to GIAC_jll 2.0.1 (ABI break in 2.0.2)

s-celles:
## What

Restrict `libgiac_julia_jll` v0.5.0 to `GIAC_jll = "2.0.1"`.

```diff
-[0]
-GIAC_jll = "2.0.1-2"
+["0.5.0"]
+GIAC_jll = "2.0.1-2.0.1"
```

(Only `0.5.0+0` is registered, so pinning the section to that version is equivalent today and leaves the range free for the upcoming `0.5.1`.)

## Why

`libgiac_julia_jll` is a CxxWrap shim compiled against GIAC's C++ headers, so it is bound to GIAC's ABI, not just its API.

GIAC 2.0.2 ([Yggdrasil#13717](https://github.com/JuliaPackaging/Yggdrasil/pull/13717), not yet merged) changes the memory layout of `giac::gen`: the type tag moves from a 5-bit bitfield to a plain byte (`GIAC_TYPE_ON_8BITS`). The bitfield was packed differently by GCC 8 and GCC 10, which is the bug being fixed — but it does mean a wrapper compiled against 2.0.1 headers misreads every `gen` it is handed by a 2.0.2 library.

v0.5.0 declared `compat = "2.0.1"`, which resolves as `[2.0.1, 3.0.0)`. So the moment `GIAC_jll` 2.0.2 is registered, an existing `Giac.jl` user doing `] up` can be resolved onto **old wrapper + new library**, with no error at install or load time — the corruption shows up later as numeric types being misidentified (`REAL` read as `DOUBLE`).

This is exactly the class of failure 2.0.2 exists to remove, so it should not be reintroduced through the resolver.

## Timing

This bound wants to be in place **before or together with** the registration of `GIAC_jll` 2.0.2 — after is too late for anyone who updates in between.

`libgiac_julia_jll` v0.5.1 is being rebuilt against 2.0.2 and will carry its own exact bound; it is unaffected by this change.

## Checks

- Only `0.5.0+0` is registered (`Versions.toml`), so no other version's resolution changes.
- `GIAC_jll` 2.0.1 remains resolvable, so existing working installs keep working.

Assisted by AI


jll/L/libgiac_julia_jll/Compat.toml
```diff
@@ -1,6 +1,6 @@
-[0]
+["0.5.0"]
 Artifacts = ["0.0.0", "1"]
-GIAC_jll = "2.0.1-2"
+GIAC_jll = "2.0.1-2.0.1"
```
imciner2: Couldn't you force this by just doing `=2.0.1` in the compat syntax instead of defining a range?

s-celles: Thanks @imciner2 

My agent tells me

```
julia +release -e '
  using Pkg
  V = Pkg.Versions
  for s in ("=2.0.1", "2.0.1-2.0.1", "2.0.1", "2.0.1-2")
      try
          spec = V.VersionSpec(s)
          ins = [v for v in (v"2.0.0", v"2.0.1", v"2.0.2", v"2.1.0") if v in spec]
          println(rpad(repr(s),16), " → ", rpad(string(spec),22), " contient: ", ins)
      catch e
          println(rpad(repr(s),16), " → ERROR: ", sprint(showerror, e)[1:min(end,90)])
      end
  end'

"=2.0.1"        → ERROR: invalid base-10 digit '=' in "=2"
"2.0.1"         → 2.0.1        includes 2.0.1, 2.0.1+0, 2.0.1+1
"2.0.1-2.0.1"   → 2.0.1        includes 2.0.1, 2.0.1+0, 2.0.1+1
"2.0.1-2"       → 2.0.1 - 2    also includes 2.0.2, 2.0.2+0   ← current behavior

=2.0.1 is the syntax for [compat] in a Project.toml file, not for Compat.toml, where values ​​are parsed as VersionSpec objects. The = causes a parsing error there.

On the other hand, just "2.0.1" does exactly what is required, and it is simpler than my "2.0.1-2.0.1". The two are strictly equivalent, including regarding the point that concerned me: JLL build numbers—2.0.1+0 and 2.0.1+1 remain included, while 2.0.2 remains excluded.
```

So you may probably prefer here "2.0.1", isn't it?

s-celles: Good catch on the intent, though `=2.0.1` isn't valid here — `Compat.toml` values are parsed as `VersionSpec`, not as `Project.toml` compat syntax:

```
VersionSpec("=2.0.1")  →  ArgumentError: invalid base 10 digit '=' in "=2"
```

But plain `"2.0.1"` does exactly what you're after, and is simpler than the range I wrote:

```
"2.0.1"        →  2.0.1      (matches 2.0.1, 2.0.1+0, 2.0.1+1)
"2.0.1-2.0.1"  →  2.0.1      (identical)
"2.0.1-2"      →  2.0.1 - 2  (also matches 2.0.2 — the current, too-wide bound)
```

Build numbers are still matched, which matters here since the registered version is 2.0.1+0. Updated to `"2.0.1"`.

jll/L/libgiac_julia_jll/Compat.toml
```diff
@@ -1,6 +1,6 @@
-[0]
+["0.5.0"]
```
giordano: Why this change?

giordano: I'm convinced this is wrong and should be reverted
```suggestion
[0]
```

giordano (approved):

s-celles:
Thanks @giordano

PR #165283, opened 2026-08-21
HTTP: raise Reseau floor to 1.3.5 (earlier versions cannot precompile — expired TLS fixture)

kahliburke:
Retroactively raises the `Reseau` lower bound on HTTP's 2.x compat ranges to `1.3.5`.

## Why

Reseau `<= 1.3.4` ships a TLS fixture (`test/resources/unittests.crt`) that its own
`PrecompileTools` workload connects to and verifies with `verify_peer = true`. That
certificate expired on **2026-08-06 17:07:04 UTC**, so those versions can no longer
precompile on any machine:

```
Failed to precompile Reseau [802f3686-a58f-41ce-bb0c-3c43c75bba36]
ERROR: LoadError: tls handshake failed: tls: certificate has expired or is not yet valid
  (valid unix range 1714842424-1786036024)
  [6] _pc_run_tls_roundtrip_states!(...)
      @ Reseau .../src/6_precompile_workload.jl:558
```

The failure propagates: HTTP fails to precompile, and so does everything depending on it.

Certificate validity by tag:

| Reseau | `notAfter` |
|---|---|
| v1.3.4 | Aug 6 17:07:04 2026 |
| v1.3.5 | Jan 1 00:00:00 2036 |

Upstream issue: JuliaServices/Reseau.jl#139, fixed by `f5b0b4ca` and released in **1.3.5**.

## Why a registry change rather than a new HTTP release

HTTP's declared bounds still admit Reseau `1.1.1 - 1.3.4`, so any resolution landing in that
range produces an environment that cannot build. A new HTTP release raising the floor would
only help from that release onward — every existing 2.x would keep admitting the broken
range, which is precisely how users are hitting this. Hence the retroactive edit.

This excludes only versions that cannot work at all today, so nothing functional is lost.

## Scope

All three of HTTP's `Reseau` ranges are affected and all three are raised here:

```diff
 ["2 - 2.1.0"]
-Reseau = "1.1.1 - 1"
+Reseau = "1.3.5 - 1"

 ["2.1.1 - 2.1"]
-Reseau = "1.2.1 - 1"
+Reseau = "1.3.5 - 1"

 ["2.2 - 2"]
-Reseau = "1.3.0 - 1"
+Reseau = "1.3.5 - 1"
```

Reseau is 1.x, so `1.3.5` is semver-compatible with each of the older declared floors.

If you'd prefer a smaller blast radius, restricting the change to `["2.2 - 2"]` still covers
every current HTTP release; the older two ranges are included only because leaving them as-is
keeps admitting an environment that cannot precompile. Happy to trim on request.


kahliburke:
@goerz Wondering if someone might look at this, this bit me before and I just ran into it again today.

goerz:
@quinnj?

goerz:
@kahliburke Did you open an issue about this with `HTTP.jl`? Normally, I'd expect this kind of PR to come from the maintainers of the package (whoever would be authorized to manage releases). I guess I _could_ merge this, if there's no action from an `HTTP` maintainer. Certainly seems legit. Not sure what to do… I don't really want package maintainers coming at me with "why did you mess with my package without my consent?"

quinnj:
Yes, this is a good change/update.

quinnj:
This does seem a little.....out of hte ordinary? Like I don't think we generally add restrictive bounds like this when teh solution is just "update the dep" because HTTP isnt' holding you back from updating at all. I think we can probably close this because it's probably an AI agent who bumped into something and got a little too excited/pedantic about bumping the registry bounds.

goerz:
@quinnj Thanks for the response!

You're giving me a bit of mixed signals here:

> Yes, this is a good change/update.
>
> I think we can probably close this

So, merge this? Close it? You want to think about this more, or talk this through more with @kahliburke?


I'm going to let you guys figure this out, just let me know at any point if you need help with me clicking a "merge" or "close" button

quinnj:
Sorry about that. I initially scanned and said it was fine. But thinking more about it and the discussion in slack, I don't think it's necessary.

goerz:
Alright, closing then

kahliburke:
This is a real issue that people run into. The issue is that there are versions that can be installed where some other change (for me it was upgrading to Julia 1.12.7) cause a precompilation step, and any package using HTTP break. I've had users of my packages run into this issue as well, they've message me about it.

The issue is that there are versions of HTTP in the registry which worked when they were installed but no longer work. I'd urge you to reconsider.

quinnj:
Yeah, I get it. It is particularly annoying because these versions of Reseau w/ expired certs are just not useable. I think teh right course of action there though is to yank the expired, "bad" versions though and then it should similarly resolve the issue you're describing.

kahliburke:
@quinnj Ok, that sounds like it could work too, I don't know that there is a perfect solution for this situation that doesn't result in some error condition for somebody in some situation. It's your package @quinnj, you should take the action you feel is best. 

Thanks for taking the time to fix it. It's an unusual case to be sure, and my idea here was not a perfect solution either. Yanking the Reseau versions that can't precompile is likely cleaner overall.

Is there anything you'd like me to do further on this?

PR #165754, opened 2026-08-24
Revert "wrappers must have suffix" rule

goerz:
This revises the package naming guidelines, following discussion among registry maintainers about the naming rules having gone a bit too far.

The main change reverts [rule 6 ("wrappers")](https://github.com/JuliaRegistries/General/blob/9a6bf4107b3b9045ac4deff12d851b445139690a/NAMING_GUIDELINES.md#rule-wrappers) to its pre-February form, dropping the request that wrapper packages indicate in their name that they wrap an external library (`ABCWrapper`, `LibABC`, etc.). A significant number of registry maintainers consider such prefixes and suffixes noise at best. The reverted rule also changes "should" to "may", and nothing in it overrides the other rules: wrappers that fail AutoMerge (4-letter acronyms) still need manual approval, and a wrapper that passes AutoMerge can still be engaged with by the community with suggestions for a more appropriate name.

In addition, this PR makes the following refinements:

- In rule 1 ("avoid jargon and acronyms"), guidance for disambiguating acronyms:

  > * Follow existing commonly used prefixes or suffixes to disambiguate acronyms, e.g.:
  >   - Suffixes `Files`, `Format`, `Data`, `Writer`, `Reader`, `Parser`, for file format I/O packages, e.g., `PNGFiles.jl`, not `PNG.jl`
  >   - Suffix `Client` for protocol or database clients, e.g. `SMTPClient.jl`, not `SMTP.jl`
  >   - Suffixes like `Solver`, `Models`, `Optimizer` for optimization methods: `ADNLPModels.jl`, not `ADNLP.jl`

  These are based on patterns that exist in a significant number of registered packages.

- In rule 4 ("err on the side of clarity"): "Three-letter package names are not suitable." This reflects current behavior; three-letter names fail AutoMerge and there is nobody willing to merge them. Four letters is a hard minimum, five letters a soft minimum decided case by case.

- In rule 5 ("less systematic name"):

  > * Pronounceable acronyms (backronyms) are usually fine, and may count as "less systematic names", e.g. `ADRIA.jl` ("Adaptive Dynamic Reef Intervention Algorithms")

[...]

PR #165756, opened 2026-08-24
Naming guidelines: add example for a name with identical scope to an existing package

goerz:
Stacked on top of #165754 (base branch `mg-naming-guidelines`); merge that one first.  Splitting this out from #165754 since it deserves its own discussion.

This adds one more example to rule 7 ("avoid close names"), for a case where the new name implies the same scope as an existing package rather than a typo-level similarity:

> * `QuasiMC` indicates an identical scope to `QuasiMonteCarlo`. Choose a name that is descriptive of what distinguishes the new package, or use a ["less systematic name"](#rule-less-systematic). Or, contribute to the existing package.

There is a more narrow version of this: package names that differ from existing registered packages only by writing out or abbreviating some part of the name.

But I would endorse the more broader implications for the general governance of the registry: systematically named packages have a moral obligation to be maintained within the scope that their name implies, so that an equivalently named package can reasonably be denied with "please contribute to the existing package". Julia's strength is a coherent ecosystem grown out of our descriptive naming tradition, and it would be a shame if the registry ended up littered with equivalent variations of the same name.


goerz:
Not stale, see https://github.com/JuliaRegistries/General/pull/165754#issuecomment-5400498072

PR #166465, opened 2026-08-29
Require SciMLBase 3.46 for OrdinaryDiffEqBDF/OrdinaryDiffEqNonlinearSolve versions using AutoDespecialize

ChrisRackauckas-Claude:
## What changed and why

Six already-registered versions reference `SciMLBase.AutoDespecialize` but declare a `SciMLBase` lower bound below the version that introduced it. Pkg can therefore resolve them against a `SciMLBase` that lacks the symbol, and the resulting environment fails at precompile time.

`AutoDespecialize` was added in **SciMLBase v3.46.0**:

```console
$ for v in v3.45.0 v3.46.0; do
    printf "%s : " $v
    curl -sL "https://raw.githubusercontent.com/SciML/SciMLBase.jl/$v/src/scimlfunctions.jl" \
      | grep -c "^struct AutoDespecialize"
  done
v3.45.0 : 0
v3.46.0 : 1
```

There are no SciMLBase releases between 3.45.0 and 3.46.0, so `3.46.0` is the exact boundary.

The affected versions, and the first version of each package to reference the symbol:

| Package | Versions referencing `AutoDespecialize` | Declared `SciMLBase` | Required |
|---|---|---|---|
| `OrdinaryDiffEqBDF` | 2.4.3, 2.4.4, 2.4.5 | `3.39.0 - 3` | `3.46.0 - 3` |
| `OrdinaryDiffEqNonlinearSolve` | 2.9.0, 2.9.1, 2.9.2 | `3.43.0 - 3` | `3.46.0 - 3` |

The reference is not behind a runtime branch — it is evaluated during precompilation, in the module-level precompile workload:

```julia
# OrdinaryDiffEqBDF/src/OrdinaryDiffEqBDF.jl:151
if Preferences.@load_preference("PrecompileAutoDespecialize", true)
    dae_f = DAEFunction{true, SciMLBase.AutoDespecialize}(precompile_dae)
```

so any resolution that pairs these versions with SciMLBase < 3.46.0 fails to load at all.

This follows the retroactive-compat path described in the README under *When is yanking a release appropriate?* — the bounds were set too wide, and a patch release cannot fix it because Pkg would still resolve the existing versions.

## Failing before

Observed in the wild on `SciML/SciMLBenchmarks.jl`, whose weekly dependency refresh resolved `OrdinaryDiffEqBDF v2.4.4` alongside `SciMLBase v3.39.1` and then could not precompile:

```console
$ julia +1.12 --project=. -e 'using Pkg; Pkg.precompile(; strict=true)'
              ✗ OrdinaryDiffEqBDF
              ✗ OrdinaryDiffEqDefault
              ✗ OrdinaryDiffEq
ERROR: LoadError: UndefVarError: `AutoDespecialize` not defined in `SciMLBase`
exit=1
```

## Verification of this diff

Effective `SciMLBase` bound per registered version, computed with `Pkg.Versions.VersionRange` against the file before and after the change — only the affected versions move, everything else is byte-identical:

```
#### OrdinaryDiffEqBDF
  v2.4.0    same  before=["3.35.1 - 3"]  after=["3.35.1 - 3"]
  v2.4.1    same  before=["3.39.0 - 3"]  after=["3.39.0 - 3"]
  v2.4.2    same  before=["3.39.0 - 3"]  after=["3.39.0 - 3"]
  v2.4.3  CHANGED  before=["3.39.0 - 3"]  after=["3.46.0 - 3"]
  v2.4.4  CHANGED  before=["3.39.0 - 3"]  after=["3.46.0 - 3"]
  v2.4.5  CHANGED  before=["3.39.0 - 3"]  after=["3.46.0 - 3"]

#### OrdinaryDiffEqNonlinearSolve
  v2.7.0    same  before=["3.39.0 - 3"]  after=["3.39.0 - 3"]
  v2.8.0    same  before=["3.43.0 - 3"]  after=["3.43.0 - 3"]
  v2.9.0  CHANGED  before=["3.43.0 - 3"]  after=["3.46.0 - 3"]
  v2.9.1  CHANGED  before=["3.43.0 - 3"]  after=["3.46.0 - 3"]
  v2.9.2  CHANGED  before=["3.43.0 - 3"]  after=["3.46.0 - 3"]
```

Both files parse as valid TOML.

## What I did not verify

- I did not re-resolve every downstream environment in General that depends on these packages; the check above is confined to the effective bounds of the two edited files.
- The scan for affected packages covered package versions present in a local depot plus the full registered version lists of these two packages. Other packages referencing `AutoDespecialize` with a too-low bound may exist and would need the same treatment.

## Note for maintainers

`master` of `SciML/OrdinaryDiffEq.jl` still carries the too-low bounds (`lib/OrdinaryDiffEqBDF/Project.toml` → `SciMLBase = "3.39"`, `lib/OrdinaryDiffEqNonlinearSolve/Project.toml` → `SciMLBase = "3.43"`), so the next release of either would reintroduce the problem. That needs a separate PR upstream; this one only repairs the already-registered versions.

🤖 Generated with Claude Code (model: claude-opus-5[1m])
https://claude.ai/code/session_016yBsJDGXDqPHqaJCVzpyr9

### DIFF
--- a/O/OrdinaryDiffEqBDF/Compat.toml
+++ b/O/OrdinaryDiffEqBDF/Compat.toml
@@ -194,10 +194,15 @@ MuladdMacro = "0.2.4 - 0.2"
 ["2.4.1 - 2"]
 ArrayInterface = "7.28.0 - 7"
 OrdinaryDiffEqDifferentiation = "3.3.0 - 3"
+
+["2.4.1 - 2.4.2"]
 SciMLBase = "3.39.0 - 3"
 
 ["2.4.1 - 2.4.4"]
 OrdinaryDiffEqCore = "4.12.0 - 4"
+
+["2.4.3 - 2"]
+SciMLBase = "3.46.0 - 3"
 
 ["2.4.5 - 2"]
 OrdinaryDiffEqCore = "4.15.1 - 4"
--- a/O/OrdinaryDiffEqNonlinearSolve/Compat.toml
+++ b/O/OrdinaryDiffEqNonlinearSolve/Compat.toml
@@ -283,15 +283,16 @@ CommonSolve = "0.2.6 - 0.2"
 
 ["2.8"]
 NonlinearSolveBase = "2.41.0 - 2"
+SciMLBase = "3.43.0 - 3"
 
 ["2.8 - 2"]
 NonlinearSolve = "4.25.0 - 4"
-SciMLBase = "3.43.0 - 3"
 SciMLPublic = "1.1.0 - 1"
 
 ["2.9 - 2"]
 NonlinearSolveBase = "2.45.0 - 2"
 OrdinaryDiffEqCore = "4.13.0 - 4"
+SciMLBase = "3.46.0 - 3"
 
 ["2.9 - 2.9.1"]
 OrdinaryDiffEqDifferentiation = "3.9.0 - 3"


ChrisRackauckas-Claude:
Adding a minimal standalone reproduction that isolates the bound, independent of the downstream repo where I first hit this. Two environments differing **only** in the pinned `SciMLBase` version, same `OrdinaryDiffEqBDF@2.4.4` in both.

### Fails on the currently-allowed floor

```console
$ julia +1.12 --project=$(mktemp -d) -e '
    using Pkg
    Pkg.add(name="OrdinaryDiffEqBDF", version="2.4.4")
    Pkg.add(name="SciMLBase", version="3.39.1")
    Pkg.precompile(; strict=true)'

⌃ [bbf590c4] ↓ OrdinaryDiffEqCore           v4.15.2 ⇒ v4.14.1
⌃ [127b3ac7] ↓ OrdinaryDiffEqNonlinearSolve v2.9.2  ⇒ v2.7.0
⌅ [0bca4576] ↓ SciMLBase                    v3.50.0 ⇒ v3.39.1
              ✗ OrdinaryDiffEqBDF
ERROR: LoadError: UndefVarError: `AutoDespecialize` not defined in `SciMLBase`
exit=1
```

Note the resolve **succeeds** — `OrdinaryDiffEqBDF@2.4.4` declares `OrdinaryDiffEqCore = "4.12.0 - 4"`, so the resolver is free to walk ODEqCore back to 4.14.1, which in turn permits SciMLBase 3.39.1. Nothing in the declared bounds prevents it; the failure only shows up at precompile.

### Passes at the proposed floor

```console
$ julia +1.12 --project=$(mktemp -d) -e '
    using Pkg
    Pkg.add(name="OrdinaryDiffEqBDF", version="2.4.4")
    Pkg.add(name="SciMLBase", version="3.46.0")
    Pkg.precompile(; strict=true)'

⌃ [bbf590c4] ↓ OrdinaryDiffEqCore v4.15.2 ⇒ v4.15.1
⌃ [0bca4576] ↓ SciMLBase          v3.50.0 ⇒ v3.46.0
  36890.5 ms  ✓ OrdinaryDiffEqBDF
exit=0
```

Same package version, only the SciMLBase pin differs, so `3.46.0` is the exact floor rather than an approximation.

### Which versions are actually exposed

The exposure depends on each version's `OrdinaryDiffEqCore` bound, since a sufficiently strict one masks the problem transitively:

| Version | Declared `OrdinaryDiffEqCore` | Reachable SciMLBase < 3.46? |
|---|---|---|
| BDF 2.4.3, 2.4.4 | `4.12.0 - 4` (allows 4.12 – 4.14.1) | **yes — reproduces above** |
| BDF 2.4.5 | `4.15.1 - 4` (forces ≥ 3.46) | no — masked |
| NLS 2.9.0 – 2.9.2 | `4.13.0 - 4` (allows 4.13 – 4.14.1) | **yes** |

So for `OrdinaryDiffEqBDF@2.4.5` this PR is a no-op in practice and only makes the declared bound truthful rather than dependent on a sibling's stricter constraint. The versions that actually break today are BDF 2.4.3/2.4.4 and NLS 2.9.0–2.9.2, and those are what the split targets.

I have a companion PR queued against `SciML/OrdinaryDiffEq.jl` to correct the bounds on `master` so future releases carry the right floor.

🤖 Generated with Claude Code (model: claude-opus-5[1m])
https://claude.ai/code/session_016yBsJDGXDqPHqaJCVzpyr9

ChrisRackauckas-Claude:
Correcting something I overstated in my previous comment. I claimed `OrdinaryDiffEqNonlinearSolve` 2.9.0–2.9.2 "genuinely break today" alongside the BDF versions. That is not right, and the two packages fail in materially different ways.

I ran the same discriminating pair for NLS that I ran for BDF, and it came back clean:

```console
$ julia +1.12 --project=$(mktemp -d) -e '
    using Pkg
    Pkg.add(name="OrdinaryDiffEqNonlinearSolve", version="2.9.2")
    Pkg.add(name="SciMLBase", version="3.43.0")
    Pkg.precompile(; strict=true)'

⌅ [0bca4576] ↓ SciMLBase v3.50.0 ⇒ v3.43.0
 371485.3 ms  ✓ OrdinaryDiffEqNonlinearSolve
exit=0
```

The difference is where the symbol is referenced:

- **`OrdinaryDiffEqBDF`** — module-level, inside the precompile workload (`src/OrdinaryDiffEqBDF.jl:152`). Evaluated at precompile time, so it is a hard load-time failure. This is the case my earlier reproduction demonstrates, and it stands.
- **`OrdinaryDiffEqNonlinearSolve`** — inside a function body, `initialize!(nlsolver::NLSolver{<:NonlinearSolveAlg, true}, ...)` at `src/newton.jl:149`. Not evaluated at precompile, so it loads fine and would instead throw `UndefVarError` at runtime, and only on the opt-in `NonlinearSolveAlg` path.

So the corrected picture:

| Package | Versions | Failure mode with SciMLBase below the floor |
|---|---|---|
| `OrdinaryDiffEqBDF` 2.4.3, 2.4.4 | precompile | **demonstrated** — hard failure at load |
| `OrdinaryDiffEqBDF` 2.4.5 | none reachable | masked by `OrdinaryDiffEqCore = "4.15.1"` |
| `OrdinaryDiffEqNonlinearSolve` 2.9.0–2.9.2 | runtime | latent — **not** demonstrated by me |

The NLS bound is still wrong on its own terms: both `AutoDespecialize` and `DespecializedParameters` are absent from SciMLBase 3.43.0 and present in 3.46.0, so those versions declare compatibility with a range in which names they reference do not exist.

```console
$ for v in v3.43.0 v3.46.0; do
    printf "%s: " $v
    curl -sL "https://raw.githubusercontent.com/SciML/SciMLBase.jl/$v/src/scimlfunctions.jl" \
      | grep -c "DespecializedParameters"
  done
v3.43.0: 0
v3.46.0: 4
```

Whether a latent runtime error justifies a retroactive compat edit is a maintainer call, and I did not construct a runtime reproduction for it. If you would rather this PR only cover the demonstrated load-time failure, I am happy to drop the `OrdinaryDiffEqNonlinearSolve` half and reduce it to the `OrdinaryDiffEqBDF` split alone — say the word and I will push that.

🤖 Generated with Claude Code (model: claude-opus-5[1m])
https://claude.ai/code/session_016yBsJDGXDqPHqaJCVzpyr9

PR #167474, opened 2026-09-08
Revert "New version: LazyJLLWrappers v2.0.0 (#167244)"

staticfloat:
This reverts commit 66c2d12e86625461722c383bf71100cbc99f5248.


staticfloat:
This reversion is explained [here](https://github.com/JuliaPackaging/BinaryBuilder2.jl/commit/2335ed26df318dc42ec20cb481e4a6e6626a4ae6).  Basically, I didn't think hard enough about what it meant to make a breaking change to this package, and because the change, while technically breaking due to renaming an exported type, didn't actually impact the generated JLLs I feel it's better to revert this commit and re-register as `v1.2.0`

goerz:
~~You can’t revert a release~~

I thought that reverting releases was not possible, due to in incorrect mental model of how package servers interact with the registry. In particular, I was under the impression that reverting a `v2.0.0` release would prevent the release of a future (different) `v2.0.0` version. Since that's not the case, reverting this relatively recent commit seems unlikely to cause any harm to the ecosystem, and thus seems fine to me to merge.

goerz (dismissed):

goerz (approved):

PR #168810, opened 2026-09-19
ClangCompiler: restrict v0.1.0 to Julia 1.12

Gnimuc:
Restrict the registered version, v0.1.0, to the Julia minor it can actually load on.

**The problem.** ClangCompiler binds the Clang that matches the LLVM Julia is built on, and picks its generated bindings at load time:

```julia
const llvm_version = string(Base.libllvm_version.major)
include(joinpath(libdir, llvm_version, "LibClangEx.jl"))
```

v0.1.0 ships only `lib/18` (Julia 1.12's LLVM). Its `Project.toml` said `julia = "1.12"`, which registered as `1.12.0 - 1`, so since Julia 1.13 (LLVM 20) was released `add ClangCompiler` resolves v0.1.0 there and it cannot load — the `lib/20/LibClangEx.jl` it includes is not in that version.

**The change.** One bound in `C/ClangCompiler/Compat.toml`: `julia = "1.12.0 - 1"` → `julia = "1.12"`. Nothing else in the file changes and no version is added or removed. On 1.13 Pkg will then refuse to resolve instead of installing something that fails at precompile.

The series for Julia 1.13 is 0.2 (Gnimuc/ClangCompiler.jl#56), which bounds `julia = "~1.13"` so this cannot recur on the next minor.

~🤖 Generated with [Claude Code](https://claude.com/claude-code)~


Gnimuc:
Please merge this before I tag a new version for Julia 1.13.

